PR #6, opened 2017-11-28
Feature/auth all hands

InfiniteLee:
added mr-social-client as a dependency 
added basic auth
Guardian generates a token with user info and passes it to client via cookies
added health controller for health check endpoint
enabled CDN


assets/package.json
```diff
@@ -6,6 +6,7 @@
     "watch": "brunch watch --stdin"
   },
   "dependencies": {
+    "mr-social-client": "mozilla/mr-social-client#feature/auth_for_allhands",
```
InfiniteLee: this needs to be switched back to master once that branch is merged in

habitat/plan.sh
```diff
@@ -15,7 +15,7 @@ pkg_deps=(
 pkg_build_deps=(
     core/coreutils
     core/git
-    core/yarn
+    idolgirev/yarn
```
InfiniteLee: this plan has an updated version of yarn (which I *think* helps speed up build times). I tried to build our own but failed. This can be reverted without problem if desired.

PR #13, opened 2018-04-03
Set NODE_ENV properly in package build

gfodor:
We weren't building the bundle with NODE_ENV=production


brianpeiris:
Hmm, I think `yarn build` should already cover this. https://github.com/mozilla/mr-social-client/blob/master/package.json#L13

gfodor:
Oh, thanks!

PR #20, opened 2018-04-20
add docker-compose setup

brianpeiris:


config/dev.exs
```diff
@@ -62,7 +63,7 @@ config :ret, Ret.Repo,
   username: "postgres",
   password: "postgres",
   database: "ret_dev",
-  hostname: "localhost",
+  hostname: "db",
```
brianpeiris: Not sure if there is a better way to configure the database host name specifically for docker-compose.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM elixir:1.5
```
gfodor: FYI 1.6 is out

config/dev.exs
gfodor: hmm yeah i am not sure. on the fence which one makes more sense.

Dockerfile
brianpeiris: I stuck to the version used in plan.sh, assuming that's what we actually use https://github.com/mozilla/reticulum/blob/master/habitat/plan.sh#L21

gfodor: ah yes, waiting on habitat to upgrade

gfodor:
lgtm. i might try to fix up the host thing if it ends up being annoying

brianpeiris:
cool, thanks

PR #21, opened 2018-04-20
add hub name to api response

brianpeiris:


gfodor:
deploy process is the same as JS code, fyi

PR #27, opened 2018-04-29
Fix typo: Posion->Poison

johnshaughnessy:
I think this was a typo back when it was added in https://github.com/mozilla/socialmr/commit/857a6f84c244f738c4b0471ed725833232a27fa2#diff-1b1ba7f7894ccf1a0c686635b9d5b942R12

Not sure if this has actually affected anything.


gfodor:
whoa, i wonder why that didn't break anything

PR #42, opened 2018-07-03
Feature/media resource resolution and shutdown

gfodor:
This PR:

- Adds media URL resolution for 2d videos and images. It calls out to yt-dl for known-good domains for videos, then falls back on APIs for imgur, giphy, and deviantart, and then bottoms out with OpenGraph crawling.

- Adds a graceful shutdown handler to Phoenix


lib/ret/media_resolver.ex
```diff
@@ -0,0 +1,225 @@
...
+    end
+  end
+
+  defp get_root_host(nil) do
+    nil
+  end
+
+  defp get_root_host(host) do
+    # Drop subdomains
+    host |> String.split(".") |> Enum.slice(-2..-1) |> Enum.join(".")
```
brianpeiris: Would this break for URLs like http://jenbartel.tumblr.com/image/175384943654 ?

gfodor: this is grabbing the last two parts of the domain name, so it will return `tumblr.com`

brianpeiris: Right, but tumblr.com/image/175384943654 is a 404

gfodor: ah, yeah the `root_host` that is extracted here is not actually used for the underlying requests -- it's just use as a way to pattern match the various handlers

PR #44, opened 2018-07-06
Remove some sites for now, refactor resolver

gfodor:
Refactoring into pipelines, remove some ytdl sites for now.

### DIFF
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -5,12 +5,12 @@ defmodule Ret.MediaResolver do
   @success_status_codes [200]
 
   @ytdl_root_hosts [
-    "youtube.com",
+    # "youtube.com",
     "imgur.com",
-    "instagram.com",
-    "soundcloud.com",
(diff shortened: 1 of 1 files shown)


lib/ret/media_resolver.ex
```diff
@@ -64,21 +66,24 @@ defmodule Ret.MediaResolver do
...
+          |> Map.get(:body)
+          |> Poison.decode!()
+          |> Map.get("access_token")
+
+        "#{api_host}/deviation/#{deviant_id}?access_token=#{token}"
+        |> retry_get_until_success
+        |> Map.get(:body)
+        |> Poison.decode!()
+        |> Kernel.get_in(["content", "src"])
+        |> URI.parse()
```
brianpeiris: Nice. I can't believe this is how we usually code, assigning things and calling functions like like plebs. Pipe all the things!

gfodor: Haha I know crazy right :)

PR #46, opened 2018-07-11
Feature/sketchfab

gfodor:
This PR:

- Adds support for the Sketchfab API
- Causes the GET retry loop to terminate immediately for 4XX HTTP responses
- Fixes an issue where the wrong kind of argument was being passed from YT-DL, causing YT-DL resolution to break

### DIFF
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -49,7 +49,7 @@ defmodule Ret.MediaResolver do
         |> retry_get_until_valid_ytdl_response
 
       case ytdl_resp do
-        %HTTPoison.Response{status_code: 302, headers: headers} = res ->
+        %HTTPoison.Response{status_code: 302, headers: headers} ->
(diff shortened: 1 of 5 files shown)


lib/ret/media_resolver.ex
```diff
@@ -49,8 +49,8 @@ defmodule Ret.MediaResolver do
         |> retry_get_until_valid_ytdl_response
 
       case ytdl_resp do
-        %HTTPoison.Response{status_code: 302, headers: headers} ->
-          {:commit, headers |> media_url_from_ytdl_headers |> resolved}
+        %HTTPoison.Response{status_code: 302, headers: headers} = res ->
```
brianpeiris: Is `res` unused?

gfodor: yup, nice catch. not sure why linter missed it

gfodor: ah, needed to recompile. caught 2 more.

PR #47, opened 2018-07-13
Feature/page cache

gfodor:
This PR completely changes the way reticulum deals with Hubs JS client code. Instead of packing the client code into the Habitat package and serving the files out directly from disk, it now assumes the user will configure an (optional) origin URL that points to the latest set of Hubs HTML pages. These pages will be pro-actively fetched every 15 seconds in a background polling loop, and the main reticulum endpoint will serve up the pages (with META tags injected) as usual.


.gitignore
```diff
@@ -18,7 +18,7 @@ assets/.yarn/
 # Since we are building assets from assets/,
 # we ignore priv/static. You may want to comment
 # this depending on your deployment strategy.
-priv/static/
+# priv/static/
```
brianpeiris: Just delete these lines

lib/ret_web/controllers/page_controller.ex
```diff
@@ -2,30 +2,24 @@ defmodule RetWeb.PageController do
...
-    render_file(conn, "#{get_file_prefix(conn)}link.html")
+    conn |> render_page("link")
+  end
+
+  def render_for_path("/avatar-selector.html", conn) do
+    conn |> render_page("avatar-selector")
+  end
+
+  def render_for_path("/smoke-avatar-selector.html", conn) do
+    conn |> render_page("avatar-selector")
```
brianpeiris: `"smoke-avatar-selector"`?

lib/ret_web/endpoint.ex
```diff
@@ -14,8 +14,7 @@ defmodule RetWeb.Endpoint do
     gzip: false,
     # Due to cloudfront, we want to include max-age in responses	
     cache_control_for_etags: "public, max-age=31536000",
-    only:
-      ~w(assets robots.txt favicon.ico hub-preview.png avatar-selector.html smoke-avatar-selector.html),
```
brianpeiris: So how does this PR handle `assets`? I can't tell.

lib/ret_web/controllers/page_controller.ex
gfodor: confusing but this is actually right -- think of the function arg as matching the route, but the `render_page` function doesn't require you include `smoke-`, it adds the `smoke-` prefix to the file key based on the `conn`

lib/ret_web/endpoint.ex
gfodor: the end-user URLs for assets don't change after this is pushed, they still point to the cloudfront CDN for `assets-prod.reticulum.io`. however, the origin of that cloudfront URL will change. currently, the origin is reticulum, and this line made it so that the JS/CSS/etc files from the packed-in client code got served out statically (and then would be CDN cached from there.) once we push this out, we will update the CDN origin to a (new) S3 bucket dedicated for the Hubs client assets. The new CI build and deploy process for the client simply manipulates the state in this bucket. Reticulum is still responsible for going to that bucket to fetch the proper root html files (for when the non-CDN, root URLs are hit like the index page and the hub page), but the remainder of the assets (the digested JS and CSS, etc) will go directly from S3 -> CDN.

brianpeiris: Got it, thanks for explaining.

PR #48, opened 2018-07-19
Feature/uploads

gfodor:
This PR adds the ability to upload files to Reticulum.

The `/media` API now accepts a file in addition to a URL. If a file is uploaded, a `access_token` is returned in the `meta` object that must be used in an `Authorization: Token` header when accessing the `raw` or `origin` URLs.

Uploaded files are encrypted with one-time-use encryption keys that are returned to the uploader, so the files are inaccessible unless that person has shared the key with the person trying to read the file. Since the key is not persisted, if the user disconnects then the file is effectively inaccessible by anyone, including the server operator.

In addition, a TTL can be added for uploaded files, and a singleton job across the Reticulum cluster vacuums the filesystem for old, no longer accessed files. For multiple servers, this assumes the upload filesystem mount is a NFS share, etc.


lib/ret/uploads.ex
```diff
@@ -0,0 +1,140 @@
...
+          File.rm!(meta_file)
+        end
+      end
+
+      :filelib.fold_files(
+        "#{uploads_storage_path}/expiring",
+        "\\.meta\\.json$",
+        true,
+        process_meta,
+        :acc
```
brianpeiris: This confused me for a bit. Why not just pass in `nil` for the accumulator?

lib/ret/uploads.ex
```diff
@@ -0,0 +1,140 @@
...
+    Logger.info("Uploads: Beginning Vacuum.")
+
+    with uploads_storage_path when is_binary(uploads_storage_path) <-
+           module_config(:storage_path),
+         uploads_ttl when is_integer(uploads_ttl) <- module_config(:ttl) do
+      process_meta = fn meta_file, _acc ->
+        meta = File.read!(meta_file) |> Poison.decode!()
+        blob_file = meta["blob"]
+
+        {:ok, %{atime: atime}} = File.stat(blob_file)
```
brianpeiris: Are we sure `atime` is going to behave as expected? From what I've read, it depends on how the filesystem was mounted.

lib/ret_web/endpoint.ex
```diff
@@ -33,7 +33,9 @@ defmodule RetWeb.Endpoint do
     Plug.Parsers,
     parsers: [:urlencoded, :multipart, :json],
     pass: ["*/*"],
-    json_decoder: Poison
+    json_decoder: Poison,
+    length: 1_000_000_000,
```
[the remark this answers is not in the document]
brianpeiris: 1GB seem far too big to me. I'd cap it at 500MB or even 100MB and adjust based on how often people hit the limit.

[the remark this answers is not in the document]
gfodor: cool, I'll make it 100MB and we'll see where we stumble.

lib/ret/uploads.ex
gfodor: ah, no reason. I'll admit I didn't grok the accumulator in this context (this is an erlang API) -- will clarify and if nil is an acceptable value I'll change it

lib/ret/uploads.ex
gfodor: I will double check -- we're using EFS over NFS so good call making sure it does what we want.

PR #51, opened 2018-08-02
Add super simple sketchfab search API

gfodor:
This adds a super simple /media/search API which currently just allows looking up the first page of a sketchfab user's downloadable models.

### DIFF
--- a/lib/ret/media_search.ex
+++ b/lib/ret/media_search.ex
@@ -1,0 +1,47 @@
+defmodule Ret.MediaSearchQuery do
+  @enforce_keys [:api]
+  defstruct [:api, :user]
+end
+
+defmodule Ret.MediaSearch do
+  import Ret.HttpUtils
+
+  def search(%Ret.MediaSearchQuery{api: "sketchfab", user: user}) do
+    with api_key when is_binary(api_key) <- resolver_config(:sketchfab_api_key) do
+      res =
+        "https://api.sketchfab.com/v3/search?type=models&downloadable=true&user=#{user}"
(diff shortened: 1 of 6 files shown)


lib/ret/media_search.ex
```diff
@@ -0,0 +1,47 @@
...
+  defstruct [:api, :user]
+end
+
+defmodule Ret.MediaSearch do
+  import Ret.HttpUtils
+
+  def search(%Ret.MediaSearchQuery{api: "sketchfab", user: user}) do
+    with api_key when is_binary(api_key) <- resolver_config(:sketchfab_api_key) do
+      res =
+        "https://api.sketchfab.com/v3/search?type=models&downloadable=true&user=#{user}"
```
brianpeiris: I guess we're ignoring pagination and sorting? Picking a `sort_by` might be useful. The default "by relevance" is not very useful if you're just getting a user's model without an actual keyword query. Maybe by newest first?

lib/ret/media_search.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  defp sketchfab_api_result_to_media_result(result) do
+    %{
+      media_url: "https://sketchfab.com/models/#{result["uid"]}",
+      images: %{
+        preview:
+          result["thumbnails"]["images"]
+          |> Enum.sort_by(fn x -> -x["size"] end)
+          |> Enum.at(0)
+          |> Kernel.get_in(["url"])
```
brianpeiris: Will this fail if there are no thumbnails? I think sketchfab allows that.

PR #53, opened 2018-08-07
Recognize MIME type based upon path for uploads

gfodor:
This PR updates the reticulum media API to try to extract the mime type from the uploaded file path if the browser did not specify a content type.


gfodor:
This was causing things like uploading GLTF files to fail to have proper Content-Types since the browser does not recognize them, but our reticulum server can based on the file extension.

PR #54, opened 2018-08-07
Don't store the filename in the meta file

gfodor:
This tiny PR removes storage of the original filename in the meta file for uploads, because we don't need it and it may contain private information.

### DIFF
--- a/lib/ret/uploads.ex
+++ b/lib/ret/uploads.ex
@@ -18,7 +18,6 @@ defmodule Ret.Uploads do
           meta = %{
             content_type: content_type,
             content_length: content_length,
-            filename: filename,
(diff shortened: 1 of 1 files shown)


lib/ret/uploads.ex
```diff
@@ -18,7 +18,6 @@ defmodule Ret.Uploads do
           meta = %{
             content_type: content_type,
             content_length: content_length,
-            filename: filename,
```
brianpeiris: Oh, I guess you could also remove it from the pattern match in this method's signature?

gfodor: ah yep, thanks!

PR #56, opened 2018-08-14
Allow accessing upload with token in query param

brianpeiris:


lib/ret_web/controllers/upload_controller.ex
```diff
@@ -9,30 +17,33 @@ defmodule RetWeb.UploadController do
...
 
-          {:error, :not_allowed} ->
-            conn |> send_resp(401, "")
-        end
+      {:error, :not_allowed} ->
+        conn |> send_resp(401, "")
+    end
+  end
 
+  defp handle(conn, upload_id) do
```
gfodor: I'd either rename this to `extract_header_and_fetch_upload` or just make it `fetch_upload/2`

gfodor:
lgtm, 1 comment!

PR #57, opened 2018-08-16
WIP Scenes API

brianpeiris:


priv/repo/migrations/20180816191344_create_scenes.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Ret.Repo.Migrations.CreateScenes do
+  use Ecto.Migration
+
+  def change do
+    create table(:scenes, prefix: "ret0", primary_key: false) do
+      add :scene_id, :bigint, default: fragment("ret0.next_id()"), primary_key: true
```
brianpeiris: nit: Not sure what convention we're using for parentheses around function arguments. They have parens in most other parts of the code. I just copied the existing migrations.

test.sh
```diff
@@ -0,0 +1,15 @@
...
+	"attribution_link": "twitter.com/a_guy",
+	"author_account_id": "123123",
+	"upload_id": "234234"
+}}')
+echo "$result"
+
+scene_id=$(echo $result | jq -r '.scene_id')
+echo "$scene_id"
+
+curl -k "https://localhost:4000/api/v1/scenes/$scene_id"
```
brianpeiris: I'll remove this for the final merge, but should I add some actual tests instead?

lib/ret_web/controllers/api/v1/scene_controller.ex
```diff
@@ -0,0 +1,29 @@
+defmodule RetWeb.Api.V1.SceneController do
+  use RetWeb, :controller
+
+  alias Ret.Scene
+  alias Ret.Repo
+
+  # Limit to 1 TPS
+  plug(RetWeb.Plugs.RateLimit)
+
+  def show(conn, %{"id" => scene_sid}) do
```
brianpeiris: On second thoughts, not sure if this is the right pattern for API endpoints. Maybe SIDs should only be used in non-API URLs

priv/repo/migrations/20180816191344_create_scenes.exs
gfodor: ah we're using the elixir reformatter (new in last release) -- you can rig up ale in vim to use it

priv/repo/migrations/20180816191344_create_scenes.exs
```diff
@@ -0,0 +1,22 @@
...
+      add :scene_sid, :string
+      add :slug, :string, null: false
+      add :name, :string, null: false
+      add :description, :string
+      add :author_account_id, :integer, null: false
```
gfodor: these id's should be bigints

config/dev.exs
```diff
@@ -70,7 +70,7 @@ config :ret, Ret.Repo,
   username: "postgres",
   password: "postgres",
   database: "ret_dev",
-  hostname: "localhost",
+  hostname: "#{System.get_env("DB_HOST")}" || "localhost",
```
gfodor: this is probably OK, but worth noting that when running this as an erlang release, it'll burn in the environment variable here I think since this is going to be compiled.

gfodor: (since this is `dev` its basically inconsequential)

lib/ret_web/controllers/api/v1/scene_controller.ex
```diff
@@ -0,0 +1,29 @@
+defmodule RetWeb.Api.V1.SceneController do
+  use RetWeb, :controller
+
+  alias Ret.Scene
+  alias Ret.Repo
+
+  # Limit to 1 TPS
+  plug(RetWeb.Plugs.RateLimit)
```
gfodor: this probably should only be applied to the `create` -- i am not sure how to do it offhand

lib/ret_web/controllers/api/v1/scene_controller.ex
gfodor: i think that sounds right

test.sh
gfodor: i think an integration test would be a good idea -- some on me for not getting the other ones rigged up

PR #58, opened 2018-08-28
WIP Scene upload API

brianpeiris:
TODO:
- [ ] Decide whether we get rid of meta files for uploads and just store meta data as db records, even for ephemeral uploads
- [ ] Add a state machine library
- [ ] Add a DELETE action and transition states
- [ ] Have the upload vacuum ignore permanent uploads (based on meta file or DB record)

### DIFF
--- a/lib/ret_web/views/api/v1/scene_view.ex
+++ b/lib/ret_web/views/api/v1/scene_view.ex
@@ -22,10 +22,10 @@ defmodule RetWeb.Api.V1.SceneView do
           name: scene.name,
           attribution_name: scene.attribution_name,
           attribution_link: scene.attribution_link,
-          author_account_id: scene.author_account_id,
+          author_account_id: to_string(scene.author_account_id),
           # TODO BP: Maybe these should be upload fetch urls instead
-          model_upload_id: scene.model_upload_id,
-          screenshot_upload_id: scene.screenshot_upload_id,
+          model_upload_id: to_string(scene.model_upload_id),
+          screenshot_upload_id: to_string(scene.screenshot_upload_id),
           url: url_for_scene(scene)
         }
       ]
--- a/lib/ret_web/views/api/v1/upload_view.ex
+++ b/lib/ret_web/views/api/v1/upload_view.ex
@@ -5,7 +5,7 @@ defmodule RetWeb.Api.V1.UploadView do
   def render("create.json", %{upload: upload}) do
     %{
       status: :ok,
-      upload_id: upload.upload_id
+      upload_id: to_string(upload.upload_id)
     }
   end
 end


lib/ret/scene.ex
```diff
@@ -21,23 +21,26 @@ defmodule Ret.Scene do
     field(:slug, SceneSlug.Type)
     field(:name, :string)
     field(:description, :string)
-    # TODO: BP account and upload tables don't exist yet.
+    # One of "unlisted", "listed", "pinned"
+    field(:state, :string)
```
gfodor: this assertion for the state field probably belongs somewhere, i don't know the idiomatic way, probably in the changeset validation

priv/repo/migrations/20180824204501_create_uploads_table.exs
```diff
@@ -0,0 +1,18 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:uploads, prefix: "ret0", primary_key: false) do
+      add :upload_id, :bigint, default: fragment("ret0.next_id()"), primary_key: true
+      add :upload_uuid, :string
+      add :uploader_account_id, :integer, null: false
+      # TODO BP: These should probably be non-null
+      add :state, :string
+      add :size, :bigint
```
gfodor: probably should add the mime type here

gfodor: ie, for persistent uploads, instead of having a `meta` file, we'll rely upon the database record. we should talk through if we should just convert everything to using the database though, probably would be better.

lib/ret/upload.ex
```diff
@@ -0,0 +1,31 @@
...
+
+  @schema_prefix "ret0"
+  @primary_key {:upload_id, :integer, []}
+
+  schema "uploads" do
+    field(:upload_uuid, :string)
+    field(:uploader_account_id, :integer)
+    # TODO BP: Use a state machine library?
+    # One of "available" or "marked_for_deletion"
+    field(:state, :string)
```
gfodor: assertion probably belongs somewhere

lib/ret_web/controllers/api/v1/upload_controller.ex
```diff
@@ -0,0 +1,42 @@
+defmodule RetWeb.Api.V1.UploadController do
```
gfodor: hmm, I'm not sure if I would break this out into a new controller or not. the name "media controller" may be a bit of a misnomer, but in general I think the conceptual shoe fits to augment that controller to generate these database records.

gfodor: my expectation is there are going to be couple of cases here:

- ephemeral content uploads for in-room object spawning (current use case)
- permanent content uploads for in-room object spawning (persistent objects case, tbd semantics but uploads will land as permanent)
- conversion from ephemeral to permanent upload for in-room object spawning (a room or object converts from ephemeral to persistent)
- scene GLTF/GLB uploading, screenshot/profile image uploading

the last one is def a bit iffy, but the payload we return from the media API seems like it could work for all these cases.

lib/ret/scene.ex
brianpeiris: Sure, though I was actually thinking of using an existing library for these state machine fields. Either https://github.com/joaomdmoura/machinery or https://github.com/asiniy/ecto_state_machine
Would you rather we write our own small layer, or use a library?

gfodor: library is probably fine. didn't dig into those, but it'd be probably nice to be able to use PostgreSQL's `enum` field type if possible

lib/ret_web/controllers/api/v1/upload_controller.ex
brianpeiris: Marshall and I were actually leaning towards switching to the new /uploads/ API for media uploads as well. I think it makes sense to use a RESTful API for things that actually create records. Conceptually, it feels better to me than an amorphous /media/ api, but I could go either way.

gfodor: i see your point. i'll mull this over -- i can't quickly come up with a strong reason for either approach. one soft argument is that adding uploads to the existing API was really smooth, despite not being RESTful, so that might be at least part of an argument that its a good design.

PR #60, opened 2018-09-01
Feature/accounts

gfodor:
This adds a super simple account system to reticulum.

There are a few new tables involved -- accounts, logins, and login_tokens. The accounts table is 1-1 with what would typically be a users table, and the logins table keeps track of what login identifiers (right now, hashed email addresses) can be used to log into an account.

This PR only addresses authentication, not authorization.

Accounts are authenticated via an email "magic link" (client isn't implemented yet.) The protocol is:

- When a client wants to sign in, it connects to a (secret) phoenix channel topic and sends an email address over the channel

- The server generates a temporary login token for that email address and sends an email with the magic link

- Once the link is clicked, the web page will connect to the same phoenix channel topic and send a message with the token, which will effectively authenticate the user.

- Credentials (a JWT) are then broadcast over the channel topic to the originating requestor.

We will probably want to have a verification step on the landing page, perhaps describing the device that is requesting login and some kind of security sigil, since this mechanism for authentication definitely sacrifices security in the name of convenience of logging in on your VR headset remotely. (It's kind of like Duo push 2fa.) For example, if I know the email address of a user, I can send an email to their account which will effectively remotely log me into their account if they make the mistake of clicking the link and confirming the sign in.

Internally, we can get away without storing email addresses since they are only needed incidentally to send the email, and from that point forward hashes and tokens are used, which is nice.


lib/ret_web/email.ex
```diff
@@ -0,0 +1,19 @@
+defmodule RetWeb.Email do
+  use Bamboo.Phoenix, view: RetWeb.EmailView
+
+  def auth_email(to_address, signin_args) do
+    new_email()
+    |> to(to_address)
+    |> from(from_address())
+    |> subject("Your Hubs Sign-In Link")
```
brianpeiris: "Mozilla Hubs" if we're not formatting the from address.

lib/ret_web/email.ex
```diff
@@ -0,0 +1,19 @@
+defmodule RetWeb.Email do
+  use Bamboo.Phoenix, view: RetWeb.EmailView
+
+  def auth_email(to_address, signin_args) do
+    new_email()
+    |> to(to_address)
+    |> from(from_address())
```
brianpeiris: We probably want to format the from field as "Mozilla Hubs <info@hubs-mail.com>". I think Bamboo can take a `{name, address}` tuple here, if I'm reading their docs correctly.

lib/ret_web/email.ex
```diff
@@ -0,0 +1,19 @@
...
+    |> subject("Your Hubs Sign-In Link")
+    |> text_body(
+      "To sign-in to Hubs, please visit the link below. If you did not make this request, please ignore this e-email.\n\n#{
```
brianpeiris: Typo. "e-mail" not "e-email".

lib/ret_web/email.ex
```diff
@@ -0,0 +1,19 @@
...
+        RetWeb.Endpoint.url()
+      }/signin?#{URI.encode_query(signin_args)}"
```
brianpeiris: Will this actually be clickable if we use `text_body`?

lib/ret_web/auth_channel.ex
```diff
@@ -0,0 +1,73 @@
...
+
+  def handle_in("auth_request", %{"email" => email}, socket) do
+    if !Map.get(socket.assigns, :used) do
+      socket = socket |> assign(:used, true)
+
+      # Create token + send email
+      token = LoginToken.new_token_for_email(email)
+      signin_args = %{topic: socket.topic, token: token}
+
+      RetWeb.Email.auth_email(email, signin_args) |> Ret.Mailer.deliver_now()
```
brianpeiris: Is mail rate-limited somewhere?

lib/ret_web/email.ex
brianpeiris: The `/signin` endpoint doesn't exist yet, right?

priv/repo/migrations/20180830224557_create_login.exs
```diff
@@ -0,0 +1,16 @@
...
+    create table(:logins, prefix: "ret0", primary_key: false) do
+      add(:login_id, :bigint, default: fragment("ret0.next_id()"), primary_key: true)
+      add(:identifier_hash, :string, null: false)
+      add(:account_id, references(:accounts, column: :account_id, on_delete: :delete_all))
+
+      timestamps()
+    end
+
+    create(index(:logins, [:identifier_hash], unique: true))
+    create(index(:logins, [:account_id]))
```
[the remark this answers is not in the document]
gfodor: I'm actually not sure -- this system might eventually support more than one login identifier for an account. I'm still undecided about the sanity of separating these tables but that is one reason having them separate would be a good idea. I think you're right that we can add the index now though since we're not supporting it.

lib/ret_web/auth_channel.ex
gfodor: the channel joins are rate limited and you can only send one of these per topic. but its still a pretty high limit (2 TPS) and obviously you can open many sockets simultaneously. as far as globally rate limiting emails I'm not sure if there's a knob on our side, I can check. SES almost certainly will cut us off (but obviously that would suck)

lib/ret/account.ex
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [])
+    |> validate_required([])
+  end
```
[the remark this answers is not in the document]
gfodor: right now I think this is probably dead code, so I can just take it out. I don't think `Repo.insert!` implicitly calls this or anything like that. the `changeset` functions aren't part of an interface or anything afaik, just an idiom.

test/ret/login_token_test.exs
```diff
@@ -0,0 +1,21 @@
...
+    token = LoginToken.new_token_for_email("test@mozilla.com")
+    assert LoginToken.identifier_hash_for_token(token) == Ret.Crypto.hash("test@mozilla.com")
+  end
+
+  test "should allow expiring a token" do
+    token = LoginToken.new_token_for_email("test@mozilla.com")
+    LoginToken.expire!(token)
+    assert LoginToken.identifier_hash_for_token(token) == nil
+  end
+end
```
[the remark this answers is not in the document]
gfodor: haha yeah, this forced me to push most of the logic out of the channel and into the schemas (which are easier to test) so I guess it was a good move :)

lib/ret/hub.ex
```diff
@@ -16,7 +16,7 @@ defmodule Ret.Hub do
   use Bitwise
 
   @schema_prefix "ret0"
-  @primary_key {:hub_id, :integer, []}
+  @primary_key {:hub_id, :id, autogenerate: true}
```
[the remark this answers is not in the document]
brianpeiris: Yeah, but I thought that's what the `:returning` option on `insert` was for. https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2-options

Seems odd to have Ecto seemingly autogenerate here and then define a default `next_id()` fragment in the migration.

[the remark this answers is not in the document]
gfodor: oh, i'm pretty sure this isn't instructing Ecto about autogeneration, its *telling* Ecto that this field is autogenerated by the database (which is true because of `next_id()`, but would also be true if we say used `serial` or another method to generate the id via a sequence) so by having this Ecto will know to implicitly include it as a `RETURNING` column in all relevant queries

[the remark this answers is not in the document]
gfodor: in my testing, making this change does what I'd expect, `next_id()` generates the new value and then that value actually shows up in the returned data structure.

lib/ret_web/auth_channel.ex
gfodor: we're currently using `plug_attack` to rate limit certain requests by IP, but that doesn't seem to expose a generic ETS based rate limiting API. It seems like we should explore this alternative which would let us rate limit HTTP requests or anything else:

https://github.com/ExHammer/hammer

I'll create an issue for it

gfodor:
Superceded by https://github.com/mozilla/reticulum/pull/62

PR #62, opened 2018-09-06
Feature/scene api auth and uploads

gfodor:
This PR:

- Renames a bunch of concepts. "Uploads" is no longer really a thing outside of the media API. Instead we have a "Storage API" (internal API, at `Ret.Storage`) and files in storage are referred to as "stored files." The "Upload API" end point is now the "File API" (so instead of reading files from `/upload/:id` you now get them from `/files/:id`. The motivation for the rename is (in part) due to the fact that there will likely be stored files that are not direct uploads. For example, when we remix scenes, we'll make copies of existing files and assign the copies to a different account.

- When uploading to the Media API, it now returns a `file_id` which can be used to specify files when updating scenes via the scene API.

- Stored files can be in one of two states, expiring (the default state when uploaded via the Media API) and "owned". A file is "promoted" from expiring to owned. When a stored file is owned, it is associated with an account, and has its metadata and encryption key stored in the database. This is kept in the `owned_files` table and the new `Ret.OwnedFile` schema is used to access the table.

- Owned files can be accessed by the File API without an encryption key (though as usual the stored file is encrypted on disk.) The encryption algorithm has been changed slightly to hash + salt the key since now the key is stored in the database. This will make it so an attacker who compromises the system and gains access to the files and the database will *also* need the reticulum secret key in order to decrypt the files, etc.

- Cleans up and implements the scene schema and API (`GET`, `POST`, and `PATCH`.) The API requires a JWT bearer auth header for mutations and also implicitly promotes the scene's relevant files to be owned by the logged in account (if they haven't already been promoted.) 

- Added a bunch of tests (yay)


habitat/config/ret.prod.conf
```diff
@@ -29,9 +29,9 @@ ret.Elixir.Ret.MediaResolver.imgur_client_id = "{{ cfg.resolver.imgur_client_id
 ret.Elixir.Ret.MediaResolver.google_poly_api_key = "{{ cfg.resolver.google_poly_api_key }}"
 ret.Elixir.Ret.MediaResolver.sketchfab_api_key = "{{ cfg.resolver.sketchfab_api_key }}"
 ret.Elixir.Ret.PageOriginWarmer.page_origin = "{{ cfg.pages.page_origin }}"
-ret.Elixir.Ret.Uploads.storage_path = "{{ cfg.uploads.storage_path }}"
-ret.Elixir.Ret.Uploads.ttl = {{ cfg.uploads.ttl }}
-ret.Elixir.Ret.Uploads.host = "{{ cfg.uploads.host }}"
+ret.Elixir.Ret.Storage.storage_path = "{{ cfg.uploads.storage_path }}"
```
gfodor: note `uploads` here is still burned in to the habitat toml configs, we'll probably want to rename these at some point but didn't want the ops creep atm

brianpeiris:
Probably makes sense to delete my adhoc "test/uploads/*" files in this PR since you have actual tests now.

lib/ret_web/controllers/api/v1/scene_controller.ex
```diff
@@ -1,29 +1,71 @@
...
-  alias Ret.Scene
-  alias Ret.Repo
+  alias Ret.{Repo, Scene, Storage}
 
-  plug(RetWeb.Plugs.RateLimit when action in [:create])
+  plug(RetWeb.Plugs.RateLimit when action in [:create, :update])
 
-  def show(conn, %{"id" => scene_id}) do
-    scene = Repo.get(Scene, scene_id)
+  def show(conn, %{"id" => scene_sid}) do
```
brianpeiris: Are we back to using SIDs in APIs?

priv/repo/migrations/20180904222223_create_scenes_table.exs
```diff
@@ -8,15 +8,15 @@ defmodule Ret.Repo.Migrations.CreateScenesTable do
       add(:slug, :string, null: false)
       add(:name, :string, null: false)
       add(:description, :string)
-      add(:author_account_id, :bigint, null: false)
-      add(:model_upload_id, :bigint, null: false)
-      add(:screenshot_upload_id, :bigint, null: false)
-      add(:attribution_name, :string, null: false)
-      add(:attribution_link, :string)
```
brianpeiris: Leaving out attribution for now?

lib/ret/owned_file.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Ret.OwnedFile do
+  use Ecto.Schema
+  import Ecto.Changeset
+  alias Ret.{OwnedFile, Account}
+
+  @schema_prefix "ret0"
+  @primary_key {:owned_file_id, :id, autogenerate: true}
+
+  schema "owned_files" do
+    field(:owned_file_sid, :string)
```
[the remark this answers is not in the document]
gfodor: ok. i'm cool with that. the way i've used "sid" in the past was pretty generic -- "string id" -- I can refactor these to be uuid though.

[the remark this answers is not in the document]
gfodor: if so i'll look into storing these as actual `uuid` types in postgres

lib/ret/storage.ex
```diff
@@ -0,0 +1,235 @@
...
+              {:error, :not_found}
+          end
+
+        :error ->
+          {:error, :not_found}
+      end
+    else
+      _ -> {:error, :not_allowed}
+    end
+  end
```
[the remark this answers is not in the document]
gfodor: ah thanks! i didn't know how to collapse this

lib/ret_web/controllers/api/v1/scene_controller.ex
gfodor: I forgot if we had a discussion about it, but I was just matching the current use of sids for hub objects

priv/repo/migrations/20180904222223_create_scenes_table.exs
gfodor: yeah, I took this out since I wasn't sure where we'll land for this. if spoke is more about mixing existing web content i'm thinking there will be a list of attributions we have to crawl out of the GLTFs involved

lib/ret_web/controllers/api/v1/scene_controller.ex
brianpeiris: Okay, yeah, I'm was also waffling on it, so let's just stick with SIDs and UUIDs (i.e. never expose DB IDs in either end-user facing or API endpoints).

lib/ret/owned_file.ex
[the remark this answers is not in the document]
gfodor: actually i think that could get pretty brittle. i'll leave them as string fields but will change the naming convention

PR #70, opened 2018-10-05
Feature/spoke landing

gfodor:
This adds the bits to proxy the spoke landing page


lib/ret_web/controllers/page_controller.ex
```diff
@@ -46,6 +46,10 @@ defmodule RetWeb.PageController do
     end
   end
 
+  def render_for_path("/spoke", conn) do
```
brianpeiris: Does this cover `/spoke/` as well? Seems not if we have a route for `/link/` above.

gfodor: ah the `/link/` thing is kind of a hack, i put that there because of some incidental stuff around the way we're using S3 as a web redirector for `hub.link` -- the only way that can work is if it ends up redirecting with the trailing slash. (hit hub.link to see what I mean)

as is this PR will not result in the page loading if you hit `/spoke/` but I think that's ok since I don't think we'll be linking to that URL.

brianpeiris: I would argue that eventually someone, for example, doing a write up in marketing will eventually add a trailing slash, so we should just support it.

gfodor: good point, will do

PR #72, opened 2018-10-09
Feature/ret channel

gfodor:
This PR adds a new "ret" channel which is connected to by all clients in the cluster, and is used to calculate total CCU and active room counts via presence. The channel is write-only, presence updates are not sent out to others connected.

As part of this, re-enables the CCU stats job and no longer tries to use undocumented genserver APIs for connecting to presence.


gfodor:
client: https://github.com/mozilla/hubs/pull/617

PR #75, opened 2018-10-14
Short permalinks

gfodor:
This PR tries to put things in a bit of a saner place wrt sids and permalinks. The goal is to:

- Enable permalinks via hub.link
- Keep them relatively short

Currently, the invite dialog in the client shares `hub.link/entry code` links, which unintuitively expire after 24h. It seems like a good idea to just use entry codes for the case of "go to hub.link directly, type in entry code" since having those expire is OK since it's likely you are getting that information verbally or due to an immediate need, whereas any URL we share ideally should not expire (unless the user wants it to, which is something we may add later.)

This PR:

- Changes the sid generation to 7 alphanumeric chars for a keyspace of 78364164096. (This may not be enough in the long run, but for now is probably fine given the small number of rooms.)

- Drops the leading "0" in hub sids that are fixed -- this was there to future proof the key space for URL redirects but I think we can figure out a better solution if/when the time comes. (Better to have shorter URLs.)

- Updates the /link/ redirector to expect hub sids. You can also include the slug at the end of the redirect link, which is used in the client PR for the longer permalink. (So eg `https://hub.link/acD93Ad/gregs-room`) edit: Upon reflection I'm abandoning this in the client PR -- we should only show the slug for rooms that have custom names I think, but will leave that to a future PR

- Note that hub.link/`entry code` URLs will no longer work, but they expired in 24h anyway.

The net effect is now permanent links to rooms can take on the form of `hub.link/Ab83Dcq` for example, which will redirect to the full canonical URL.

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -29,11 +29,16 @@ defmodule RetWeb.PageController do
   def render_for_path("/link", conn), do: conn |> render_page("link")
   def render_for_path("/link/", conn), do: conn |> render_page("link")
 
-  def render_for_path("/link/" <> entry_code, conn) do
+  def render_for_path("/link/" <> hub_sid_and_slug, conn) do
+    hub_sid = hub_sid_and_slug |> String.split("/") |> List.first()
+    conn |> redirect_to_hub_sid(hub_sid)
+  end
+
+  defp redirect_to_hub_sid(conn, hub_sid) do
     # Rate limit requests for redirects.
     :timer.sleep(500)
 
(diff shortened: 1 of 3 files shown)


lib/ret_web/controllers/page_controller.ex
```diff
@@ -61,6 +56,16 @@ defmodule RetWeb.PageController do
     |> send_resp(200, chunks)
   end
 
+  defp redirect_to_hub_sid(conn, hub_sid) do
+    # Rate limit requests for redirects.
+    :timer.sleep(500)
```
brianpeiris: We don't need to rate limit these any more.

PR #76, opened 2018-10-15
Feature/web push

gfodor:
This PR adds web push subscriptions. The push notifications are sent at most once a minute, and are sent to all subscribers when the room goes from 0 to 1 participants. The auth keys are encrypted using symmetric encryption.

Goes with: https://github.com/mozilla/hubs/pull/630

In testing, firefox works but I was having trouble getting Chrome to work. It might be a security issue due to hubs.local SSL context etc, so I was going to just test it live. Theoretically there should be no need to do anything special for Chrome vs FF.


lib/ret_web/channels/hub_channel.ex
```diff
@@ -128,6 +139,11 @@ defmodule RetWeb.HubChannel do
 
       send(self(), {:begin_tracking, socket.assigns.session_id, hub.hub_sid})
 
+      # Send join push notification if this is the first joiner
+      if Presence.list(socket.topic) |> Enum.count() == 0 do
+        Task.start_link(fn -> hub |> Hub.send_push_messages_for_join() end)
```
brianpeiris: Shouldn't the push be sent on the second person joining (`Presence.list(socket.topic) |> Enum.count() == 1`)?

brianpeiris: Ah, nvm, I get it now, we only send push notifications when the subscriber has closed the tab.

lib/ret/web_push_subscription.ex
```diff
@@ -0,0 +1,91 @@
...
+          _ -> :error
+        end
+      after
+        result -> result
+      else
+        error -> error
+      end
+
+      Statix.increment("ret.web_push.hub.sent", 1)
+      web_push_subscription |> changeset_for_notification_sent |> Repo.update!()
```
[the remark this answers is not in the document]
gfodor: ah yes I agree the metrics and last notified state should only be updated on success will fix. i'm not sure what you mean about `is_attempting_to_send` -- is the concern around concurrent joins happening in different rooms resulting in multiple sends within the rate limiting window? if that's what you meant it's a fair point, i hadn't considered that, though I am not sure if I want to introduce additional state to mitigate that atm since I expect that case to be very rare.

gfodor:
https://github.com/mozilla/reticulum/pull/78 merged into

PR #78, opened 2018-10-25
Feature/room objects

gfodor:
This PR adds the ability to persist room objects via `pin` and `unpin` hub channel message, and adds the ability to hit `/<hub id>/objects.gltf` to get the glTF rendering of these objects. The glTF is cached and also the data is encrypted at rest.


lib/ret/room_object.ex
```diff
@@ -0,0 +1,73 @@
...
+      |> Poison.encode!()
+
+    {:commit, gltf}
+  end
+
+  defp changeset(%RoomObject{} = room_object, %Hub{} = hub, attrs) do
+    room_object
+    |> cast(attrs, [:room_object_sid, :gltf_node])
+    |> unique_constraint(:room_object_sid)
+    |> unique_constraint(:hub_id)
```
brianpeiris: Reading up on unique constraints, should we be using the combined constraint name here? i.e. `unique_constraint(:room_objects_room_object_sid_hub_id_index)`
https://hexdocs.pm/ecto/Ecto.Changeset.html#unique_constraint/3-complex-constraints

lib/ret/room_object.ex
```diff
@@ -0,0 +1,73 @@
...
+        extensionsUsed: ["HUBS_components"]
+      }
+      |> Poison.encode!()
+
+    {:commit, gltf}
+  end
+
+  defp changeset(%RoomObject{} = room_object, %Hub{} = hub, attrs) do
+    room_object
+    |> cast(attrs, [:room_object_sid, :gltf_node])
```
brianpeiris: This is odd. Why are we accepting an `sid` from the client?

lib/ret/room_object.ex
```diff
@@ -0,0 +1,73 @@
...
+      else
+        0..((nodes |> length) - 1) |> Enum.to_list()
+      end
+
+    gltf =
+      %{
+        asset: %{version: "2.0", generator: "reticulum"},
+        scenes: [%{nodes: node_indices, name: "Room Objects"}],
+        nodes: nodes,
+        extensionsUsed: ["HUBS_components"]
```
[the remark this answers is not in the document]
gfodor: that makes sense. i can walk the nodes and extract the extensions list. i think i was mostly being lazy here.

lib/ret/room_object.ex
gfodor: ah, i'm using "sid" in the sense I originally meant it (string identifier) not the way we've been using it up to this point (X character, server generated identifier.) I can rename this field to reduce the confusion -- this is a client-generated, string, unique key for the object. (Incidentally the network id generated by network-aframe)

gfodor: i don't think we need a server-generated string-based surrogate key for these objects given the use-case -- we're already relying/assuming clients are generating these keys so the server is just going to now have a role in "enforcing" this for pinned objects by disallowing duplicates in the db. i didn't implement any error handing paths to gracefully handle dupes, the channel will just die and reconnect if they send "pin" for a non-unique id. (not the websocket connection, the channel connection)

gfodor: the new name i'll suggest using is `object_id`

lib/ret/room_object.ex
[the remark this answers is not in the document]
brianpeiris: To be clear, I think we can just remove the `extensionsUsed` property entirely.

lib/ret/room_object.ex
gfodor: interesting, TIL. i'm not sure what the effect of adding this to the changeset actually is, so I'll look into that too.

lib/ret/room_object.ex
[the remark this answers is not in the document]
gfodor: isn't this necessary? the `nodes` will include those extensions.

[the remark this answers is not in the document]
brianpeiris: We don't use it in Spoke. The GLTF loader uses this property to instantiate specialized extension classes for the handful of ratified extensions that it is aware of. Specifying `"HUBS_components"` or `"MOZ_hubs_components"` here will just result in a "Unknown extension" console warning.

lib/ret/room_object.ex
brianpeiris: Gotcha. `object_id` sounds okay. I'd suggest `client_object_id` maybe?

lib/ret/room_object.ex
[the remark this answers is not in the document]
gfodor: ah ok. i can remove for now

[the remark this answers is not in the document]
gfodor: ok -- this is simple enough to change if we need to.

lib/ret/room_object.ex
gfodor: I think I prefer `object_id` since once it's landed in our db, it's effectively a primary key for the object we may be passing around from that point forward to refer to it. it just incidentally is generated client side

PR #93, opened 2018-11-14
Handle unknown object type spawning

gfodor:
Fixes a bug causing: https://sentry.prod.mozaws.net/operations/reticulum/issues/4926378/?query=is:unresolved because we're not handling the 31'st bit case properly in telemetry.

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -72,7 +72,8 @@ defmodule Ret.Hub do
   def changeset_for_new_spawned_object_type(%Hub{} = hub, object_type)
       when object_type in 0..31 do
     # spawned_object_types is a bitmask of the seen object types
-    new_spawned_object_types = hub.spawned_object_types ||| 1 <<< object_type
+    <<new_spawned_object_types::integer-signed-32>> =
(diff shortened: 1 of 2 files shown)


lib/ret/hub.ex
```diff
@@ -72,7 +72,8 @@ defmodule Ret.Hub do
   def changeset_for_new_spawned_object_type(%Hub{} = hub, object_type)
       when object_type in 0..31 do
     # spawned_object_types is a bitmask of the seen object types
-    new_spawned_object_types = hub.spawned_object_types ||| 1 <<< object_type
+    <<new_spawned_object_types::integer-signed-32>> =
+      <<hub.spawned_object_types ||| 1 <<< object_type::integer-signed-32>>
```
brianpeiris: Took me a while to realize what you were doing here. Do we really want to store overflowed values in the DB, instead of just making the field a `bigint`? 
I would add a comment here at the very least.

gfodor: i'll take a pass at converting the column type -- i was a bit nervous about it since it's a a change to the telemetry table but shouldn't be too bad

gfodor:
closing in favor of https://github.com/mozilla/reticulum/pull/113

PR #96, opened 2018-11-14
Persistence ownership

brianpeiris:
- RoomObjects now have Accounts.
- Pin action now requires authentication.

Goes with https://github.com/mozilla/hubs/pull/739


lib/ret_web/channels/hub_channel.ex
```diff
@@ -93,9 +93,10 @@ defmodule RetWeb.HubChannel do
     {:reply, {:ok, %{has_remaining_subscriptions: has_remaining_subscriptions}}, socket}
   end
 
-  def handle_in("pin", %{"id" => object_id, "gltf_node" => gltf_node}, socket) do
+  def handle_in("pin", %{"id" => object_id, "gltf_node" => gltf_node, "token" => token}, socket) do
+    {:ok, %Account{} = account, _claims} = Guardian.resource_from_token(token)
     hub = socket |> hub_for_socket
-    RoomObject.perform_pin!(hub, %{object_id: object_id, gltf_node: gltf_node})
+    RoomObject.perform_pin!(hub, account, %{object_id: object_id, gltf_node: gltf_node})
```
gfodor: is there a reason you didn't want to create a separate `auth` message that associates the user with the socket? I think this is fine for now since we only have one authenticated message but I'm guessing that's what we'll want eventually

priv/repo/migrations/20181113033726_add_account_to_room_objects.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Ret.Repo.Migrations.AddAccountToRoomObjects do
+  use Ecto.Migration
+
+  def change do
+    alter table("room_objects") do
+      add(:account_id, references(:accounts, column: :account_id))
```
gfodor: I suggest we make this `NOT NULL` and either:

- Nuke all the existing pinned objects (debatable, probably at this point a bad idea)
- Set all the legacy ones to account id of 0

The `NOT NULL` will help reduce the risk of bugs causing privacy holes since if we fail to capture that account id properly there is no way to purge the data if you lose the URL

gfodor:
Looking good, one thing missing here is we need to extend the pin API to also promote files to become owned. I think that's going to require adding arguments similar to the scene REST API and also threading that metadata through in the networked object state for media (so unfortunately non trivial.) If that proves difficult, we can also fish it out of the URL on the server side, but that's going to be potentially more brittle.

gfodor:
Oh, and we also need to make sure we address this issue (or verify we're already doing something sane, which I doubt):

https://github.com/mozilla/reticulum/issues/80

gfodor:
(also if you prefer to roll this out and then do a second pass on uploaded files that's cool w me)

gfodor:
I think the API we need is probably something like `ensure_owned` that gets in a file uuid and the encryption key and if its not already an owned file then associates it with the account. and then the client should just fire a bunch of those in addition to the pin for media objects with uploaded content.

lib/ret_web/channels/hub_channel.ex
brianpeiris: Ah, Initially I bound the creds to the socket via `socketSettings` on the client side and then used `Guardian.Phoenix.Socket.authenticate(socket, ...)` on connect, but realized that I'd have to re-connect the socket if the user decided to sign out, which would obviously require them to re-join the room.
I hadn't considered a separate auth message and re-associating the socket. I'll try that.

gfodor: if you have a sign_in and sign_out message you can just update the `socket.assigns` to include the `account_id` I think

gfodor: and you'll need to also handle it in `join` as well, so rejoins will work -- you'll just need to have the token passed in optionally

gfodor: this is common enough i'd also just look at the phoenix examples out there -- i think there's a happy path here with the socket key being the user id. i am not sure if we want to go that route yet, since it will have an effect on presence.

brianpeiris:
Superseded by #97. Review changes also applied to #97.

PR #97, opened 2018-11-27
Persistence ownership and pinned upload promotion

brianpeiris:
- RoomObjects now have Accounts.
- Pin action now requires authentication.
- Adds a `promotion_token` field to file meta data
- `pin` message now provides optional `file_id`, `file_token` and `promotion_token` fields when pinning uploads
- pinned uploads are promoted if the promotion token is verified
- Adds a new `:inactive` state to owned files
- `unpin` marks owned files as inactive if unpinned by owner
- Adds a new scheduled job that demotes inactive files
- Adds a `min_token_issued_at` to Account which can be used to revoke all tokens issued before a given date.

Goes with https://github.com/mozilla/hubs/pull/759

Supersedes #96


lib/ret/owned_file.ex
```diff
@@ -28,4 +29,19 @@ defmodule Ret.OwnedFile do
...
   end
+
+  def set_inactive(owned_file_uuid, account_id) do
+    OwnedFile
+    |> where(owned_file_uuid: ^owned_file_uuid, account_id: ^account_id)
+    |> Repo.one()
+    |> set_inactive
+  end
+
+  def set_inactive(nil), do: nil
```
gfodor: `defp`? and maybe the one below?

lib/ret/storage.ex
```diff
@@ -124,6 +129,25 @@ defmodule Ret.Storage do
     end
   end
 
+  # If an owned file does not have a promotion token associated with it, it can be promoted with any given 
+  # promotion token, including nil.
+  defp check_promotion_token(nil, _promotion_token), do: {:ok}
+
+  defp check_promotion_token(actual_promotion_token, promotion_token) do
```
gfodor: you might be able to break this up into a few functions using guard clauses (`when actual_promotion_token == promotion_token` etc). nbd though

lib/ret/storage.ex
```diff
@@ -159,6 +183,20 @@ defmodule Ret.Storage do
     Logger.info("Stored Files: Vacuum Finished.")
   end
 
+  def demote_inactive_owned_files do
+    inactive_owned_files =
```
gfodor: perhaps this query should be returned from a function `OwnedFile.inactive`?

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -11,20 +11,24 @@ defmodule RetWeb.Api.V1.MediaController do
...
+  def create(conn, %{
+        "media" => %Plug.Upload{content_type: content_type} = upload,
+        "with_promotion_token" => with_promotion_token
+      }) do
+    render_upload(conn, upload, content_type, with_promotion_token)
   end
 
-  defp render_upload(conn, %Plug.Upload{} = upload, content_type) do
+  defp render_upload(conn, %Plug.Upload{} = upload, content_type, with_promotion_token) do
     token = SecureRandom.hex()
```
gfodor: `token` -> `access_token`

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -11,20 +11,24 @@ defmodule RetWeb.Api.V1.MediaController do
   end
 
   def create(conn, %{
-        "media" =>
-          %Plug.Upload{filename: filename, content_type: "application/octet-stream"} = upload
+        "media" => %Plug.Upload{filename: filename, content_type: "application/octet-stream"} = upload,
+        "with_promotion_token" => with_promotion_token
```
gfodor: don't have a strong opinion, but maybe worth doing `promotion_mode: [always|with_token]` (leaving the option open for `never`)

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -11,20 +13,31 @@ defmodule RetWeb.Api.V1.MediaController do
   end
 
   def create(conn, %{
-        "media" =>
-          %Plug.Upload{filename: filename, content_type: "application/octet-stream"} = upload
+        "media" => %Plug.Upload{filename: filename, content_type: "application/octet-stream"} = upload,
+        "with_promotion_token" => "true"
```
gfodor: github swallowed my comment -- perhaps `promotion_mode: [always|token]` (leaving option open for `never` eventually?)

gfodor:
lgtm! minor comments

PR #102, opened 2018-12-27
Feature/janus load balancing

gfodor:
Adds the facility to dynamically assign Janus hosts to rooms:

- Introduces a cluster-global process that is responsible for Janus load balancing

- Adds a Cachex warmer that keeps a local cached data structure with maps all healthy Janus nodes (as per the Habitat cansus) to their current CCU (as per the Janus admin API)

- Implements a simple weighted load balancing algorithm (at time of this writing, Janus nodes with > 600 CCU get weight 1, > 300 CCU weight 50, otherwise weight 500)

- When a hub channel is joined, the room (re-)assignment is performed via the gen server

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -27,6 +27,7 @@ defmodule Ret.Hub do
   schema "hubs" do
     field(:name, :string)
     field(:hub_sid, :string)
+    field(:host, :string)
     field(:entry_code, :integer)
     field(:entry_code_expires_at, :utc_datetime)
     field(:default_environment_gltf_bundle_url, :string)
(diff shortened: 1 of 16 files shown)


lib/ret/hub.ex
```diff
@@ -123,6 +128,20 @@ defmodule Ret.Hub do
...
+  def ensure_room!(hub) do
+    if RoomAssigner.is_alive?(hub.host) do
+      hub
+    else
+      host = RoomAssigner.get_available_host(hub.host)
+
+      if host && host != hub.host do
+        hub |> changeset_for_new_host(host) |> Repo.update!()
+      else
+        hub
```
brianpeiris: Maybe we should actually throw in this case, since this is `ensure_room!`?

gfodor: ah actually I think we should fall through as a worst-case-scenario type of situation (perhaps the supervisor census is dead, but janus is still alive) but we can probably drop the `!` from the name

PR #105, opened 2019-01-09
Implement basic Discord bot functionality

mquander:
Includes basic chat mirroring and presence functionality.

Infrastructurally, the goal for now is:

- Attempt to start a `DiscordBotManager` as a subprocess on each Reticulum node.
- The job of `DiscordBotManager` is to launch a `DiscordBot` iff no `DiscordBot` is running yet in the cluster.
- `DiscordBot` supervises the Alchemy Discord API client and bound channel state.
- On initialization failure of any of the above, we just give up and run the rest of Reticulum on that node with no `DiscordBotManager`.
- On runtime failure of any of the above, it tries to restart itself, or perhaps another node's `DiscordBotManager` will take over. I think whichever one gets there first will get it started.

Whether we achieve these goals, uh, well, maybe we do. I can't prove we don't.

If you want to run this locally, you'll need to stick a Discord bot API token in your dev config. You can find the tokens in the Ansible config vault.


lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,226 @@
...
+      RetWeb.Endpoint.broadcast!("hub:#{hub_id}", "message", %{
+        type: "chat",
+        body: msg.content,
+        from: "#{msg.author.username}##{msg.author.discriminator}"
+      })
+    end
+
+    ###
+
+    @hub_url_regex ~r"https?://(?:hubs.local(?:\:\d+)?|hubs.mozilla.com)/(\w+)/?\S*"
```
gfodor: This should be changed to use `RetWeb.Endpoint.url`

lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,226 @@
+defmodule DiscordBot do
+  use Supervisor
+  alias Alchemy.Client
+  alias Alchemy.Embed
+
+  defmodule BotState do
```
gfodor: This `GenServer` may be better replaced by just using `Cachex` which we are using for other in-memory caching. You can register a new cache (which will incidentally start a new genserver for it) and then just read/write to it via the cache APIs.

gfodor: https://github.com/whitfin/cachex

lib/discord_bot/discord_bot_manager.ex
```diff
@@ -0,0 +1,54 @@
...
+  # managed process exited with an error
+  def handle_info({:DOWN, _, :process, pid, _reason}, state = %State{pid: pid}) do
+    {:noreply, restart(state)}
+  end
+
+  # it's incredibly unclear to me what errors in the discord bot result in what control flow, but
+  # empirically, if there's a problem when starting the bot, sometimes an error will fly out of
+  # DiscordBotManager.start_link, and sometimes it will fly out of DiscordBot.start_link, so
+  # we take care to handle both. the goal in both cases is to just give up and not be running the
+  # bot, with the theory that we can restart reticulum if we want to try again.
```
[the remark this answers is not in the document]
gfodor: is there a reason we don't want to just crash this process if we get a legit failure to start the discord bot and have the supervisor keep re-running `restart` until it resolves a `pid`?

lib/discord_bot/discord_bot_manager.ex
```diff
@@ -0,0 +1,54 @@
...
+      other -> other
+    end
+  end
+
+  defp restart(state) do
+    result = DiscordBot.start_link([], [name: {:global, DiscordBot}])
+    pid = case result do
+      {:ok, pid} -> pid
+      {:error, {:already_started, pid}} -> pid
+      {:error, {:shutdown, reason}} -> Logger.error "Failed to start Discord bot: #{inspect(reason)}"; nil
```
gfodor: it also seems like we should let the manager crash here

lib/discord_bot/discord_bot.ex
mquander: Much better!

PR #107, opened 2019-01-11
Add APIs in the hub channel to allow scene + name changes to hubs

gfodor:
This adds two new hub channel messages, `update_scene` and `update_hub`. `update_scene` expects a URL to a scene (a Spoke uploaded scene, or a media-resolvable URL to a GLTF/GLB) and updates the database appropriately. `update_hub` allows updating the other aspects to a hub, currently just the name/slug, but eventually likely other user-modifiable attributes of the hub.

additionally, whenever a the state about a hub is changed or needs to be refreshed (such as a scene reload) channel members will get a `hub_refresh` message over the channel, which also includes a list of the stale fields that need to be dealt with. eg if `scene` is a stale scene it means the scene needs to be reloaded/changed and the UI updated, and if the `name` is stale it means the UI needs to be refreshed and the browser location URL replaced with the new slug.

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -1,5 +1,5 @@
 defmodule Ret.Hub.HubSlug do
-  use EctoAutoslugField.Slug, from: :name, to: :slug
+  use EctoAutoslugField.Slug, from: :name, to: :slug, always_change: true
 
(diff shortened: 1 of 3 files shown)


brianpeiris:
Will merge with #110

PR #108, opened 2019-01-11
Feature/room ownership

brianpeiris:
- Adds [Canary](https://github.com/cpjk/canary) and [Canada](https://github.com/jarednorman/canada), lightweight authorization libraries.
- Adds an optional authentication pipeline to support both anon and authenticated API calls
- Adds `account_id` to Hubs
- Authorizes room owners to change a room name
- Hub channel join payload now accepts an optional auth token. Join response includes an `is_owner` field.
- Adds REST APIs and tests for changing room names


lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -9,6 +11,24 @@ defmodule RetWeb.Api.V1.HubController do
   # Only allow access with secret header
   plug(RetWeb.Plugs.HeaderAuthorization when action in [:delete])
 
+  plug(:authorize_resource, model: Hub, id_field: "hub_sid", only: [:update])
+
+  def update(conn, %{"id" => hub_sid, "hub" => params}) do
```
brianpeiris: Hadn't realized that you'd want to do name changes as phoenix messages in #107. I can remove this or leave it in if we want the REST API anyway.

lib/ret/hub.ex
```diff
@@ -37,6 +37,7 @@ defmodule Ret.Hub do
     field(:entry_mode, Ret.Hub.EntryMode)
     belongs_to(:scene, Ret.Scene, references: :scene_id)
     has_many(:web_push_subscriptions, Ret.WebPushSubscription, foreign_key: :hub_id)
+    belongs_to(:account, Ret.Account, references: :account_id)
```
gfodor: not totally sure, but we might want to consider naming this `created_by_account_id` since hubs are going to have a variety of semantic relationships to users -- eg there's already the concept of "owner" and "creator" implicitly so naming it this way would make it clear what exactly is being recorded

lib/ret/hub.ex
```diff
@@ -62,6 +62,19 @@ defmodule Ret.Hub do
     |> HubSlug.unique_constraint()
   end
 
+  def changeset_for_new_name(%Hub{} = hub, attrs) do
```
gfodor: I already ran into this on my branch, but this also needs to regenerate the slug and you need to flip a setting on the slug module (see my PR)

gfodor: most likely you should just pull/merge the implementation of this bit from my branch

lib/ret/hub.ex
```diff
@@ -62,6 +62,19 @@ defmodule Ret.Hub do
...
+    hub
+    |> Ecto.Changeset.change()
+    |> add_name_to_changeset(attrs)
+  end
+
+  defp add_name_to_changeset(changeset, attrs) do
+    changeset
+    |> cast(attrs, [:name])
+    |> validate_length(:name, min: 4, max: 64)
```
gfodor: I forgot to do this in my PR!

lib/ret_web/channels/hub_channel.ex
```diff
@@ -18,11 +27,11 @@ defmodule RetWeb.HubChannel do
     socket |> assign(:profile, profile) |> assign(:context, context) |> perform_join(hub_sid)
   end
 
-  defp perform_join(socket, hub_sid, push_subscription_endpoint \\ nil) do
+  defp perform_join(socket, hub_sid, push_subscription_endpoint \\ nil, token \\ nil) do
```
gfodor: we probably want to refer to tokens consistently o/w it is gonna get confusing:

- `access_token` - I think this is what we're using for the file upload bits
- (proposed) `auth_token` - Token for authentication
- (proposed) `perms_token` - Token for authorization (not relevant yet)

lib/ret_web/controllers/api/v1/hub_controller.ex
gfodor: Let's remove for now, since I can't think of a reason right now to have it

brianpeiris:
Will merge with #110

PR #109, opened 2019-01-14
Un-revert, fix Discord bot

mquander:
This restores the Discord bot and fixes the problem where incoming messages and events from Hubs would not be correctly handed to the singleton Discord bot process.

As of the time of this writing, this should work, but there's still some mystery web socket connectivity problem when this is running in AWS that needs to be tracked down before the bot is usable.


lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,188 @@
...
+      def broadcast_to_hubs(hub_id, msg) do
+        RetWeb.Endpoint.broadcast!("hub:#{hub_id}", "message", %{
+              type: "chat",
+              body: msg.content,
+              from: "#{msg.author.username}##{msg.author.discriminator}"
+                                   })
+      end
+
+      def update_bound_channels do
+        hostnames = Application.get_env(:ret, Elixir.DiscordBot)[:hostnames] |> String.split
```
gfodor: I'd suggest we fall through to using `RetWeb.Endpoint.host` if the config is not set, and then just don't bother setting it in prod (unless you think we can get away without `localhost` in dev, in which case this config isn't needed at all then I think.)

gfodor: also there's a pattern I started using for grabbing module configs: 

https://github.com/mozilla/reticulum/blob/master/lib/ret/media_resolver.ex#L349

and then using with to handle case where not defined:

https://github.com/mozilla/reticulum/blob/master/lib/ret/media_resolver.ex#L329

lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,188 @@
...
+          matches = Regex.scan(hub_url_regex, topic),
+          [_, hub_id] <- matches do
+          {hub_id, channel.id}
+        end
+
+        channels_by_hub = bound_channels |> Enum.group_by(&elem(&1, 0), &elem(&1, 1))
+        hubs_by_channel = bound_channels |> Enum.group_by(&elem(&1, 1), &elem(&1, 0))
+
+        Logger.debug "Bound Discord channels per hub: #{inspect(channels_by_hub)}"
+        Cachex.transaction!(:discord_bot_state, ["hubs_for_channel", "channels_for_hub"], fn(cache) ->
```
gfodor: i'd suggest using atoms not strings for these keys if possible

lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,188 @@
...
+    end
+
+    def init(_args) do
+      {:ok, nil}
+    end
+
+    def start_link() do
+      # conventionally, we will only have one of these running at once on the cluster,
+      # since the supervisor that starts it is one-per-cluster (enforced by DiscordBotManager)
+      result = GenServer.start_link(__MODULE__, [], [name: {:global, DiscordBot}])
```
gfodor: It seems unnecessary to use `:global` here -- and I think in general avoiding the use of `:global` is a good idea because it results in cluster-wide side effects. Instead, for a local, named, singleton process I think it's idiomatic to use Elixir's process registry:

https://hexdocs.pm/elixir/master/Registry.html

lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,188 @@
...
+      with {:ok, _} <- result do
+        use HubsCommands
+        use HubsEvents
+      end
+      result
+    end
+
+    def handle_cast(data, state) do
+      %{hub_sid: hub_sid} = data
```
gfodor: this can be destructured in the function definition as `handle_case(%{hub_sid: hub_sid} = data,...`

lib/discord_bot/discord_bot.ex
```diff
@@ -0,0 +1,188 @@
...
+      if channel_ids = Cachex.get!(:discord_bot_state, "channels_for_hub")[hub_sid] do
+        %{event: event, context: %{:profile => %{"displayName" => username}}} = data
+        case event do
+          :join ->
```
gfodor: or `e when e == :join or e == :part` if you want to DRY the `Enum.each`

gfodor:
lgtm, would consider moving from `:global` to `Process.Registry` for the local bot process management, minor comments

lib/discord_bot/discord_bot_manager.ex
```diff
@@ -0,0 +1,59 @@
...
+  def start_link() do
+    result = GenServer.start_link(__MODULE__, [], name: __MODULE__)
+    case result do
+      {:error, e} -> Logger.error "Failed to start Discord bot: #{inspect(e)}"; :ignore
+      other -> other
+    end
+  end
+
+  defp restart(state) do
+    result = DiscordSupervisor.start_link([], [name: {:global, DiscordSupervisor}])
```
gfodor: this makes way more sense to me!

gfodor:
OK I think I understand better now. We register two global process pids, one global process is the singleton supervisor, but then the child gen server registers itself globally as well. We need the second global registration because that's the way we send messages from the local manager genserver over to the bot process (where ever it is on the cluster.)

Because of that, I don't think we need to stash away the first global pid into the DiscordBotManager. Presumably that's there so you can look up the registered process by the manager locally, but since we don't do it then we may as well not have it stored (if for no other reason than it's confusing that we keep it in state and don't consume it.)

lib/discord_bot/discord_bot.ex
gfodor: nevermind, see comment below.

gfodor:
question: if the bot process is present on the cluster, but for some reason the supervisor process isn't, what happens?

that condition should be rare/impossible, but it would be good to know if the ideal behavior occurs or not: the supervisor should continually retry spawning the bot process as a global process, until it can. (this may hit the "if a supervisor can't start its children, it fails out completely" case you've been running into though.)

lib/discord_bot/discord_bot.ex
mquander: I prefer setting this explicitly, because each environment is associated with multiple hostnames. For example, production is associated with both hubs.mozilla.com and smoke-hubs.mozilla.com.

mquander: > also there's a pattern I started using for grabbing module configs:

I explicitly did not use this pattern because the Discord configuration is used by two different modules -- `DiscordSupervisor` and `DiscordBot`, and I didn't want to split it into separate configuration sections.

lib/discord_bot/discord_bot.ex
mquander: Agreed.

lib/discord_bot/discord_bot.ex
mquander: I prefer the way I wrote it because I dislike destructuring some stuff in the function definition and other stuff in the body. If you feel strongly that your way is more idiomatic Elixir style I'll change it.

lib/discord_bot/discord_bot.ex
mquander: I agree, I will change it.

mquander:
> OK I think I understand better now. We register two global process pids, one global process is the singleton supervisor, but then the child gen server registers itself globally as well. We need the second global registration because that's the way we send messages from the local manager genserver over to the bot process (where ever it is on the cluster.)

Yep, that's right.

I think a good Elixir programmer would have sent messages to and from the supervisor, but I am too tired of trying to debug Elixir IPC stuff to get there unless there is some reason besides cleanliness.

> Because of that, I don't think we need to stash away the first global pid into the DiscordBotManager. Presumably that's there so you can look up the registered process by the manager locally, but since we don't do it then we may as well not have it stored (if for no other reason than it's confusing that we keep it in state and don't consume it.)

That's true, I'll get rid of it if you prefer.

mquander:
> question: if the bot process is present on the cluster, but for some reason the supervisor process isn't, what happens?

that condition should be rare/impossible, but it would be good to know if the ideal behavior occurs or not: the next successfully registered bot supervisor should continually retry spawning the bot process as a global process, until it can. (this may hit the "if a supervisor can't start its children, it fails out completely" case you've been running into though.)

I definitely assumed that would be impossible. The bot is only started by the supervisor, so the only way it could happen is if the supervisor terminated, but the bot continued.

I will investigate a little bit into whether this is possible and whether the robustness you described is working or is easily attainable, because I agree that this would be a hairy kind of failure.

lib/discord_bot/discord_bot.ex
gfodor: Ah that makes sense, because we share the cluster with smoke `RetWeb.Endpoint.host` varies intra-cluster, which seems fairly problematic now that we're trying to do fancier things like this.

Re: configs, it's probably OK either way, I just wanted to point out the other pattern. I think using a different module's name for the configs is not a big deal.

lib/discord_bot/discord_bot.ex
gfodor: yeah doing it the way I suggest is more idiomatic afaik

gfodor:
> > Because of that, I don't think we need to stash away the first global pid into the DiscordBotManager. Presumably that's there so you can look up the registered process by the manager locally, but since we don't do it then we may as well not have it stored (if for no other reason than it's confusing that we keep it in state and don't consume it.)
> 
> That's true, I'll get rid of it if you prefer.

I think we should remove it since it confused me into thinking we were tracking that process pid for a reason.

lib/discord_bot/discord_bot.ex
mquander: I changed my mind. I will fix my aesthetics to prefer your way because the pattern of always-destructure-as-much-as-possible-in-the-type-signature reduces nesting in many cases.

mquander:
OK, I'm optimistic this will work in terms of not crashing everything, so I'm going to merge this and then muck around on smoke to evaluate any further connectivity-to-Discord-API problems.

PR #114, opened 2019-01-31
Feature/kick

brianpeiris:
Adds `get_perms_token` and `kick` messages to `hub_channel`

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -1,6 +1,7 @@
 use Mix.Config
 
-host = "hubs.local"
+host = "10.252.26.69"
+dev_janus_host = "dev-janus.reticulum.io"
 
 # For development, we disable any cache and enable
 # debugging and code reloading.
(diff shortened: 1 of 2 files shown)


config/dev.exs
```diff
@@ -1,6 +1,7 @@
 use Mix.Config
 
-host = "hubs.local"
+host = "10.252.26.69"
```
gfodor: probably a mistake

lib/ret_web/channels/hub_channel.ex
```diff
@@ -238,6 +238,23 @@ defmodule RetWeb.HubChannel do
     {:noreply, socket}
   end
 
+  def handle_in("get_perms_token", _args, socket) do
```
gfodor: i'm using `refresh_perms_token` on my branch for the global refresh one on `ret_channel`, i defer to you which name you prefer

PR #115, opened 2019-01-31
Feature/admin credentials

gfodor:
This PR:

- Adds an `is_admin` flag to user accounts
- Adds the necessary permissions to allow access to PostgREST to admin perms tokens
- Updates the `ret_channel` to have a way to refresh global perm tokens
- Adds routes for `/admin` (pending)

One question I had is we needed to add an `aud` to the perms token to properly enforce it on the PostgREST side, I am not sure if this matters but wanted to highlight it.

### DIFF
--- a/lib/ret/account.ex
+++ b/lib/ret/account.ex
@@ -9,6 +9,7 @@ defmodule Ret.Account do
 
   schema "accounts" do
     field(:min_token_issued_at, :utc_datetime)
+    field(:is_admin, :boolean)
     has_one(:login, Ret.Login, foreign_key: :account_id)
     has_many(:owned_files, Ret.OwnedFile, foreign_key: :account_id)
     has_many(:created_hubs, Ret.Hub, foreign_key: :created_by_account_id)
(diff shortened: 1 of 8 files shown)


lib/ret/account.ex
```diff
@@ -46,4 +47,10 @@ defmodule Ret.Account do
   def identifier_hash_for_email(email) do
     email |> String.downcase() |> Ret.Crypto.hash()
   end
+
+  def add_global_perms_for_account(perms, %Ret.Account{is_admin: true}) do
+    perms |> Map.put(:postgrest_role, :ret_admin)
```
brianpeiris: Arguably this could have been defined as a Canada rule that grants a `:administrate_postgrest` permission. That way we could use `can?` in other places if we wanted to have a unified way of defining and looking up perms, but I guess if we're just using this role for one thing, it's fine.

gfodor: ah yes, the reason it's done this way is because this is how PostgREST deals with security, this isn't used for reticulum -- you encode which PostgreSQL role you'd like it to "sudo" to by encoding a role name into the JWT like this.

gfodor: and specifically the way I've been thinking about this is that we have "reticulum administrators" and right now what that gives you access to is the ability to interact with PostgREST as a `ret_admin`. And that in turn gives you access to a set of views in the `ret0_admin` schema (which is assumed in order to use the hubs admin console.)

gfodor: also the new schema is defined here:

https://github.com/mozilla/hubs-ops/tree/master/db

the views are:

https://github.com/mozilla/hubs-ops/tree/master/db/ret_admin/sql

PR #116, opened 2019-02-08
Feature/scene listings

gfodor:
Adds `scene_listings` to reticulum and a `/media` search API to search for listings. A scene listing is a publicly visible, search indexed, tagged listing for a point-in-time snapshot of a scene. The `scene_listings` table is managed by the admin tools.

Scenes now have a `reviewed_at` timestamp that is used to keep track of the review workflow (in the admin tools) for scenes. (When a scene is approved or denied for listing, this timestamp is updated.) `hubs` now have either a `scene_id` or a `scene_listing_id`, and various points in the code have to consider both possibilities. If a room is created with a `scene_listing_id`, the actual scene information it is associated with will come from the listing, so it will not change until the latest version of a scene has been reviewed and approved.

However, the "scene listing" concept is not user facing. The URLs and ids for scenes and scene listings are the same, the scene listing for a given scene just has a different sid. So querying for scenes and scene listings via the `/scene` API is the same, and the scene landing page exists for both scenes and scene listings.

An small included change is that the enums in our db were living in `public` not `ret0`, so this PR moves them over properly.


lib/ret/media_search.ex
```diff
@@ -0,0 +1,118 @@
...
+
+  def search(%Ret.MediaSearchQuery{source: "scene_listings", page: page, filter: filter, q: query}) do
+    scene_listing_search(page, query, filter)
+  end
+
+  def search(%Ret.MediaSearchQuery{source: "sketchfab", user: user}) do
+    with api_key when is_binary(api_key) <- resolver_config(:sketchfab_api_key) do
+      res =
+        "https://api.sketchfab.com/v3/search?type=models&downloadable=true&user=#{user}"
+        |> retry_get_until_success([{"Authorization", "Token #{api_key}"}])
```
brianpeiris: We could limit poly size with a filter here.

lib/ret/media_search.ex
```diff
@@ -0,0 +1,118 @@
...
+    |> add_query_to_listing_search_query(query)
+    |> add_tag_to_listing_search_query(filter)
+    |> preload([:screenshot_owned_file, :model_owned_file, :scene_owned_file])
+    |> order_by(^order)
+    |> Repo.paginate(%{page: page, page_size: @page_size})
+    |> result_for_scene_listing_page()
+  end
+
+  def add_query_to_listing_search_query(query, nil), do: query
+  def add_query_to_listing_search_query(query, q), do: query |> where([l, s], ilike(l.name, ^"%#{q}%"))
```
brianpeiris: Will something in the ORM/Postgres do sanitization here? If not this looks like a potential SQL injection.

brianpeiris: Same for the tag query below, I think

lib/ret/media_search.ex
```diff
@@ -0,0 +1,118 @@
...
+      entries:
+        page.entries
+        |> Enum.map(&scene_listing_to_entry/1)
+    }
+  end
+
+  defp scene_listing_to_entry(scene_listing) do
+    %{
+      id: scene_listing.scene_listing_sid,
+      url: "#{RetWeb.Endpoint.url()}/scenes/#{scene_listing.scene_listing_sid}/#{scene_listing.slug}",
```
brianpeiris: Use `to_url` here?

lib/ret/media_search.ex
```diff
@@ -0,0 +1,118 @@
...
+      description: scene_listing.description,
+      attributions: scene_listing.attributions,
+      images: %{
+        preview: scene_listing.screenshot_owned_file |> OwnedFile.uri_for() |> URI.to_string()
+      }
+    }
+  end
+
+  defp sketchfab_api_result_to_entry(result) do
+    # TODO when missing thumbnails
```
brianpeiris: Punting this for now?

lib/ret/media_search.ex
brianpeiris: Also, I think these can be `defp`

priv/repo/migrations/20190114232257_add_reviewed_at_to_scenes.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Ret.Repo.Migrations.AddReviewedAtToScenes do
+  use Ecto.Migration
+
+  def change do
+    alter table("scenes") do
+      add(:reviewed_at, :utc_datetime, null: true)
+    end
+
+    create(index(:scenes, [:reviewed_at], where: "reviewed_at is null or reviewed_at < updated_at"))
```
brianpeiris: neat

lib/ret_web/views/api/v1/hub_view.ex
```diff
@@ -11,6 +11,19 @@ defmodule RetWeb.Api.V1.HubView do
   end
 
   def render("show.json", %{hub: %Hub{scene: %Scene{}} = hub}) do
+    hub |> render_with_scene
+  end
+
+  def render("show.json", %{hub: %Hub{scene_listing: %SceneListing{}} = hub}) do
```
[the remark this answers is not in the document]
gfodor: ah, because the field name differs (`scene` vs `scene_listing`)

lib/ret/media_search.ex
gfodor: yep, this is a WIP from a while ago, I'm planning on working on this API next and agree we should be doing some smart stuff here

lib/ret/media_search.ex
gfodor: the ilike is sanitized for quote escaping but i'll check the tag one. however it's still a risk if/when this table gets too big because it's possible to construct some gnarly like clauses here (because you can inject `%`'s

lib/ret/media_search.ex
gfodor: yep sketchfab was just some old code that isn't working yet

priv/repo/migrations/20190206165445_move_enums_into_ret0.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Ret.Repo.Migrations.MoveEnumsIntoRet0 do
+  use Ecto.Migration
+
+  def up do
+    execute("alter type public.scene_state set schema ret0")
```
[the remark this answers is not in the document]
gfodor: ah good catch, this is because the migration of enums wrt schema is dictated by the code at HEAD (since it dispatches to the enum definitions to find the schema) -- so if you replay the migrations it will end up being in the right schema the first time through now since the schema is now correct at HEAD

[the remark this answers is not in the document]
gfodor: this is a bit gnarly, what i'm going to do instead is just let this run when i push to fix the problem, and then comment out + commit the migration script. at that point, everything will work as expected and the live schemas will be correct.

lib/ret/media_search.ex
gfodor: ah yes bc we are using `?` variable replacement w `fragment` there should be no injection afaik

gfodor: also the solution to the `ilike` one is to eventually migrate this into a proper search engine (postgresql has one built in which is probably where i'd start)

brianpeiris: Ah, you're right, `ilike` is already escaped. 👍

PR #118, opened 2019-02-09
Avatar Reskinning

netpro2k:
Adds the ability to new GLTFs by combining a source GLTF + replacement texture maps represented as an "Avatar" in our database.

Avatar records can contain a gltf file, a bin file as well as 4 image maps. All file fields are references to "owned files" in our database and may be null. An avatar MUST contain either a gltf+bin file or a ancestor that contains them. All file fields will shadow those in ancestors ultimately resulting in a single collapsed avatar which gets generated at request time.

The current customization is a hardcoded replacement of the maps of a single named material (Currently named  `Bot_PBS`), this will likely later be expanded to be configurable by the gltf asset itself and will likely grow to allow other forms of customization outside of texture map replacements. At this point the avatar table will likely be broken up some, though the migration process should be relatively straightforward since all avatars that exist before the migration will be of the same hardcoded type.

The GLTF files are currently rewritten on the fly and the queries used to do so are not particularly optimal, so we may want to stick some form of caching in front of it at some point down the line.

The client will at first also be hardcoded to only expose reskinning of a particular avatar mesh which we provide. A single Avatar is created per account and just updated in place. We will later roll out an avatar browser that works very much like the scene selector which will allow avatars to be promoted and featured just like scenes.


lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+    conn |> show_gltf(Avatar |> Repo.get_by(avatar_sid: avatar_sid))
+  end
+
+  def show_gltf(conn, nil = _avatar) do
+    conn |> send_resp(404, "")
+  end
+
+  def show_gltf(conn, %Avatar{} = avatar) do
+    # TODO we ideally don't need to be featching the OwnedFiles until after we collapse them
+    avatar_files =
```
gfodor: i'd move this merging stuff out of the controller and probably into its own module `GLTFUtil` or `AvatarUtil` or something. also it'd be nice to have a simple positive unit test or two in place against the pure function parts just so as the semantics evolve over time we have the foundation to capture bugs and improvements under test.

gfodor:
some warnings

```

warning: variable "content_type" is unused
  lib/ret_web/controllers/api/v1/avatar_controller.ex:139

warning: function create_or_update/4 is unused
  lib/ret_web/controllers/api/v1/avatar_controller.ex:32

warning: function url_for_avatar/1 is unused
  lib/ret_web/views/api/v1/avatar_view.ex:5

warning: Map.replace/3 is deprecated. Use Map.fetch/2 + Map.put/3 instead
  lib/ret_web/controllers/api/v1/avatar_controller.ex:161
```

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+    ]
+  }
+  @image_columns Map.keys(@texture_paths)
+  @file_columns [:gltf_owned_file, :bin_owned_file] ++ @image_columns
+
+  def create(conn, %{"avatar" => params}) do
+    create_or_update(conn, params, %Avatar{})
+  end
+
+  def update(conn, %{"id" => avatar_sid, "avatar" => params}) do
```
gfodor: should pull account here

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+         %Account{account_id: account_id}
+       )
+       when not is_nil(avatar_account_id) and avatar_account_id != account_id do
+    conn |> send_resp(401, "")
+  end
+
+  defp create_or_update(conn, params, avatar) do
+    account = conn |> Guardian.Plug.current_resource()
+
+    files_to_promotoe =
```
gfodor: typo

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+        files ->
+          files
+          |> Enum.map(fn {k, v} -> {String.to_atom(k), List.to_tuple(v)} end)
+          |> Enum.into(%{})
+      end
+
+    owned_file_results = Storage.promote(files_to_promotoe, account)
+
+    promotion_error =
+      owned_file_results |> Map.values() |> Enum.filter(&(elem(&1, 0) == :error)) |> Enum.at(0)
```
gfodor: it may be worth revisiting the return value of the map variant of `promote` to return a wrapper tuple that encodes if there is any error, so this weird destructuring can go away here and in scene controller

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+
+    case promotion_error do
+      nil ->
+        owned_files =
+          owned_file_results |> Enum.map(fn {k, {:ok, file}} -> {k, file} end) |> Enum.into(%{})
+
+        parent_avatar =
```
gfodor: subjective: i believe this is equivalent to `params["parent_avatar_id"] && ...`

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+
+  def show_gltf(conn, nil = _avatar) do
+    conn |> send_resp(404, "")
+  end
+
+  def show_gltf(conn, %Avatar{} = avatar) do
+    # TODO we ideally don't need to be featching the OwnedFiles until after we collapse them
+    avatar_files =
+      avatar
+      |> Avatar.load_parents(@file_columns)
```
gfodor: on the fence somewhat, but i might want to push the preload concern into `load_parents` here, and just contractually say when you call that function we have a fixed set of preloads. the issue is just that the other places where we do preloading it's the controller literally calling `preload` so it's very clear what is going on -- we aren't typically parameterizing other functions to include preload fields and pass them around.

lib/ret/avatar.ex
```diff
@@ -0,0 +1,95 @@
...
+
+    timestamps()
+  end
+
+  def load_parents(avatar, preload_fields \\ [])
+
+  def load_parents(%Avatar{parent_avatar: nil} = avatar, preload_fields),
+    do: avatar |> Repo.preload(preload_fields)
+
+  def load_parents(%Avatar{parent_avatar: %Ecto.Association.NotLoaded{}} = avatar, preload_fields) do
```
gfodor: i think we can drop the `{ parent_avatar: ` match here -- that `NotLoaded` struct might be subject to change, and i think it's still semantically the same fall through to just ignore what non-nil value `parent_avatar` is taking on here

lib/ret/avatar.ex
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  defp put_owned_files(changeset, owned_files) do
+    Enum.reduce(owned_files, changeset, fn {key, file}, changes ->
+      put_owned_file(changes, key, file)
+    end)
+  end
+
+  defp put_owned_file(changeset, key, owned_file) do
+    changeset |> put_change("#{key}_owned_file_id" |> String.to_atom(), owned_file.owned_file_id)
```
gfodor: i'd just inline this

gfodor: also you can do `:"#{key}_owned_file_id"`

lib/ret/avatar.ex
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  defp maybe_add_avatar_sid_to_changeset(changeset) do
+    avatar_sid = changeset |> get_field(:avatar_sid) || Sids.generate_sid()
+    put_change(changeset, :avatar_sid, "#{avatar_sid}")
```
gfodor: string interp unnecessary. not sure if this was copy pasted from somewhere, probably worth changing it there too if so

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+      |> Map.from_struct()
```
gfodor: any reason to prefer struct here vs `Ret.Avatar`?

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+
+      {:error, :not_found} ->
+        conn |> send_resp(404, "no such file(s)")
+
+      {:error, :not_allowed} ->
+        conn |> send_resp(401, "")
+    end
+  end
+
+  defp collapse_avatar_files(%{parent_avatar: nil} = avatar),
```
gfodor: these kind of mapping functions i like naming `X_to_Y` so i'd call this `avatar_to_collapsed_files`

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+    do: avatar |> Map.take(@file_columns)
+
+  defp collapse_avatar_files(%{parent_avatar: parent} = avatar) do
+    Map.merge(collapse_avatar_files(parent), avatar |> Map.take(@file_columns), fn _k, v1, v2 ->
```
gfodor: you could also refactor this into a match. probably nbd, but in general `if` is less idiomatic in elixir than matching

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+
+      {:error, :not_allowed} ->
+        conn |> send_resp(401, "")
+    end
+  end
+
+  defp collapse_avatar_files(%{parent_avatar: nil} = avatar),
+    do: avatar |> Map.take(@file_columns)
+
+  defp collapse_avatar_files(%{parent_avatar: parent} = avatar) do
```
gfodor: i'd refactor this into:
```
parent
|> avatar_to_collapsed_files
|> Map.merge(...)

File: lib/ret_web/controllers/api/v1/avatar_controller.ex [hunk 14]
```diff
@@ -0,0 +1,192 @@
...
+      avatar
+      |> Avatar.load_parents(@file_columns)
+      |> Map.from_struct()
+      |> collapse_avatar_files()
+
+    case Storage.fetch(avatar_files.gltf_owned_file) do
+      {:ok, %{"content_type" => content_type, "content_length" => content_length}, stream} ->
+        gltf =
+          stream
+          |> Enum.join("")
```
gfodor: You can try 
```
stream
|> Stream.concat
|> Stream.run
|> Poison.decode!
```

if that works it will result in less blocking and i believe also less copying -- will just end up concatenating everything streamed through into a materialized enum i think

gfodor: (because of the `run` it will block the call to decode until the stream is finished, which is less than ideal, so you could try it without the `run` too, but i think `Poison` expects an `iolist`)

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -0,0 +1,192 @@
...
+      nil ->
+        gltf
+
+      material ->
+        image_files
+        |> Enum.filter(fn {_, v} -> v end)
+        |> Enum.flat_map(fn {col, file} ->
+          Enum.map(@texture_paths[col], fn path ->
+            texture_index = material |> get_in(path ++ ["index"])
+            image_index = gltf |> get_in(["textures", Access.at(texture_index), "source"])
```
gfodor: i don't usually see `Access` referenced directly, it may be more idiomatic to just use `Enum.at`

lib/ret_web/views/api/v1/avatar_view.ex
```diff
@@ -0,0 +1,43 @@
...
+  def render_avatar(avatar) do
+    %{
+      avatar_id: avatar.avatar_sid,
+      parent_avatar_id: unless(is_nil(avatar.parent_avatar), do: avatar.parent_avatar.avatar_sid),
+      name: avatar.name,
+      description: avatar.description,
+      attributions: if(is_nil(avatar.attributions), do: [], else: avatar.attributions),
+      allow_remixing: avatar.allow_remixing,
+      allow_promotion: avatar.allow_promotion,
+      files: %{
```
gfodor: you could refactor this into a list comprehension but it might turn out to be uglier

gfodor:
looks good nice work -- main issue is the account check that should block, but I don't like red X'ing :)

lib/ret/avatar.ex
netpro2k: Yep it was from scene, will change there as well

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Ended up moving this into the avatar model itself so it feels less odd. The idea is that you might want to load parents but only with specific fields.

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Probably left this here when trying to figure out how to collapse, don't think its needed anymore.

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Hmm messed with this for a bit but couldn't get another solution to work

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Hmm, `Enum.at` does not appear to be a function. I am pretty sure this is the "correct" way to do this sort of programmatic access with the `X_in` functions. https://hexdocs.pm/elixir/Access.html#at/1. Looking this up I also just learned about `Access.all` which is pretty nifty.

gfodor: cool. Enum.at def exists fyi https://hexdocs.pm/elixir/Enum.html#at/3

gfodor: cool. Enum.at def exists fyi https://hexdocs.pm/elixir/Enum.html#at/3

gfodor: oh i was mis-reading what this did. didn't realize Access.at was higher order, but clearly that makes sense

netpro2k: Sorry, I meant `Enum.at/1` does not exist.

PR #123, opened 2019-02-28
Render scene owned file

gfodor:
Adds the .spoke file in API responses for re-mixable Spoke scenes.


gfodor:
(Needed to add remixing to hubs)

PR #125, opened 2019-03-08
Add project routes for hosted Spoke (WIP)

robertlong:


priv/repo/migrations/20190306032224_create_projects.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Ret.Repo.Migrations.CreateProjects do
+  use Ecto.Migration
+
+  def change do
+    create table(:projects, prefix: "ret0", primary_key: false) do
+      add(:project_id, :bigint, default: fragment("ret0.next_id()"), primary_key: true)
+      add(:project_sid, :string)
+      add(:name, :string, null: false)
+      add(:created_by_account_id, references(:accounts, column: :account_id))
```
gfodor: `null: false` I think

lib/ret/project.ex
```diff
@@ -0,0 +1,56 @@
...
+    belongs_to(:project_owned_file, Ret.OwnedFile, references: :owned_file_id)
+    belongs_to(:thumbnail_owned_file, Ret.OwnedFile, references: :owned_file_id)
+
+    timestamps()
+  end
+
+  def project_by_sid(sid) do
+    Project |> Repo.get_by(project_sid: sid)
+  end
+
```
gfodor: the `to_sid` function was overloaded to deal with `Scene` and `SceneListing`s, so you probably can inline it in this case since I don't think there's an analog

lib/ret/project.ex
```diff
@@ -0,0 +1,56 @@
...
+    ])
+    |> validate_length(:name, min: 4, max: 64)
+    # TODO BP: this is repeated from hub.ex. Maybe refactor the regex out.
+    |> validate_format(:name, ~r/^[A-Za-z0-9-':"!@#$%^&*(),.?~ ]+$/)
+    |> maybe_add_project_sid_to_changeset
+    |> unique_constraint(:project_sid)
+    |> put_assoc(:created_by_account, account)
+  end
+
+  # Update a Project with new project and thumbnail files
```
gfodor: it's generally idiomatic to write minimal-covering `changeset` functions -- for example, if we are going to have an access pattern where we *just* update the files, then this changeset function should not have the generic `params` argument. this helps reduce the API surface area, etc.

gfodor: and you'd name this `changeset_for_uploaded_project` or something

lib/ret_web/controllers/api/v1/project_controller.ex
```diff
@@ -0,0 +1,112 @@
...
+  defp get_projects(account) do
+    Repo.all from p in Project,
+      where: p.created_by_account_id == ^account.account_id,
+      preload: [:project_owned_file, :thumbnail_owned_file]
+  end
+
+  defp get_project(project_sid) do
+    project_sid
+    |> Project.project_by_sid()
+    |> Repo.preload([:created_by_account, :project_owned_file, :thumbnail_owned_file])
```
gfodor: I'd just put these preloads into `@` global and use them everywhere

gfodor: this particular function i'd move into the model

lib/ret_web/controllers/api/v1/project_controller.ex
```diff
@@ -0,0 +1,112 @@
...
+    account = conn |> Guardian.Plug.current_resource()
+
+    case project_sid |> get_project() do
+      %Project{} = project -> update(conn, params, project, account)
+      _ -> conn |> send_resp(404, "not found")
+    end
+  end
+
+  defp get_projects(account) do
+    Repo.all from p in Project,
```
gfodor: i'd move this into `Ret.Project.projects_by_account`

lib/ret_web/controllers/api/v1/project_controller.ex
robertlong: What's `@` global?

lib/ret/project.ex
gfodor: (afaict that's not the case here, but just mentioning it)

gfodor:
looking good. main thing i am not sure of is if we want to just store these project files into the db directly via `jsonb`. i'm not nervous about the sizes, i'm mostly wondering which will be simpler. if we have any scenarios where we want to be able to re-write or query against the project file then we should probably start by putting them in the db in a column. I can think of a few scenarios (like attribution modification) that seem pretty clear, so unless we can rule those out I'd advocate for a `jsonb` column. Easy to migrate off of if we find the size being problematic, but PostgreSQL should be good at dealing with this.

lib/ret_web/controllers/api/v1/project_controller.ex
robertlong: I tried this:
```
  def projects_by_account(account) do
    Repo.all from p in Project,
      where: p.created_by_account_id == ^account.account_id,
      preload: [:project_owned_file, :thumbnail_owned_file]
  end
```

And got this error:
```
lib/ret/project.ex:25: cannot use ^account.account_id() outside of match clauses
```

I'm confused, shouldn't that be equivalent?

lib/ret/project.ex
robertlong: Yeah I also allow updating the name, otherwise I could do that.

lib/ret_web/controllers/api/v1/project_controller.ex
gfodor: you can add `@foo XXX` to the top of a module to add global static values

lib/ret_web/controllers/api/v1/project_controller.ex
gfodor: you may need to move the account_id into a local var for the macro expansion to work right, it may expect `^` to be on the left hand side of a binding, not an arbitrary expression

PR #126, opened 2019-03-12
Handle removed scenes, add removed scene enum

gfodor:
Adds the `removed` enum state to scenes, and now masks out all places where we return scene information in the API if the scene (or listing) has been taken down.

### DIFF
--- a/lib/ret_web/views/api/v1/scene_view.ex
+++ b/lib/ret_web/views/api/v1/scene_view.ex
@@ -10,6 +10,9 @@ defmodule RetWeb.Api.V1.SceneView do
     %{scenes: [render_scene(scene)]}
   end
 
+  def render_scene(%Scene{state: :removed}), do: %{}
+  def render_scene(%SceneListing{state: :delisted}), do: %{}
+
   # scene var passed in can be either a Ret.Scene or Ret.SceneListing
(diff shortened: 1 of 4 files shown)


lib/ret_web/views/api/v1/scene_view.ex
```diff
@@ -10,6 +10,9 @@ defmodule RetWeb.Api.V1.SceneView do
     %{scenes: [render_scene(scene)]}
   end
 
+  def render_scene(%Scene{state: :removed}), do: nil
+  def render_scene(%SceneListing{state: :delisted}), do: nil
```
brianpeiris: Shouldn't we also filter these out of `Scene.scene_or_scene_listing_by_sid` in order to stop the scene being available at `hubs.mozilla.com/scenes/...`?

brianpeiris: Oh, I see, we're still using the scene.html page. Though I think this is still a problem, since we're leaking scene content in the meta tags, if I'm not mistaken.

gfodor: ah good point. i forgot about that part good thinking -- i thought by busting the API we were removing all visibility onto the scene info.

PR #127, opened 2019-03-13
Add ability to remvoe maps from avatars

netpro2k:
When updating an avatar files that are `null` in the request will correctly be unassociated with the avatar record.

### DIFF
--- a/lib/ret/avatar.ex
+++ b/lib/ret/avatar.ex
@@ -40,13 +40,18 @@ defmodule Ret.Avatar do
     field(:allow_promotion, :boolean)
     belongs_to(:account, Account, references: :account_id)
 
-    belongs_to(:gltf_owned_file, OwnedFile, references: :owned_file_id)
-    belongs_to(:bin_owned_file, OwnedFile, references: :owned_file_id)
+    belongs_to(:gltf_owned_file, OwnedFile, references: :owned_file_id, on_replace: :nilify)
+    belongs_to(:bin_owned_file, OwnedFile, references: :owned_file_id, on_replace: :nilify)
 
-    belongs_to(:base_map_owned_file, OwnedFile, references: :owned_file_id)
-    belongs_to(:emissive_map_owned_file, OwnedFile, references: :owned_file_id)
-    belongs_to(:normal_map_owned_file, OwnedFile, references: :owned_file_id)
-    belongs_to(:orm_map_owned_file, OwnedFile, references: :owned_file_id)
+    belongs_to(:base_map_owned_file, OwnedFile, references: :owned_file_id, on_replace: :nilify)
+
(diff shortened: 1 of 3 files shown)


lib/ret/storage.ex
```diff
@@ -83,6 +83,7 @@ defmodule Ret.Storage do
     |> Enum.map(fn
       {k, {id, key}} -> {k, promote(id, key, nil, account)}
       {k, {id, key, promotion_token}} -> {k, promote(id, key, promotion_token, account)}
+      {k, :remvoe} -> {k, {:ok, :remove}}
```
gfodor: typo. this probably deserves an integration test :)

lib/ret/avatar.ex
```diff
@@ -103,14 +108,18 @@ defmodule Ret.Avatar do
...
     |> AvatarSlug.unique_constraint()
   end
 
-  defp put_owned_files(changeset, owned_files) do
-    Enum.reduce(owned_files, changeset, fn {key, file}, changes ->
-      changes |> put_change(:"#{key}_owned_file_id", file.owned_file_id)
+  defp put_owned_files(changeset, owned_files_map) do
+    Enum.reduce(owned_files_map, changeset, fn
+      {key, :remove}, changes ->
+        changes |> put_assoc(:"#{key}_owned_file", nil)
```
gfodor: i'd probably extract the string expression into a var just to DRY it up a bit

gfodor:
i didn't run this locally, but assuming it works code lgtm :)

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -42,10 +42,19 @@ defmodule RetWeb.Api.V1.AvatarController do
...
+        {k, v} -> {String.to_atom(k), List.to_tuple(v)}
+      end)
       |> Enum.into(%{})
 
-    owned_file_results = Storage.promote(files_to_promote, account)
+    owned_file_results =
+      files_to_promote
+      |> Enum.map(fn
+        {k, {id, key, promotion_token}} -> {k, Storage.promote(id, key, promotion_token, account)}
+        {k, :remove} -> {k, {:ok, :remove}}
```
netpro2k: Don't particularly love this but I want to be able to carry this map through without having to split off the removals

lib/ret/avatar.ex
netpro2k: I could break it into another function but that seems overkill, unless there is a way to define some intermediate stuff in a functions with multiple clauses like this.

lib/ret/storage.ex
netpro2k: woops, changed that right before commit, but ended up moving this elsewhere anyway.

PR #129, opened 2019-03-22
Fix setting multiple avatar maps at once and add cachebusting url to avatar response

netpro2k:
When setting multiple maps at once on an avatar only the last map would actually apply because the wrong changeset was being carried forward. This regressed when adding the ability to remove maps.

Also adds `gltf_url` to the response for an avatar which points at the reticulum endpoint for the generated skinned avatar, with a query param for cachebusting based on the `updated_at` column. We should eventually return a CDN cached url for this.

### DIFF
--- a/lib/ret_web/views/api/v1/avatar_view.ex
+++ b/lib/ret_web/views/api/v1/avatar_view.ex
@@ -18,6 +18,7 @@ defmodule RetWeb.Api.V1.AvatarView do
   end
 
   def render_avatar(avatar) do
+    version = avatar.updated_at |> NaiveDateTime.to_erl |> :calendar.datetime_to_gregorian_seconds
     %{
       avatar_id: avatar.avatar_sid,
       parent_avatar_id: unless(is_nil(avatar.parent_avatar), do: avatar.parent_avatar.avatar_sid),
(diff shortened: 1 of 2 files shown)


lib/ret_web/views/api/v1/avatar_view.ex
```diff
@@ -26,6 +27,7 @@ defmodule RetWeb.Api.V1.AvatarView do
       attributions: if(is_nil(avatar.attributions), do: [], else: avatar.attributions),
       allow_remixing: avatar.allow_remixing,
       allow_promotion: avatar.allow_promotion,
+      gltf_url: "#{RetWeb.Endpoint.url()}/api/v1/avatars/#{avatar.avatar_sid}/avatar.gltf?v=#{version}",
```
gfodor: thinking about it more, it seems like the right way to do this is to update things to encode the sid/version into the ETag for the response

netpro2k: Adding ETAGS to this response (the avatar json) makes sense. Due to a few issues in our client (gltf caching, networking, etc) we expect the gltf url itself to change. I do think we should return a CDN url here so that it is cached.

PR #130, opened 2019-03-23
Adds a local file cache for Sketchfab

gfodor:
This adds a local file cache that expires files out after 24 hours, and wires it up to sketchfab so we'll make at most one hit to the download API per day per model id.


gfodor:
actually still seeing timeouts on long running downloads, investigating

priv/repo/migrations/20190322222314_add_cached_files.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Ret.Repo.Migrations.AddResolvedMediaAssets do
```
brianpeiris: nit: rename to `AddCachedFiles` for consistency

config/dev.exs
```diff
@@ -77,6 +77,8 @@ config :ret, Ret.Repo,
   database: "ret_dev",
   hostname: if(env_db_host == "", do: "localhost", else: env_db_host),
   template: "template0",
+  # Downloads from Sketchfab to file cache hold connections open
+  ownership_timeout: 60_000,
```
brianpeiris: Hmm. Is there a better solution to this that you're punting on for now?

lib/ret/cached_file.ex
```diff
@@ -0,0 +1,82 @@
...
+  # cached, expects loader to be a function that will receive a path to a temp file to
+  # write the data to cache to, and is expected to return a { :ok, %{ content_type: } } tuple
+  # with the content type.
+  def fetch(cache_key, loader) do
+    Repo.transaction(fn ->
+      # Use a PostgreSQL advisory lock on the cache key as a mutex across all
+      # nodes for accessing this cache key
+      <<lock_key::little-signed-integer-size(64), _::binary>> = :crypto.hash(:sha256, cache_key)
+
+      Ecto.Adapters.SQL.query!(Repo, "select pg_advisory_xact_lock($1);", [lock_key])
```
brianpeiris: neat

config/dev.exs
gfodor: Ideally we could bump this dynamically for expected long-running transactions -- this is a mechanism to avoid pool depletion for hung processes. I think we are going to be stuck with picking a global value though. We need to hold the connection because we are using the database as a locking mechanism

PR #132, opened 2019-03-25
Add Assets routes for hosted Spoke

robertlong:


lib/ret/asset.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  alias Ret.{Asset}
+
+  @schema_prefix "ret0"
+  @primary_key {:asset_id, :id, autogenerate: true}
+
+  schema "assets" do
+    field(:asset_sid, :string)
+    field(:name, :string)
+    field(:type, :string)
```
gfodor: I think this should be refactored to use an enum (in elixir and postgres) -- see `enums.ex` and the relevant schema/migrations on how to set it up.

lib/ret/asset.ex
```diff
@@ -0,0 +1,48 @@
...
+    timestamps()
+  end
+
+  # Create an Asset
+  def changeset(%Asset{} = asset, account, asset_owned_file, params) do
+    asset
+    |> cast(params, [:name])
+    |> put_change(:type, content_type_to_asset_type!(asset_owned_file.content_type))
+    |> validate_required([:name, :type])
+    |> validate_inclusion(:type, ["image", "video", "model"])
```
gfodor: if you move to enums this validator will no longer be necessary

lib/ret/asset.ex
```diff
@@ -0,0 +1,48 @@
...
+    # Asset names are defaulted to the owned file name so allow pretty much anything
+    |> validate_length(:name, min: 1, max: 256)
+    |> maybe_add_asset_sid_to_changeset
+    |> unique_constraint(:asset_sid)
+    |> put_assoc(:account, account)
+    |> put_change(:asset_owned_file_id, asset_owned_file.owned_file_id)
```
gfodor: any reason to not use `put_assoc` here?

lib/ret/asset.ex
```diff
@@ -0,0 +1,48 @@
...
+  @schema_prefix "ret0"
+  @primary_key {:asset_id, :id, autogenerate: true}
+
+  schema "assets" do
+    field(:asset_sid, :string)
+    field(:name, :string)
+    field(:type, :string)
+    belongs_to(:account, Ret.Account, references: :account_id)
+    belongs_to(:asset_owned_file, Ret.OwnedFile, references: :owned_file_id)
+    many_to_many(:projects, Ret.Project, join_through: Ret.ProjectAsset, join_keys: [ asset_id: :asset_id, project_id: :project_id], on_replace: :delete)
```
gfodor: you should add a `state` column (and check it in the various queries for assets) similar to how we do with scenes/listings/etc

gfodor: it can just have one possible value for now, `:active`

lib/ret/media_search.ex
```diff
@@ -250,6 +254,52 @@ defmodule Ret.MediaSearch do
...
+
+  defp add_query_to_asset_search_query(query, nil), do: query
+  defp add_query_to_asset_search_query(query, q), do: query |> where([a], ilike(a.name, ^"%#{q}%"))
+
+  defp result_for_assets_page(page, cursor) do
+    %Ret.MediaSearchResult{
+      meta: %Ret.MediaSearchResultMeta{
+        next_cursor:
+          if page.total_pages > cursor do
+            (cursor || 1) + 1
```
gfodor: the pagination recently changed today on master -- you'll need to update this code after pulling from master. the `cursor` being passed in now should always be a string, and be parsed in this module. (see how its done for scenes on master, and in media_search_controller for scene_listing search)

lib/ret/project_asset.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Ret.ProjectAsset do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @schema_prefix "ret0"
+  @primary_key false
+
+  schema "project_assets" do
+    belongs_to(:project, Ret.Project, references: :project_id)
```
gfodor: this needs a primary key as we talked about

lib/ret_web/controllers/api/v1/media_search_controller.ex
```diff
@@ -54,4 +64,32 @@ defmodule RetWeb.Api.V1.MediaSearchController do
   def index(conn) do
     conn |> send_resp(422, "")
   end
+
+  defp user_assets_index(conn, account, %{"source" => "assets"} = params) do
+    {:commit, results} =
+      %Ret.MediaSearchQuery{source: "assets", user: account.account_id, type: "all", cursor: params["cursor"] || 1} |> Ret.MediaSearch.search()
```
gfodor: if you pull from master you'll see the cursor is now handled differently, you'll need to fall back to passing `"1"` here

lib/ret_web/controllers/api/v1/media_search_controller.ex
```diff
@@ -54,4 +64,32 @@ defmodule RetWeb.Api.V1.MediaSearchController do
...
+
+    conn |> render("index.json", results: results)
+  end
+
```
gfodor: once you refactor the schema to use enums, I'd DRY this up a bit by just checking which enum value to use for the `type` in a lookup table and extracting the rest into its own function

lib/ret_web/controllers/api/v1/project_assets_controller.ex
```diff
@@ -0,0 +1,62 @@
...
+            conn |> send_resp(422, "Invalid asset")
+          {:error, :project_asset, _changeset, %{}} ->
+            conn |> send_resp(500, "Error creating project asset")
+        end
+      {:error, _reason} ->
+        conn |> send_resp(422, "invalid owned file")
+    end
+  end
+
+  defp get_project(account, project_sid) do
```
gfodor: I'd move this into `Ret.Project.project_by_sid_for_account`

lib/ret_web/controllers/api/v1/project_assets_controller.ex
```diff
@@ -0,0 +1,62 @@
...
+      %Project{} = project -> create(conn, params, account, project)
+      nil -> conn |> send_resp(404, "Project not found")
+    end
+  end
+
+  defp create(conn, params, account, project) do
+    case Storage.promote(params["file_id"], params["access_token"], nil, account) do
+      {:ok, asset_owned_file} ->
+
+        asset_changeset = Asset.changeset(%Asset{}, account, asset_owned_file, params)
```
gfodor: i'd move this operation into a function on `Ret.Asset` that returns the result of `Repo.transaction`

lib/ret_web/controllers/api/v1/project_controller.ex
```diff
@@ -0,0 +1,112 @@
...
+  def update(conn, %{"id" => project_sid, "project" => params}) do
+    account = conn |> Guardian.Plug.current_resource()
+
+    case project_sid |> get_project() do
+      %Project{} = project -> update(conn, params, project, account)
+      _ -> conn |> send_resp(404, "not found")
+    end
+  end
+
+  defp get_projects(account) do
```
gfodor: I'd move these db interacting function into `Ret.Projects.projects_for_account` and `Ret.Projects.project_by_sid`

PR #133, opened 2019-03-26
Fix bug in scene listing pagination

gfodor:
Fixes a bug which caused scene listing pagination in the search API to not return results after the second page -- we were passing a string instead of an integer to the pagination routine :P


gfodor:
That's true, I guess we should just assume cursors are always strings, and I should refactor accordingly.

gfodor:
closing in favor of https://github.com/mozilla/reticulum/pull/134

PR #135, opened 2019-03-28
Properly render farspark urls for pdf uploads

gfodor:
Fixes a bug where we were not properly returning farspark paginating URLs for PDF uploads.

### DIFF
--- a/lib/ret_web/controllers/api/v1/media_controller.ex
+++ b/lib/ret_web/controllers/api/v1/media_controller.ex
@@ -37,14 +37,20 @@ defmodule RetWeb.Api.V1.MediaController do
 
     case Ret.Storage.store(upload, content_type, access_token, promotion_token) do
       {:ok, uuid} ->
-        uri = Ret.Storage.uri_for(uuid, content_type)
+        origin_uri = Ret.Storage.uri_for(uuid, content_type)
+
+        raw_uri =
+          case content_type do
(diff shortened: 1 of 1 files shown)


lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -37,14 +37,20 @@ defmodule RetWeb.Api.V1.MediaController do
 
     case Ret.Storage.store(upload, content_type, access_token, promotion_token) do
       {:ok, uuid} ->
-        uri = Ret.Storage.uri_for(uuid, content_type)
+        origin_uri = Ret.Storage.uri_for(uuid, content_type)
+
+        raw_uri =
+          case content_type do
+            "application/pdf" -> gen_farspark_url(origin_uri, 0, "raw", "") |> URI.parse()
```
brianpeiris: This looks fine, but it might be an opportunity to refactor `gen_farspark_url`. The only other usage also uses `"raw", ""`, so maybe those arguments are redundant now.

gfodor:
this fix ended up being wrong. backing it out. i wasn't able to integration test it locally bc i didn't feel like setting up farspark, but on smoke i realized the raw URL we return should not be farsparked. (the client is failing to forward the `expected_content_type` for uploads along to the media loader so it will properly deal with the PDF)

PR #146, opened 2019-04-04
Discord permission mappings

brianpeiris:
Refactor and generalize permissions checks. Add permissions based on discord permissions.

### DIFF
--- a/lib/ret/discord_client.ex
+++ b/lib/ret/discord_client.ex
@@ -43,17 +43,17 @@ defmodule Ret.DiscordClient do
     |> Poison.decode!()
   end
 
-  def member_of_channel?(nil, _), do: false
+  def has_permission?(nil, _, _), do: false
 
-  def member_of_channel?(%Ret.OAuthProvider{} = oauth_provider, %Ret.HubBinding{} = hub_binding) do
-    oauth_provider.provider_account_id |> member_of_channel?(hub_binding)
+  def has_permission?(%Ret.OAuthProvider{} = oauth_provider, %Ret.HubBinding{} = hub_binding, permission) do
+    oauth_provider.provider_account_id |> has_permission?(hub_binding, permission)
   end
 
-  def member_of_channel?(account_id, %Ret.HubBinding{} = hub_binding) when is_binary(account_id) do
+  def has_permission?(account_id, %Ret.HubBinding{} = hub_binding, permission) when is_binary(account_id) do
     permissions =
       compute_permissions(account_id, hub_binding.community_id, hub_binding.channel_id) |> permissions_to_map
 
-    permissions[:view_channel]
+    permissions[permission]
   end
 
   def api_request(path) do
--- a/lib/ret/discord_client.ex
+++ b/lib/ret/discord_client.ex
@@ -43,7 +43,13 @@ defmodule Ret.DiscordClient do
     |> Poison.decode!()
   end
 
-  def member_of_channel?(account_id, %Ret.HubBinding{} = hub_binding) do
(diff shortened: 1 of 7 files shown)


lib/ret/discord_client.ex
```diff
@@ -43,11 +43,17 @@ defmodule Ret.DiscordClient do
...
   end
 
-  def member_of_channel?(account_id, %Ret.HubBinding{} = hub_binding) do
+  def has_permission?(nil, _, _), do: false
+
+  def has_permission?(%Ret.OAuthProvider{} = oauth_provider, %Ret.HubBinding{} = hub_binding, permission) do
+    oauth_provider.provider_account_id |> has_permission?(hub_binding, permission)
+  end
+
+  def has_permission?(account_id, %Ret.HubBinding{} = hub_binding, permission) when is_binary(account_id) do
```
gfodor: i'd rename `account_id` to `provider_account_id`, since `account_id` implied to me it was expecting a `Ret.Account` id

lib/ret/hub.ex
```diff
@@ -260,30 +260,28 @@ defmodule Ret.Hub do
...
+  def can?(%Ret.Account{} = account, :join_hub, %Ret.Hub{hub_bindings: hub_bindings})
+      when hub_bindings |> length > 0 do
+    hub_bindings |> Enum.any?(&(account |> Ret.HubBinding.member_of_channel?(&1)))
+  end
 
-  # Anyone can join an unbound hub
-  def can?(_, :join_hub, %Ret.Hub{hub_bindings: []}), do: true
+  def can?(%Ret.Account{} = account, :update_hub, %Ret.Hub{hub_bindings: hub_bindings})
+      when hub_bindings |> length > 0 do
+    hub_bindings |> Enum.any?(&(account |> Ret.HubBinding.can_manage_channel?(&1)))
```
gfodor: this should return true if the account is the owner of the room as usual as well

on that note -- we should discuss if we want to get things set up so the bot will associate the room with the account of the person who typed !hub create -- that seems mostly logical to me. it would probably need to be lazy in that the bot would have to note the user's discord id somewhere, and then set it downstream once they go through the OAuth flow. (to handle the case where they do not have an account yet)

lib/ret/hub.ex
```diff
@@ -260,30 +260,28 @@ defmodule Ret.Hub do
...
+  end
 
-  def can?(%Ret.Account{oauth_providers: oauth_providers}, :join_hub, %Ret.Hub{hub_bindings: hub_bindings})
-      when hub_bindings |> length > 0 and oauth_providers |> length > 0 do
-    hub_bindings
-    |> Enum.any?(fn binding ->
-      provider = oauth_providers |> Enum.find(&(&1.source == binding.type))
+  def can?(%Ret.Account{} = account, action, %Ret.Hub{hub_bindings: hub_bindings})
+      when hub_bindings |> length > 0 and action in [:kick_users, :mute_users] do
```
gfodor: same, should also respect room owner

lib/ret/hub.ex
brianpeiris: Um, I think I will leave this as it is until we make those decisions -- my working theory right now is that once a hub is bound to a channel, it should _never_ be unbound -- that helps cut down on the scenarios significantly. In that case, any owner/moderator of a community will always be able to edit and moderate any hub that was ever created in that community and that seems right to me. If the owner/moderator is removed from that community, they should no longer have access to those hubs, so it makes sense not to mark their Hubs account as the owner of the hub.

PR #149, opened 2019-04-08
New Asset, Project, and ProjectAsset routes

robertlong:
- Added POST `/api/v1/assets` route which creates a new user owned asset and accepts an owned file id and access token
- Added DELETE`/api/v1/assets/:id` route which deletes a user owned asset and all ProjectAsset relations
- Added DELETE `/api/v1/projects/:id` for deleting a project
- Added an additional POST `/api/v1/projects/:project_id/assets` route for adding an existing asset to a project by creating a ProjectAsset relation
- Added DELETE `/api/v1/projects/:project_id/assets/:id` for removing an asset from a project by deleting the ProjectAsset relation

### DIFF
--- a/lib/ret/asset.ex
+++ b/lib/ret/asset.ex
@@ -40,7 +40,7 @@ defmodule Ret.Asset do
     Repo.transaction(multi)
   end
 
-  def asset_by_sid_for_account(account, asset_sid) do
+  def asset_by_sid_for_account(asset_sid, account) do
     from(a in Asset,
       where: a.asset_sid == ^asset_sid and a.account_id == ^account.account_id,
       preload: [:account, :asset_owned_file, :thumbnail_owned_file])
--- a/lib/ret/asset.ex
+++ b/lib/ret/asset.ex
@@ -1,6 +1,7 @@
 defmodule Ret.Asset do
   use Ecto.Schema
   import Ecto.Changeset
+  import Ecto.Query
 
   alias Ecto.{Multi}
   alias Ret.{Repo, Asset, ProjectAsset}
@@ -20,6 +21,12 @@ defmodule Ret.Asset do
     timestamps()
   end
 
+  def create_asset(account, asset_owned_file, thumbnail_owned_file, params) do
+    %Asset{}
+    |> Asset.changeset(account, asset_owned_file, thumbnail_owned_file, params)
+    |> Repo.insert()
+  end
+
   def create_asset_and_project_asset(account, project, asset_owned_file, thumbnail_owned_file, params) do
     asset_changeset = Asset.changeset(%Asset{}, account, asset_owned_file, thumbnail_owned_file, params)
 
@@ -31,6 +38,13 @@ defmodule Ret.Asset do
       end)
 
     Repo.transaction(multi)
+  end
+
+  def asset_by_sid_for_account(account, asset_sid) do
+    from(a in Asset,
+      where: a.asset_sid == ^asset_sid and a.account_id == ^account.account_id,
+      preload: [:account, :asset_owned_file, :thumbnail_owned_file])
+    |> Repo.one
   end
 
   # Create an Asset
--- a/lib/ret/project.ex
+++ b/lib/ret/project.ex
@@ -3,7 +3,7 @@ defmodule Ret.Project do
   import Ecto.Changeset
   import Ecto.Query
 
-  alias Ret.{Repo, Project, OwnedFile}
+  alias Ret.{Repo, Project, ProjectAsset, OwnedFile}
 
   @schema_prefix "ret0"
   @primary_key {:project_id, :id, autogenerate: true}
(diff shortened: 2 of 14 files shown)


lib/ret/asset.ex
```diff
@@ -33,6 +40,13 @@ defmodule Ret.Asset do
     Repo.transaction(multi)
   end
 
+  def asset_by_sid_for_account(account, asset_sid) do
```
gfodor: to align with function name i'd just reverse the arguments (nit)

lib/ret/asset.ex
```diff
@@ -33,6 +40,13 @@ defmodule Ret.Asset do
...
+    from(a in Asset,
+      where: a.asset_sid == ^asset_sid and a.account_id == ^account.account_id,
+      preload: [:account, :asset_owned_file, :thumbnail_owned_file])
```
gfodor: should check `state` is active I think

lib/ret/project.ex
```diff
@@ -40,6 +40,12 @@ defmodule Ret.Project do
       preload: [:project_owned_file, :thumbnail_owned_file]
   end
 
+  def add_asset_to_project(project, asset) do
+    %ProjectAsset{}
+    |> ProjectAsset.changeset(project, asset)
+    |> Repo.insert
```
gfodor: `insert!` here so it will raise if fail

lib/ret_web/controllers/api/v1/project_assets_controller.ex
```diff
@@ -15,7 +15,26 @@ defmodule RetWeb.Api.V1.ProjectAssetsController do
     end
   end
 
-  def create(conn, %{"id" => project_sid, "asset" => params}) do
+  def create(conn, %{"project_id" => project_sid, "asset_id" => asset_sid}) do
+    account = conn |> Guardian.Plug.current_resource()
+
+    case Project.project_by_sid_for_account(account, project_sid) do
```
gfodor: you can refactor this into `with` to reduce nesting, etc. same below

gfodor: you have a few other `case` statements like this that are less nested but will likely be better if they used `with` macro

test/ret_web/controllers/api/assets_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  test "assets delete works when logged in", %{conn: conn, asset: asset} do
+    conn |> delete(api_v1_assets_path(conn, :delete, asset.asset_sid)) |> response(200)
+
+    deleted_asset = Asset |> Repo.get_by(asset_sid: asset.asset_sid)
+
+    assert deleted_asset == nil
+  end
+
+  @tag :authenticated
+  test "assets delete shows a 404 when the user does not own the asset", %{conn: conn, thumbnail_owned_file: thumbnail_owned_file} do
```
gfodor: thanks for writing these!

lib/ret/asset.ex
robertlong: On the owned files? Asset doesn't have an active field, I thought we had settled on actually deleting assets/project/project_assets instead of using an active field?

gfodor: ah if we left that column out that is fine. in general i've been adding the state column pre-emptively in the data model (but don't really care that much if we do it later.) my past experience is that doing it up-front can be valuable so you don't miss places later if/when you need to add a soft delete or other filtering, but not a big deal to leave it alone for now.

lib/ret/project.ex
gfodor: nvmd, talked through this

lib/ret_web/controllers/api/v1/assets_controller.ex
```diff
@@ -0,0 +1,31 @@
...
+      render(conn, "show.json", asset: asset)
+    else
+      {:error, error } ->  render_error_json(conn, error)
+    end
+  end
+
+  def delete(conn, %{"id" => asset_sid }) do
+    account = Guardian.Plug.current_resource(conn)
+
+    case Asset.asset_by_sid_for_account(asset_sid, account) do
```
gfodor: this can become `with` I think

gfodor: (with a guard clause checking for `nil`)

lib/ret_web/controllers/api/v1/project_controller.ex
```diff
@@ -13,86 +13,51 @@ defmodule RetWeb.Api.V1.ProjectController do
   end
 
   def show(conn, %{"id" => project_sid}) do
-    case project_sid |> Project.project_by_sid() do
+    account = Guardian.Plug.current_resource(conn)
+
+    case Project.project_by_sid_for_account(project_sid, account) do
```
gfodor: similarly can become `with` with a guard clause for nil

gfodor:
LGTM now -- two more places you could optionally refactor to `with`, may not be worth it depending on how it looks

robertlong:
I'm going to opt out of using with. The syntax isn't really better and in these cases there's only two values they can be.

PR #150, opened 2019-04-09
Render 480p videos on mobile, 720p videos on desktop

gfodor:


lib/ret/media_resolver.ex
```diff
@@ -391,4 +379,28 @@ defmodule Ret.MediaResolver do
...
+    # Prefer a version with baked in (english) subtitles. Client locale should eventually determine this
+    "best#{ext}[format_id*=hardsub-enUS][height<=?#{resolution}]/best#{ext}[format_id*=hardsub-enUS]/" <>
+      ytdl_query(query, nil)
+  end
+
+  defp ytdl_query(query, _root_host) do
+    resolution = query |> ytdl_resolution
+    ext = query |> ytdl_ext
+
+    "best#{ext}[protocol*=http][height<=?#{resolution}]/best#{ext}[protocol*=m3u8][height<=?#{resolution}]/best#{ext}[protocol*=http]/best#{
```
netpro2k: These are starting to get a tad unwieldily, might be nice to make a list and join with "/", but not a huge deal

PR #152, opened 2019-04-09
Authoritative NAF first-syncs

brianpeiris:
We want to securely associate "first sync" NAF messages with the associated socket's session id. To that end, we now generate the socket session id in reticulum, instead of allowing the client to tell us their own session id. We then use this session id to override the "creator" and "owner" fields in "first sync" networked-aframe messages.

### DIFF
--- a/lib/ret/session_token.ex
+++ b/lib/ret/session_token.ex
@@ -1,4 +1,9 @@
 defmodule Ret.SessionToken do
+  @moduledoc """
+  SessionTokens are used to securely associated a client with a session_id.
+  Session uuids are generated on the server and the tokens containing them are stored in clients' localStorage.
+  They allow us to re-establish an verify a session if a client needs to re-connect, without losing session state.
+  """
   use Guardian, otp_app: :ret, secret_fetcher: Ret.SessionTokenSecretFetcher, allowed_algos: ["HS512"]
 
   def subject_for_token(_resource, %{"session_id" => session_id}) do
--- a/lib/ret/session_token.ex
+++ b/lib/ret/session_token.ex
@@ -25,11 +25,11 @@ defmodule Ret.SessionToken do
 end
 
 defmodule Ret.SessionTokenSecretFetcher do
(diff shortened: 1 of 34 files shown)


lib/ret_web/channels/session_socket.ex
```diff
@@ -12,10 +12,25 @@ defmodule RetWeb.SessionSocket do
     "session:#{socket.assigns.session_id}"
   end
 
-  def connect(%{"session_id" => session_id}, socket) do
+  def connect(%{"session_token" => session_token}, socket) do
+    session_id =
```
gfodor: This seems worth a comment

lib/ret_web/channels/session_socket.ex
```diff
@@ -12,10 +12,25 @@ defmodule RetWeb.SessionSocket do
...
+      case session_token |> Ret.SessionToken.decode_and_verify() do
```
gfodor: I'd push this little routine into a function on `SessionToken` so you can then just do:
```
assign(:session_id, session_token |> SessionToken.session_id_for_token || SecureRandom.uuid()
```

also might be nice to extract `SecureRandom.uuid()` into a function `generate_session_id` for documentation purposes

lib/ret/session_token.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Ret.SessionToken do
```
gfodor: We have a few of these JWT tokens now, I think we should add a one or two line comment at the top of all these describing them so we can keep them straight

lib/ret_web/channels/session_socket.ex
brianpeiris: Hmm, not sure what you're getting at. The session token allows us to trust the session id contained in the token. Should I add a comment to the code saying that?

brianpeiris: Perhaps the moduledoc I just added is sufficient?

PR #155, opened 2019-04-12
Feature/screenshot thumbnail

gfodor:
Adds screenshot/thumbnail responses when media resolving web pages not able to be crawled via yt-dl.


lib/ret/media_resolver.ex
```diff
@@ -191,31 +191,109 @@ defmodule Ret.MediaResolver do
...
+
+    # Crawl og tags for hubs rooms + scenes
+    is_local_url = host === RetWeb.Endpoint.host()
+
+    # For text/html pages we use the screenshotter, otherwise return the raw URL.
+    # Try HEAD, if HEAD fails then do a GET and check content type.
+    # If HEAD works, check content type.
+    case uri |> URI.to_string() |> retry_head_until_success() do
+      :error ->
+        case uri |> URI.to_string() |> retry_get_until_success([{"Range", "bytes=0-32768"}]) do
```
netpro2k: Why this byte range?

lib/ret/media_resolver.ex
```diff
@@ -191,31 +191,109 @@ defmodule Ret.MediaResolver do
...
+  defp resolve_non_video(%URI{host: host} = uri, _root_host) do
+    photomnemonic_endpoint = module_config(:photomnemonic_endpoint)
+
+    # Crawl og tags for hubs rooms + scenes
+    is_local_url = host === RetWeb.Endpoint.host()
+
+    # For text/html pages we use the screenshotter, otherwise return the raw URL.
+    # Try HEAD, if HEAD fails then do a GET and check content type.
+    # If HEAD works, check content type.
+    case uri |> URI.to_string() |> retry_head_until_success() do
```
netpro2k: Control flow here feels pretty odd. Can't this case just always evaluate to either a valid http response or nil (Does the HEAD then GET fallback). And then do the content type check after to decide if it should pass to photomnemonic or do the og tags. Currently that second part happens in two places.

lib/ret/media_resolver.ex
```diff
@@ -191,31 +191,109 @@ defmodule Ret.MediaResolver do
...
+
+        headers =
+          if photomnemonic_api_id do
+            %{"x-amzn-apigateway-api-id" => photomnemonic_api_id}
+          else
+            %{}
+          end
+
+        url = "#{photomnemonic_endpoint}/screenshot?#{query}"
+        {:ok, _path} = Download.from(url, path: path, headers: headers)
```
netpro2k: Don't think we handle the case where this fails. I think the intent is for it to fall back to fetching OG tags.

PR #156, opened 2019-04-15
Override prescence with discord identity

brianpeiris:
Based on #152. Overrides presence updates with display names from Discord.


lib/ret_web/channels/hub_channel.ex
```diff
@@ -401,9 +401,48 @@ defmodule RetWeb.HubChannel do
   end
 
   defp presence_meta_for_socket(socket) do
-    socket.assigns |> Map.take([:presence, :profile, :context])
+    socket.assigns
+    |> override_display_name(socket)
```
gfodor: consider `maybe_override_display_name`

PR #160, opened 2019-04-22
Add close_hub permission and verbs

gfodor:
Adds the ability to close a hub if you have owner permissions.

Also updates `join_hub` permission to be denied to all users when the hub's `entry_mode`' is `deny`.

I carved out a new permission for this, vs using `update_hub` permission -- somewhat subjective but it felt like this could theoretically be conferred to different roles in the future. (Eg a community member running an event may be able to rename the room or change the scene, but not be allowed to shut down the room, which is destructive.)

Goes with https://github.com/mozilla/hubs/pull/1260

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -118,14 +118,10 @@ defmodule Ret.Hub do
     |> validate_required([:spawned_object_types])
   end
 
-  def changeset_to_deny_entry(%Hub{} = hub) do
-    hub
-    |> cast(%{entry_mode: :deny}, [:entry_mode])
-  end
+  def changeset_for_entry_mode(%Hub{} = hub, entry_mode),
+    do: hub |> cast(%{entry_mode: entry_mode}, [:entry_mode])
 
-  def changeset_for_new_host(%Hub{} = hub, host) do
-    hub |> cast(%{host: host}, [:host])
-  end
+  def changeset_for_new_host(%Hub{} = hub, host), do: hub |> cast(%{host: host}, [:host])
 
   def add_account_to_changeset(changeset, nil), do: changeset
 
@@ -253,6 +249,7 @@ defmodule Ret.Hub do
     %{
       join_hub: account |> can?(join_hub(hub)),
       update_hub: account |> can?(update_hub(hub)),
+      close_hub: account |> can?(close_hub(hub)),
       kick_users: account |> can?(kick_users(hub)),
       mute_users: account |> can?(mute_users(hub))
     }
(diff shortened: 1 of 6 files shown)


test/ret_web/controllers/api/projects_controller_test.exs
```diff
@@ -17,7 +17,7 @@ defmodule RetWeb.ProjectsControllerTest do
   end
 
   @tag :authenticated
-  test "projects index works when logged in", %{conn: conn, project: project} do
+  test "projects index works when logged in", %{conn: conn, project: _project} do
```
gfodor: formatter issues/warnings

lib/ret_web/channels/hub_channel.ex
```diff
@@ -357,6 +361,21 @@ defmodule RetWeb.HubChannel do
...
 
+  defp handle_entry_mode_change(socket, entry_mode) do
+    hub = socket |> hub_for_socket
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+
+    if account |> can?(close_hub(hub)) do
+      hub
+      |> Hub.changeset_for_entry_mode(entry_mode)
+      |> Repo.update!()
+      |> Repo.preload(@hub_preloads, force: true)
```
brianpeiris: Why was the `force` necessary?

gfodor: copy pasta, can see if i can understand why it was done elsewhere.

lib/ret/hub.ex
```diff
@@ -253,20 +249,24 @@ defmodule Ret.Hub do
...
+      close_hub: account |> can?(close_hub(hub)),
       kick_users: account |> can?(kick_users(hub)),
       mute_users: account |> can?(mute_users(hub))
     }
   end
 end
 
 defimpl Canada.Can, for: Ret.Account do
+  # Always deny access to non-enterable hubs
+  def can?(%Ret.Account{}, :join_hub, %Ret.Hub{entry_mode: :deny}), do: false
```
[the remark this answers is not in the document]
gfodor: ah yep noted in other PR -- I decided against this due to the complexity involved and questions about the implications of having it be possible to join/re-open a closed room. for example, i could imagine scenarios where an administrator's account has been exploited and this is a fail-safe to always be able to shut everything down at the room level.

lib/ret_web/channels/hub_channel.ex
brianpeiris: Ah, I think I remember why I put the force in in the original code -- in that instance we were actually changing the scene association, so the preload update had to be forced, but I think the force isn't necessary here since you're only modifying a field on the hub itself.

lib/ret/hub.ex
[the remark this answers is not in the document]
brianpeiris: Ah, I hadn't considered the security concerns. Make sense.

PR #161, opened 2019-04-24
Avatar search and thumbnails

netpro2k:
Adds a thumbnail owned file to avatar records and an "avatars" source to the search API. The avatars source currently only supports returning results by user. Also, the only user you are currently authorized to search for is the one you are authenticated as.

todo:
- [x] decide if its ok to get a listing of other people's avatars, restrict it if not
- [x] decide what to call the source, and if we want a "filter"
- [x] come up with a real placeholder image

### DIFF
--- a/lib/ret/avatar.ex
+++ b/lib/ret/avatar.ex
@@ -22,7 +22,7 @@ defmodule Ret.Avatar do
     :normal_map_owned_file,
     :orm_map_owned_file
   ]
-  @file_columns [:gltf_owned_file, :bin_owned_file] ++ @image_columns
+  @file_columns [:gltf_owned_file, :bin_owned_file, :thumbnail_owned_file] ++ @image_columns
 
   def image_columns, do: @image_columns
   def file_columns, do: @file_columns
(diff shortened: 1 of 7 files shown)


lib/ret/avatar.ex
```diff
@@ -93,6 +94,29 @@ defmodule Ret.Avatar do
     |> avatar_to_collapsed_files()
   end
 
+  def version(%Avatar{} = avatar) do
+    avatar.updated_at |> NaiveDateTime.to_erl |> :calendar.datetime_to_gregorian_seconds
+  end
+
+  def url(%Avatar{} = avatar) do
```
gfodor: nit: you can re-write these to be one line:

`def url(%Avatar{} = avatar), do: "#{RetWeb.Endpoint.url()}/api/v1/avatars/#{avatar.avatar_sid}"`

lib/ret/media_search.ex
```diff
@@ -216,7 +242,8 @@ defmodule Ret.MediaSearch do
         )
 
       res =
-        "https://api.twitch.tv/kraken/search/streams?#{query}" |> retry_get_until_success([{"Client-ID", client_id}])
+        "https://api.twitch.tv/kraken/search/streams?#{query}"
+        |> retry_get_until_success([{"Client-ID", client_id}])
```
brianpeiris: Is your editor not respecting the 120 line length in the `.formatter.exs` file?

netpro2k: ah must not be. will take a look as to why not

netpro2k: Yep that was the issue

PR #163, opened 2019-04-24
Reticulum local deploys + configs

gfodor:
This PR adds some necessary configs for supporting the local reticulum in hab case, and also sets up hot config reloading.


gfodor:
goes with https://github.com/mozilla/hubs-ops/pull/53

gfodor:
pulls in https://github.com/mozilla/reticulum/pull/162

PR #169, opened 2019-05-02
Avatar listings

netpro2k:
This PR adds the AvatarListings model as and updates the Avatars model to support listings (with the reviewed_at and parent_avatar_listing fields).

It also expands the functionality of the AvatarController to support setting the other avatar fields (description, allow_promotion, parent_avatar_listing, etc) as well as deleting avatars.

The plan is to push this out along with the admin panel for avatars, set up the base bot as a AvatarListing, and then make a PR to update existing avatars pointing at the base bot avatar to point at the base bot avatar listing instead, using it for cascading instead of parent_avatar when available.

### DIFF
--- a/priv/repo/migrations/20190501221754_create_avatar_listings.exs
+++ b/priv/repo/migrations/20190501221754_create_avatar_listings.exs
@@ -34,7 +34,9 @@ defmodule Ret.Repo.Migrations.CreateAvatarListings do
 
     alter table(:avatars) do
       add(:parent_avatar_listing_id, references(:avatar_listings, column: :avatar_listing_id))
+      add(:reviewed_at, :utc_datetime, null: true)
     end
+    create(index(:avatars, [:reviewed_at], where: "reviewed_at is null or reviewed_at < updated_at"))
     drop constraint(:avatars, :gltf_or_parent)
     create constraint(:avatars, :gltf_or_parent_or_parent_listing, check: "parent_avatar_id is not null or parent_avatar_listing_id is not null or (gltf_owned_file_id is not null and bin_owned_file_id is not null)")
   end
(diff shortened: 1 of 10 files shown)


priv/repo/migrations/20190501221754_create_avatar_listings.exs
```diff
@@ -0,0 +1,40 @@
+defmodule Ret.Repo.Migrations.CreateAvatarListings do
+  use Ecto.Migration
+
+  def change do
+    Ret.AvatarListing.State.create_type()
+
+    create table(:avatar_listings, prefix: "ret0", primary_key: false) do
+      add(:avatar_listing_id, :bigint, default: fragment("ret0.next_id()"), primary_key: true)
+      add(:avatar_listing_sid, :string)
```
gfodor: `null: false`?

priv/repo/migrations/20190501221754_create_avatar_listings.exs
```diff
@@ -0,0 +1,40 @@
...
+      add(:slug, :string, null: false)
+      add(:order, :integer)
```
gfodor: i'd match it to scene listings, but without looking i'd have made it `default: 0, null: false`

priv/repo/migrations/20190501221754_create_avatar_listings.exs
```diff
@@ -0,0 +1,40 @@
...
+      add(:state, :avatar_listing_state, null: false, default: "active")
+      add(:tags, :jsonb)
+      add(:avatar_id, :bigint, null: false)
+      timestamps()
```
gfodor: i think more idiomatic to have this at the end

priv/repo/migrations/20190501221754_create_avatar_listings.exs
netpro2k: Yeah was thinking that, but scene_listings were declared like this, so I figured it would be cleaner to make them the same and change both later if we wanted to.

priv/repo/migrations/20190501221754_create_avatar_listings.exs
netpro2k: Hmm copped from scene_listings, but that does seem right

priv/repo/migrations/20190501221754_create_avatar_listings.exs
netpro2k: Was separating fields that are conceptually part of the listing and fields that are just cloned from the avatar record the listing is for. Probably not a super important distinction.

PR #171, opened 2019-05-04
Feature/creator assign tokens

gfodor:
Adds a one-time `creator_assignment_token` to new hubs, which is returned from the create API when a hub is created when signed out. This token can be passed in when calling `sign_in` on a `hub_channel` in order to assign the ownership of the hub to the newly logged in user.

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -92,14 +92,14 @@ defmodule Ret.Hub do
 
   def changeset_for_new_scene(%Hub{} = hub, %Scene{} = scene) do
     hub
-    |> cast(%{}, [])
+    |> change()
     |> put_change(:scene_id, scene.scene_id)
     |> put_change(:scene_listing_id, nil)
   end
 
   def changeset_for_new_scene(%Hub{} = hub, %SceneListing{} = scene_listing) do
     hub
-    |> cast(%{}, [])
+    |> change()
     |> put_change(:scene_listing_id, scene_listing.scene_listing_id)
     |> put_change(:scene_id, nil)
   end
@@ -131,10 +131,10 @@ defmodule Ret.Hub do
         token
       )
       when expected_token != nil and expected_token == token do
-    hub |> cast(%{}, []) |> add_account_to_changeset(account)
+    hub |> change() |> add_account_to_changeset(account)
   end
 
-  def changeset_for_creator_assignment(hub, _, _), do: hub |> cast(%{}, [])
+  def changeset_for_creator_assignment(hub, _, _), do: hub |> change()
 
   def add_account_to_changeset(changeset, nil), do: changeset
 
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -29,6 +29,7 @@ defmodule Ret.Hub do
     field(:host, :string)
     field(:entry_code, :integer)
     field(:entry_code_expires_at, :utc_datetime)
+    field(:creator_assignment_token, :string)
     field(:default_environment_gltf_bundle_url, :string)
     field(:slug, HubSlug.Type)
     field(:max_occupant_count, :integer, default: 0)
@@ -67,6 +68,7 @@ defmodule Ret.Hub do
     |> cast(attrs, [:default_environment_gltf_bundle_url])
     |> add_name_to_changeset(attrs)
     |> add_hub_sid_to_changeset
+    |> add_creator_assignment_token_to_changeset
     |> add_entry_code_to_changeset
     |> unique_constraint(:hub_sid)
     |> unique_constraint(:entry_code)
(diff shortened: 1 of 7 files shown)


lib/ret/hub.ex
```diff
@@ -123,10 +125,21 @@ defmodule Ret.Hub do
 
   def changeset_for_new_host(%Hub{} = hub, host), do: hub |> cast(%{host: host}, [:host])
 
+  def changeset_for_creator_assignment(
+        %Ret.Hub{creator_assignment_token: expected_token} = hub,
+        account,
+        token
+      )
+      when expected_token != nil and expected_token == token do
+    hub |> cast(%{}, []) |> add_account_to_changeset(account)
```
brianpeiris: I've seen you use `cast` like this a few times, not sure if you're intentionally doing it but if you have no params to cast, you could just use [`Ecto.Changeset.change/2`](https://hexdocs.pm/ecto/2.2.11/Ecto.Changeset.html#change/2). Same in the overload below.

lib/ret_web/channels/hub_channel.ex
```diff
@@ -185,12 +186,23 @@ defmodule RetWeb.HubChannel do
...
-  def handle_in("sign_in", %{"token" => token}, socket) do
+  def handle_in("sign_in", %{"token" => token} = payload, socket) do
+    creator_assignment_token = payload["creator_assignment_token"]
+
     case Ret.Guardian.resource_from_token(token) do
       {:ok, %Account{} = account, _claims} ->
         socket = Guardian.Phoenix.Socket.put_current_resource(socket, account)
 
-        hub = socket |> hub_for_socket
+        hub = socket |> hub_for_socket |> Repo.preload(@hub_preloads)
```
brianpeiris: We're preloading a lot of things for just a `sign_in`. Maybe just restrict preloads to `created_by_account` in this case?

test/ret/hub_test.exs
```diff
@@ -65,4 +65,72 @@ defmodule Ret.HubTest do
...
     %{join_hub: true, update_hub: true, close_hub: true, mute_users: true} = hub |> Hub.perms_for_account(account)
   end
+
+  test "should have creator assignment token if no account assigned", %{scene: scene} do
+    {:ok, hub} = %Hub{} |> Hub.changeset(scene, %{name: "Test Hub"}) |> Repo.insert()
+
+    assert hub.creator_assignment_token != nil
+  end
+
+  test "show allow creator assignment if token is correct", %{scene: scene, account: account} do
```
brianpeiris:
```suggestion
  test "should allow creator assignment if token is correct", %{scene: scene, account: account} do
```

test/ret/hub_test.exs
```diff
@@ -65,4 +65,72 @@ defmodule Ret.HubTest do
...
+    hub =
+      hub
+      |> Repo.preload(created_by_account: [])
+      |> Hub.changeset_for_creator_assignment(account, hub.creator_assignment_token)
+      |> Repo.update!()
+
+    assert hub.created_by_account_id == account.account_id
+  end
+
+  test "show disallow creator assignment if token is incorrect", %{scene: scene, account: account} do
```
brianpeiris:
```suggestion
  test "should disallow creator assignment if token is incorrect", %{scene: scene, account: account} do
```

test/ret/hub_test.exs
```diff
@@ -65,4 +65,72 @@ defmodule Ret.HubTest do
...
+    hub =
+      hub
+      |> Repo.preload(created_by_account: [])
+      |> Hub.changeset_for_creator_assignment(account, "bad token")
+      |> Repo.update!()
+
+    assert hub.created_by_account_id == nil
+  end
+
+  test "show disallow creator assignment if token is already used", %{
```
brianpeiris:
```suggestion
  test "should disallow creator assignment if token is already used", %{
```

test/ret/hub_test.exs
```diff
@@ -65,4 +65,72 @@ defmodule Ret.HubTest do
...
+    hub =
+      hub
+      |> Repo.preload(created_by_account: [])
+      |> Hub.changeset_for_creator_assignment(account2, nil)
+      |> Repo.update!()
+
+    assert hub.created_by_account_id == account.account_id
+  end
+
+  test "should have creator assignment token if account assigned", %{account: account, scene: scene} do
```
brianpeiris:
```suggestion
  test "should not have creator assignment token if account assigned", %{account: account, scene: scene} do
```

lib/ret/hub.ex
gfodor: ah good to know

lib/ret_web/channels/hub_channel.ex
gfodor: i think this would be an optimization, and could introduce tricky bugs if we change the dependent objects over time, so I'd suggest we leave it as-is until we determine if any of these loads are bottlenecks. as i mentioned i'm generally not nervous about any of these kinds of queries over the long haul (hash indexes will be fast, we can add caching, we can add read repliacs, etc) many-to-many join tables are the only things i'd be a bit careful about if we expect a lot of entries since those may bottleneck out without sharding the data.

test/ret/hub_test.exs
gfodor: oops

PR #172, opened 2019-05-06
Feature/perms presence

gfodor:
Updates presence to include the connected client's permissions, which is used for badging them. Goes with  https://github.com/mozilla/hubs/pull/1306


lib/ret_web/channels/hub_channel.ex
```diff
@@ -446,9 +450,13 @@ defmodule RetWeb.HubChannel do
   end
 
   defp presence_meta_for_socket(socket) do
+    hub = socket |> hub_for_socket
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+
     socket.assigns
     |> maybe_override_display_name(socket)
-    |> Map.take([:presence, :profile, :context])
+    |> Map.put(:perms, hub |> Hub.perms_for_account(account))
```
brianpeiris: When we initially discussed badging, you suggested that we decouple badges from perms. I still think that's a good idea since the UI code would be simpler and, more importantly, we wouldn't have to broadcast a detailed list of exactly what each user is capable of, which feels like a security/privacy breach since that information could be used as a vector for abuse and harassment. We could then also use server logic to determine badges based on other information and not just perms.

So I would suggest we have a `Hub.badges_for_account` that returns a list of badges in presence like `[:moderator, :owner, :host, :premium]` or whatever we decide to have badges for in the future.

gfodor: yep I was thinking about this more -- I wasn't really sure if having the badges derived server side was right or not -- my longer term expectation is that what we'll return from the server is "roles", so maybe I will just factor it that way now. in practice today there will be no semantic difference between badges and roles, just the name. later on though the role may be used for more UI bits than the badges. for the discord OAuth the roles are "virtual roles" today but I think eventually the discord stuff will sync the higher level constructs and policies into our role based access control layer vs doing the live permission checks we do now directly against the discord policy.

PR #174, opened 2019-05-13
Add scene search by account

gfodor:


lib/ret_web/views/api/v1/avatar_view.ex
```diff
@@ -11,18 +11,19 @@ defmodule RetWeb.Api.V1.AvatarView do
   end
 
   def render_avatar(avatar) do
-    version = avatar.updated_at |> NaiveDateTime.to_erl |> :calendar.datetime_to_gregorian_seconds
+    # version = avatar.updated_at |> NaiveDateTime.to_erl |> :calendar.datetime_to_gregorian_seconds
```
netpro2k: Think this got left in by mistake before I factored it out into a function, you can just remove instead of commenting.

PR #176, opened 2019-05-14
Add last active at telemetry

gfodor:
Our KPI we want to track is monthly active rooms, which was not possible to do accurately without keeping a timestamp of the last time a room became active.

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -29,6 +29,7 @@ defmodule Ret.Hub do
     field(:host, :string)
     field(:entry_code, :integer)
     field(:entry_code_expires_at, :utc_datetime)
+    field(:last_active_at, :utc_datetime)
     field(:creator_assignment_token, :string)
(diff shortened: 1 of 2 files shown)


lib/ret/hub.ex
```diff
@@ -87,9 +88,16 @@ defmodule Ret.Hub do
 
     hub
     |> cast(%{max_occupant_count: new_max_occupant_count}, [:max_occupant_count])
+    |> maybe_add_last_active_at_to_changeset(occupant_count)
     |> validate_required([:max_occupant_count])
   end
 
+  # NOTE occupant_count is 1 when there is 1 *other* user in the room with you, so active is when >= 1.
+  defp maybe_add_last_active_at_to_changeset(changeset, occupant_count) when occupant_count >= 1,
```
netpro2k: Lol I read the comment, then read the next line and was still confused by the `occupant_count >= 1`

PR #178, opened 2019-05-16
Friendly display name

brianpeiris:
- Discord displayNames no longer contain the discriminator number
- We also now lookup the server-specific "nickname" and use that if available
- Injects a new communityIdentifier field into the presence profile that includes the discord discriminator

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -473,18 +473,24 @@ defmodule RetWeb.HubChannel do
   defp maybe_override_display_name(
          %{
            hub_requires_oauth: true,
+           hub_sid: hub_sid,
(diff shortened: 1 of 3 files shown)


lib/ret_web/channels/hub_channel.ex
```diff
@@ -495,29 +500,45 @@ defmodule RetWeb.HubChannel do
...
   defp maybe_override_display_name(
          %{
            hub_requires_oauth: false
          } = assigns,
-         _socket
+         _account
        ),
        do: assigns
 
+  defp override_display_name(assigns, oauth_provider, hub_binding) do
```
gfodor: i'd consider renaming this function to `override_display_name_via_binding`

gfodor:
lgtm minor comment

PR #179, opened 2019-05-20
Endpoints for Avatar Listings

netpro2k:
This adds endpoinds for avatar listings similar to how we handle scenes/scene_listings. `/ap1/v1/avatars` can now take either and avatar or avatar listing sid (yuck, but keeping it consistent with scenes for now).

This PR also adds support for using an avatar listing as a parent avatar, and simplifies the semantics of the collapsing behavior to only support a single level of parenting.

### DIFF
--- a/lib/ret/avatar.ex
+++ b/lib/ret/avatar.ex
@@ -70,24 +70,41 @@ defmodule Ret.Avatar do
     end)
   end
 
+  defp avatar_listing_to_collapsed_files(%{parent_avatar_listing: nil} = listing),
+    do: listing |> Map.take(Avatar.file_columns())
+
+  defp avatar_listing_to_collapsed_files(%{parent_avatar_listing: parent} = listing) do
+    parent
+    |> Repo.preload(Avatar.file_columns())
+    |> Map.take(Avatar.file_columns())
+    |> Map.merge(listing |> Map.take(Avatar.file_columns()), fn
+      _k, v1, nil -> v1
+      _k, _v1, v2 -> v2
+    end)
+  end
+
   def collapsed_files(%Avatar{} = avatar) do
     avatar
     |> Repo.preload([:parent_avatar, :parent_avatar_listing] ++ @file_columns)
     |> avatar_to_collapsed_files()
   end
 
-  def version(%Avatar{} = avatar) do
-    avatar.updated_at |> NaiveDateTime.to_erl() |> :calendar.datetime_to_gregorian_seconds()
+  def collapsed_files(%AvatarListing{} = listing) do
+    listing |> Repo.preload([:parent_avatar_listing] ++ Avatar.file_columns()) |> avatar_listing_to_collapsed_files()
   end
 
+  def version(%t{} = a) when t in [Avatar, AvatarListing],
+    do: a.updated_at |> NaiveDateTime.to_erl() |> :calendar.datetime_to_gregorian_seconds()
+
   def url(%Avatar{} = avatar), do: "#{RetWeb.Endpoint.url()}/api/v1/avatars/#{avatar.avatar_sid}"
+  def url(%AvatarListing{} = avatar), do: "#{RetWeb.Endpoint.url()}/api/v1/avatars/#{avatar.avatar_listing_sid}"
 
-  def gltf_url(%Avatar{} = avatar), do: "#{Avatar.url(avatar)}/avatar.gltf?v=#{Avatar.version(avatar)}"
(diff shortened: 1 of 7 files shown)


lib/ret/avatar_listing.ex
```diff
@@ -60,6 +62,22 @@ defmodule Ret.AvatarListing do
     end
   end
 
+  defp avatar_listing_to_collapsed_files(%{parent_avatar_listing: nil} = a), do: a |> Map.take(Avatar.file_columns())
+
+  defp avatar_listing_to_collapsed_files(%{parent_avatar_listing: p} = a) do
```
brianpeiris: nit: use full names instead of `p` and `a`. Makes for easier reading and grepping.

lib/ret_web/views/api/v1/avatar_view.ex
```diff
@@ -10,23 +10,41 @@ defmodule RetWeb.Api.V1.AvatarView do
     %{avatars: [render_avatar(avatar)]}
   end
 
-  def render_avatar(avatar) do
+  def render_avatar(%Avatar{} = a) do
```
brianpeiris: nit: `defp`

lib/ret/avatar_listing.ex
```diff
@@ -49,9 +49,11 @@ defmodule Ret.AvatarListing do
 
   def url(%AvatarListing{} = avatar), do: "#{RetWeb.Endpoint.url()}/api/v1/avatars/#{avatar.avatar_listing_sid}"
 
-  def gltf_url(%AvatarListing{} = avatar), do: "#{AvatarListing.url(avatar)}/avatar.gltf?v=#{AvatarListing.version(avatar)}"
+  def gltf_url(%AvatarListing{} = avatar),
```
brianpeiris: These utility functions are almost identical between `avatar` and `avatar_listing`. It would be better to adopt the type guard form that we use for scenes:
https://github.com/mozilla/reticulum/blob/d3578d862f7d7593ee3399c020ed8f53e7ca2c93/lib/ret/scene.ex#L43

lib/ret_web/views/api/v1/avatar_view.ex
```diff
@@ -10,23 +10,41 @@ defmodule RetWeb.Api.V1.AvatarView do
...
+      description: a.description,
+      attributions: if(is_nil(a.attributions), do: %{}, else: a.attributions),
+      allow_remixing: a.avatar.allow_remixing,
+      allow_promotion: a.avatar.allow_promotion,
+      gltf_url: a |> AvatarListing.gltf_url(),
+      base_gltf_url: a |> AvatarListing.base_gltf_url(),
+      files:
+        for col <- Avatar.file_columns(), into: %{} do
+          key = col |> Atom.to_string() |> String.replace_suffix("_owned_file", "")
+          {key, a |> AvatarListing.file_url_or_nil(col)}
```
brianpeiris: Would be good to follow the form we have in `scene_view`, where we refactor common fields where possible. Comparing the two, it also seems you're missing guards for preventing views for removed or delisted avtars.

netpro2k: Hmm agree this should be accounted for, but seems odd to do it in the view instead of the controller. Will add checks to the controller.

brianpeiris: I think the reason we did it in the view for scenes is because `render_scene` is used from `hub_view`. Doesn't apply to avatars for now, I guess.

lib/ret/avatar.ex
```diff
@@ -70,24 +70,41 @@ defmodule Ret.Avatar do
...
+
+  defp avatar_listing_to_collapsed_files(%{parent_avatar_listing: parent} = listing) do
+    parent
+    |> Repo.preload(Avatar.file_columns())
+    |> Map.take(Avatar.file_columns())
+    |> Map.merge(listing |> Map.take(Avatar.file_columns()), fn
+      _k, v1, nil -> v1
+      _k, _v1, v2 -> v2
+    end)
+  end
```
brianpeiris: Seems like this can be DRYed some more with `avatar_to_collapsed_files` now that they are in the same module. At least now you can just do `@file_columns` instead of `Avatar.file_columns()`, but everything past line 77 here is identical in `avatar_to_collapsed_files`. Maybe just nitpicking now though, not a big deal.

netpro2k: Its the pattern matching of the argumetns that differs, could break out just the body but ends up being basically just as much code anyway. Will filx `@file_columns` though

PR #182, opened 2019-05-29
Remove characters from sids that are difficult to distinguish

brianpeiris:
Add `1Il` and `O0` to list of undesirable characters, since they are hard to distinguish in most fonts. This also changes the algorithm to draw from a longer list of crypto random characters, instead of simply replacing characters with random digits.

### DIFF
--- a/lib/ret/sids.ex
+++ b/lib/ret/sids.ex
@@ -1,11 +1,11 @@
 defmodule Ret.Sids do
-  @num_random_bits_for_sid 16
+  @num_random_bits_for_sid 32
 
   def generate_sid() do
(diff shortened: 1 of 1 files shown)


brianpeiris:
Marshall also suggested making the SIDs lowercase-only, which would definitely be nicer, but I'm not sure if we want to go that far. Eliminating these particular characters gets rid of the most annoying cases.

gfodor:
oh yeah, the stripping part does that. i'd probably replace it with some filler chars so its always exactly 7 chars

gfodor:
Actually this was a mistake on my part. Igtm

PR #183, opened 2019-05-30
Add Avatar landing pages

netpro2k:
Adds routes + meta for avatar landing pages. Avatar responses now point to these pages in their `url` property.

### DIFF
--- a/lib/ret_web/templates/page/avatar-meta.html.eex
+++ b/lib/ret_web/templates/page/avatar-meta.html.eex
@@ -1,12 +1,12 @@
 <meta property="og:type" content="website" />
 <meta property="og:url" content="<%= @avatar |> Ret.Avatar.url %>" />
 <meta property="og:title" content="<%= @avatar.name %> | Hubs by Mozilla" />
-<meta property="og:description" content="<%= @avatar.description %>" />
+<meta property="og:description" content="&quot;<%= @avatar.name %>&quot; is an avatar you can use in Hubs." />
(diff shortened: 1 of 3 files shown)


lib/ret_web/templates/page/avatar-meta.html.eex
```diff
@@ -0,0 +1,16 @@
+<meta property="og:type" content="website" />
+<meta property="og:url" content="<%= @avatar |> Ret.Avatar.url %>" />
+<meta property="og:title" content="<%= @avatar.name %> | Hubs by Mozilla" />
+<meta property="og:description" content="<%= @avatar.description %>" />
```
brianpeiris: Why not have this match the twitter description?

lib/ret_web/templates/page/avatar-meta.html.eex
```diff
@@ -0,0 +1,16 @@
...
+<meta property="og:image" content="<%= @avatar.thumbnail_owned_file |> Ret.OwnedFile.uri_for |> URI.to_string %>"/>
+<meta name="twitter:card" content="summary_large_image">
+<meta name="twitter:domain" value="<%= RetWeb.Endpoint.host %>" />
+<meta name="twitter:title" value="<%= @avatar.name %> | Hubs by Mozilla" />
+<meta name="twitter:description" content="<%= @avatar.name %> is an avatar you can use in Hubs." />
```
brianpeiris: Maybe put the avatar name in quotes? So it reads like `"Foo Bar" is an avatar you can use in Hubs.`

PR #186, opened 2019-06-05
Feature/embeds

gfodor:
Adds the necessary bits for embeds:

- Adds the ability to turn on/off NAF messages on a socket (so embeds can opt-out of them until entry)

- Adds an `embed_token` which is generated when the room is created.

This embed token has the following behavior:
  - The `embed_token` is revealed at hub create time, or at hub join time to anyone who has the `hub_embed` permission (currently, the owner. discord context is excluded since embeds don't make much sense there.)

  - The room page serving checks the embed token, and doesn't render on a mismatch, and strips the `X-Frame-Options` header if it does match

- When a hub is joined, the presence context for the user now includes a `embed` flag which will flip a telemetry bit on the room so we know it was used as an embed.

Note this patch assumes we'll be stripping `frame-ancestors` from the CSP and using `X-Frame-Options` as the method for restricting embeds.

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -31,6 +31,8 @@ defmodule Ret.Hub do
     field(:entry_code_expires_at, :utc_datetime)
     field(:last_active_at, :utc_datetime)
     field(:creator_assignment_token, :string)
+    field(:embed_token, :string)
+    field(:embedded, :boolean)
     field(:default_environment_gltf_bundle_url, :string)
     field(:slug, HubSlug.Type)
     field(:max_occupant_count, :integer, default: 0)
@@ -69,7 +71,7 @@ defmodule Ret.Hub do
     |> cast(attrs, [:default_environment_gltf_bundle_url])
     |> add_name_to_changeset(attrs)
     |> add_hub_sid_to_changeset
-    |> add_creator_assignment_token_to_changeset
+    |> add_generated_tokens_to_changeset
     |> add_entry_code_to_changeset
     |> unique_constraint(:hub_sid)
     |> unique_constraint(:entry_code)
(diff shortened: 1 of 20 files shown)


lib/ret_web/controllers/page_controller.ex
```diff
@@ -60,33 +60,48 @@ defmodule RetWeb.PageController do
...
+    embed_token = params["embed_token"]
 
-  def render_for_path("/" <> path, conn) do
     [hub_sid | subresource] = path |> String.split("/")
 
     hub = Hub |> Repo.get_by(hub_sid: hub_sid)
-    render_hub_content(conn, hub, subresource |> Enum.at(0))
+
+    if embed_token && hub.embed_token != embed_token do
+      conn |> send_resp(404, "")
```
brianpeiris: nit: Might be better to render a nice page or at least respond with a useful message.

lib/ret/hub.ex
```diff
@@ -31,6 +31,8 @@ defmodule Ret.Hub do
     field(:entry_code_expires_at, :utc_datetime)
     field(:last_active_at, :utc_datetime)
     field(:creator_assignment_token, :string)
+    field(:embed_token, :string)
+    field(:embedded, :boolean)
```
brianpeiris: I guess this will just be used for our own analytics?

gfodor:
Yep I agree, lets consider permissions around pinning as a minimum required to ship this (ideally more, but that's where I think I'd OK given we'll probably have time before it gets used much anyway.)

lib/ret/hub.ex
gfodor: yep, this is just a telemetry bit

PR #192, opened 2019-06-24
Feature/photo sharing

gfodor:
Adds Twitter OAuth to reticulum, and adds support for posting tweets + uploading media to twitter. 

Note this also pulls in https://github.com/mozilla/reticulum/pull/193 so I could test it so you should review that first.


config/dev.exs
```diff
@@ -139,11 +139,11 @@ config :ret, RetWeb.AddCSPPlug,
...
       asset_hosts
-    }; style-src 'self' https://fonts.googleapis.com #{asset_hosts} 'unsafe-inline'; connect-src 'self' https://sentry.prod.mozaws.net https://dpdb.webvr.rocks #{
+    }; style-src 'self' https://fonts.googleapis.com https://cdn.jsdelivr.net #{asset_hosts} 'unsafe-inline'; connect-src 'self' https://sentry.prod.mozaws.net https://dpdb.webvr.rocks #{
       asset_hosts
-    } #{websocket_hosts} https://cdn.aframe.io https://www.mozilla.org data: blob:; img-src 'self' #{asset_hosts} https://cdn.aframe.io data: blob:; media-src 'self' #{
+    } #{websocket_hosts} https://cdn.aframe.io https://www.mozilla.org data: blob:; img-src 'self' #{
       asset_hosts
-    } data: blob:; frame-src 'self'; base-uri 'none'; form-action 'self';"
+    } https://cdn.aframe.io https://cdn.jsdelivr.net data: blob:; media-src 'self' #{asset_hosts} data: blob:; frame-src 'self'; base-uri 'none'; form-action 'self';"
```
gfodor: this new CDN is used for draftJS

config/dev.exs
```diff
@@ -139,11 +139,11 @@ config :ret, RetWeb.AddCSPPlug,
   content_security_policy:
     "default-src 'none'; script-src 'self' #{script_shas} #{asset_hosts} https://cdn.rawgit.com https://aframe.io https://www.google-analytics.com 'unsafe-eval'; worker-src 'self' blob:; font-src 'self' https://fonts.googleapis.com https://fonts.gstatic.com https://cdn.aframe.io #{
       asset_hosts
-    }; style-src 'self' https://fonts.googleapis.com #{asset_hosts} 'unsafe-inline'; connect-src 'self' https://sentry.prod.mozaws.net https://dpdb.webvr.rocks #{
+    }; style-src 'self' https://fonts.googleapis.com https://cdn.jsdelivr.net #{asset_hosts} 'unsafe-inline'; connect-src 'self' https://sentry.prod.mozaws.net https://dpdb.webvr.rocks #{
```
gfodor: this new CDN is where the emoji images are hosted. it will only need to be added to img-src and style-src so relatively safe.

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -69,6 +97,30 @@ defmodule RetWeb.Api.V1.OAuthController do
...
+        provider_access_token_secret: access_token_secret,
+        provider_account_id: twitter_user_id
+      )
+      |> Repo.insert_or_update()
+
+      conn
+    else
+      conn |> send_resp(401, "Another account is already connected to this twitter account.")
+    end
+  end
```
brianpeiris: Okay, I think I'm just thoroughly confused by this logic, so you'll have to walk me through it.

lib/ret_web/controllers/twitter_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+        [stored_file_uuid, _ext] = filename |> String.split(".")
+        parsed_query = query |> URI.decode_query()
+        stored_file_access_token = parsed_query["token"]
+
+        case TwitterClient.upload_stored_file_as_media(
+               stored_file_uuid,
+               stored_file_access_token,
+               account,
+               token,
+               token_secret
```
brianpeiris: Might make more sense for TwitterClient functions to accept an OAuthProvider instead of separate params for the token and secret

lib/ret_web/controllers/twitter_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+          media_id when is_binary(media_id) ->
+            res = TwitterClient.tweet(body, token, token_secret, media_id)
+            conn |> send_resp(200, res |> Poison.encode!())
+
+          _ ->
+            conn |> send_resp(400, "Failed uploading")
+        end
+
+      _ ->
+        conn |> send_resp(400, "Invalid media stored file url")
```
brianpeiris: Would now be a good time to start returning JSON error responses for endpoints that would normally return sucessful JSON responses?

lib/ret/media_resolver.ex
```diff
@@ -204,13 +204,17 @@ defmodule Ret.MediaResolver do
...
+              :error -> uri |> og_tag_commit_for_uri()
+              commit -> commit
+            end
+          else
+            uri |> og_tag_commit_for_uri()
           end
         else
-          uri |> og_tag_commit_for_uri()
+          {:commit, uri |> resolved(%{expected_content_type: content_type})}
         end
```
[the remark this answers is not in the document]
gfodor: this changes the behavior for local URLs -- previously, if you had the lambda setup, it would not crawl a local URL and just return it. but if you didn't have the lambda setup (as is common in the local dev case) then it would crawl the URL, which was wrong.

lib/ret_web/controllers/api/v1/oauth_controller.ex
gfodor: I believe I wrote `!= account.account_id` instead of `== account.account_id` -- oops. the inner `if` block should run if there's no oauth provider for this account already or if there is, and it belongs to this account. (the inner if block can also be slightly refactored I think)

lib/ret/media_resolver.ex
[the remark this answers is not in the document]
brianpeiris: Hmm, doesn't the change to the outer else condition affect non-local URLs as well? Previously we would return the og_tag_commit if the content_type was not text/html, but now we return the resolved URI.

lib/ret_web/controllers/api/v1/oauth_controller.ex
brianpeiris: Ah, okay, that clears things up 👍

lib/ret/media_resolver.ex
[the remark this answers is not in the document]
gfodor: ah that also seems like a bug, because we shouldn't be crawling pages that aren't text/html for OG tags (i don't think so, at least)

PR #194, opened 2019-06-25
Broadcast email over auth channel after token verified

gfodor:
This updates the auth channel protocol to also broadcast out the email address from the originating request after the token is verified, so the page which is verifying can also be logged in.

### DIFF
--- a/lib/ret_web/channels/auth_channel.ex
+++ b/lib/ret_web/channels/auth_channel.ex
@@ -20,7 +20,7 @@ defmodule RetWeb.AuthChannel do
 
   def handle_in("auth_request", %{"email" => email, "origin" => origin}, socket) do
     if !Map.get(socket.assigns, :used) do
-      socket = socket |> assign(:used, true)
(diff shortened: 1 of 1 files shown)


lib/ret_web/channels/auth_channel.ex
```diff
@@ -65,6 +65,12 @@ defmodule RetWeb.AuthChannel do
   def handle_out("auth_credentials" = event, payload, socket) do
     Process.send_after(self(), :close_channel, 1000 * 5)
     push(socket, event, payload)
+
+    # Send the email address over for all connected processes to use as well
+    if socket.assigns |> Map.has_key?(:email) do
+      broadcast!(socket, "auth_email", %{email: socket.assigns.email})
+    end
```
brianpeiris: Why broadcast this in `handle_out` instead of in `broadcast_credentials_for_identifier_hash`?

brianpeiris: Oh, right, we're dealing with two separate processes here. nvm.

PR #195, opened 2019-07-01
Hub object permissions

brianpeiris:
Adds mechanisms for object-level permissions controls on Hubs
- Adds a `permissions` integer bit field to the Hubs table
- Defines `spawn_and_move_media`, `spawn_camera`, `spawn_drawing`, and `pin_objects` permissions
- Hubs have all these permissions allowed by default
- Hubs API now accepts and returns a `perms` map in JSON
- Adds Canada capability checks for these new permissions
- Ignores associated `"naf"` and `"pin"` hub channel messages if permissions are disabled

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -308,10 +308,6 @@ defmodule Ret.Hub do
 
   @hub_perms_keys @hub_perms |> Map.values()
 
-  def hub_perms_keys do
-    @hub_perms_keys
-  end
-
   def hub_perms_to_int!(%{} = perms) do
     invalid_perms = perms |> Map.drop(@hub_perms_keys) |> Map.keys()
 
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -290,9 +290,9 @@ defmodule Ret.Hub do
 
   defp add_default_perms_to_changeset(changeset) do
     default_perms = %{
-      create_media: true,
-      create_camera: true,
-      create_drawing: true,
+      spawn_and_move_media: true,
+      spawn_camera: true,
+      spawn_drawing: true,
       pin_objects: true
     }
 
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -431,7 +431,8 @@ defimpl Canada.Can, for: Ret.OAuthProvider do
   end
 
   # Object permissions for OAuthProvider users are based on perms settings
-  def can?(_account, action, hub) when action in [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects] do
+  def can?(%Ret.OAuthProvider{} = oauth_provider, action, %Ret.Hub{hub_bindings: hub_bindings} = hub)
+      when action in [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects] do
     is_member = hub_bindings |> Enum.any?(&(oauth_provider |> Ret.HubBinding.member_of_channel?(&1)))
     is_member and hub |> Hub.has_perm!(action)
   end
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -87,7 +87,7 @@ defmodule Ret.Hub do
     |> HubSlug.maybe_generate_slug()
   end
 
-  def add_perms_to_changeset(changeset, attrs) do
+  def add_member_permissions_to_changeset(changeset, attrs) do
     member_permissions =
       attrs["member_permissions"] |> Map.new(fn {k, v} -> {String.to_atom(k), v} end) |> member_permissions_to_int
 
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -284,9 +284,7 @@ defmodule Ret.Hub do
     0x0000_0001 => :spawn_and_manipulate_media,
     0x0000_0002 => :spawn_camera,
     0x0000_0004 => :spawn_drawing,
-    0x0000_0008 => :pin_own_object,
-    0x0000_0010 => :pin_any_object,
-    0x0000_0020 => :manipulate_any_object
+    0x0000_0008 => :pin_objects
   }
 
   @hub_perms_keys @hub_perms |> Map.values()
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -432,7 +432,8 @@ defimpl Canada.Can, for: Ret.OAuthProvider do
 
   # Object permissions for OAuthProvider users are based on perms settings
   def can?(_account, action, hub) when action in [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects] do
-    hub |> Hub.has_perm!(action)
+    is_member = hub_bindings |> Enum.any?(&(oauth_provider |> Ret.HubBinding.member_of_channel?(&1)))
+    is_member and hub |> Hub.has_perm!(action)
   end
 
   def can?(_, _, _), do: false
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -354,6 +354,8 @@ end
 
 defimpl Canada.Can, for: Ret.Account do
   alias Ret.{Hub}
+  @object_actions [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects]
+  @special_actions [:update_hub, :close_hub, :embed_hub, :kick_users, :mute_users] ++ @object_actions
 
   # Always deny access to non-enterable hubs
   def can?(%Ret.Account{}, :join_hub, %Ret.Hub{entry_mode: :deny}), do: false
@@ -370,17 +372,22 @@ defimpl Canada.Can, for: Ret.Account do
     hub_bindings |> Enum.any?(&(account |> Ret.HubBinding.can_manage_channel?(&1)))
   end
 
-  # Bound hubs - Moderator and object actions
+  # Bound hubs - Moderator actions
+  def can?(%Ret.Account{} = account, action, %Ret.Hub{hub_bindings: hub_bindings})
+      when hub_bindings |> length > 0 and action in [:kick_users, :mute_users] do
+    hub_bindings |> Enum.any?(&(account |> Ret.HubBinding.can_moderate_users?(&1)))
+  end
+
+  # Bound hubs - Object permissions
   def can?(%Ret.Account{} = account, action, %Ret.Hub{hub_bindings: hub_bindings} = hub)
-      when hub_bindings |> length > 0 and
-             action in [:kick_users, :mute_users, :spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects] do
+      when hub_bindings |> length > 0 and action in @object_actions do
     is_moderator = hub_bindings |> Enum.any?(&(account |> Ret.HubBinding.can_moderate_users?(&1)))
 
-    if !is_moderator and action in [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects] do
+    if is_moderator do
+      true
+    else
       is_member = hub_bindings |> Enum.any?(&(account |> Ret.HubBinding.member_of_channel?(&1)))
       is_member and hub |> Hub.has_perm!(action)
-    else
-      is_moderator
     end
   end
 
@@ -394,23 +401,11 @@ defimpl Canada.Can, for: Ret.Account do
 
   # Unbound hubs - Creators can perform special actions
   def can?(%Ret.Account{account_id: account_id}, action, %Ret.Hub{created_by_account_id: account_id})
-      when account_id != nil and
-             action in [
-               :update_hub,
-               :close_hub,
-               :embed_hub,
-               :kick_users,
-               :mute_users,
-               :spawn_and_move_media,
-               :spawn_camera,
-               :spawn_drawing,
-               :pin_objects
-             ],
+      when account_id != nil and action in @special_actions,
       do: true
 
   # Unbound hubs - Object permissions for regular users are based on perms settings
-  def can?(_account, action, hub)
-      when action in [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects] do
+  def can?(_account, action, hub) when action in @object_actions do
     hub |> Hub.has_perm!(action)
   end
 
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -74,6 +74,7 @@ defmodule Ret.Hub do
     |> add_hub_sid_to_changeset
     |> add_generated_tokens_to_changeset
     |> add_entry_code_to_changeset
+    |> add_default_perms_to_changeset
     |> unique_constraint(:hub_sid)
     |> unique_constraint(:entry_code)
   end
@@ -287,14 +288,29 @@ defmodule Ret.Hub do
     end
   end
 
+  defp add_default_perms_to_changeset(changeset) do
+    default_perms = %{
+      create_media: true,
+      create_camera: true,
+      create_drawing: true,
+      pin_objects: true
+    }
+
+    changeset |> put_change(:permissions, default_perms |> hub_perms_to_int!)
+  end
+
   @hub_perms %{
-    0x0000_0001 => :spawn_and_manipulate_media,
+    0x0000_0001 => :spawn_and_move_media,
     0x0000_0002 => :spawn_camera,
     0x0000_0004 => :spawn_drawing,
     0x0000_0008 => :pin_objects
   }
 
   @hub_perms_keys @hub_perms |> Map.values()
+
+  def hub_perms_keys do
+    @hub_perms_keys
+  end
 
   def hub_perms_to_int!(%{} = perms) do
     invalid_perms = perms |> Map.drop(@hub_perms_keys) |> Map.keys()
@@ -306,7 +322,7 @@ defmodule Ret.Hub do
     @hub_perms |> Enum.reduce(0, fn {val, perm}, acc -> if(perms[perm], do: 1, else: 0) * val + acc end)
   end
 
-  def hub_has_perm!(hub_perms_bit_field, perm) do
+  def has_perm!(%Hub{permissions: hub_perms_bit_field}, perm) do
     case @hub_perms |> Enum.find(fn {_, perm_name} -> perm_name == perm end) do
       nil -> raise ArgumentError, "Invalid permission #{perm}"
       {val, _} -> (hub_perms_bit_field &&& val) > 0
@@ -317,6 +333,7 @@ defmodule Ret.Hub do
     hub.permissions |> BitFieldUtils.permissions_to_map(@hub_perms)
   end
 
+  # The account argument here can be a Ret.Account, a Ret.OAuthProvider or nil.
   def perms_for_account(%Ret.Hub{} = hub, account) do
     %{
       join_hub: account |> can?(join_hub(hub)),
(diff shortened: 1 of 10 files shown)


lib/ret/hub.ex
```diff
@@ -279,14 +288,60 @@ defmodule Ret.Hub do
...
+      spawn_camera: true,
+      spawn_drawing: true,
+      pin_objects: true
+    }
+
+    changeset |> put_change(:permissions, default_perms |> hub_perms_to_int!)
+  end
+
+  @hub_perms %{
+    0x0000_0001 => :spawn_and_move_media,
```
gfodor: to reduce typo risk somewhat, I usually like writing these kinds of things like `1 << 0`, `1 << 1`, etc. minor tho

priv/repo/migrations/20190613223618_add_permissions_to_hubs.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Ret.Repo.Migrations.AddPermissionsToHubs do
+  use Ecto.Migration
+
+  def change do
+    alter table("hubs") do
+      add(:permissions, :integer)
```
gfodor: do we need to initialize these with the defaults?

lib/ret/hub.ex
```diff
@@ -33,6 +33,7 @@ defmodule Ret.Hub do
     field(:creator_assignment_token, :string)
     field(:embed_token, :string)
     field(:embedded, :boolean)
+    field(:permissions, :integer)
```
gfodor: worth considering a rename to `member_permissions` or `default_permissions` -- this could get a bit tricky to understand if we do end up having more complex policies down the road

lib/ret/hub.ex
```diff
@@ -298,41 +353,70 @@ defmodule Ret.Hub do
...
       when hub_bindings |> length > 0 and action in [:embed_hub],
       do: false
 
-  # Anyone can join an unbound hub
-  def can?(_, :join_hub, %Ret.Hub{hub_bindings: []}), do: true
+  # Unbound hubs - Anyone can join an unbound hub
+  def can?(_account, :join_hub, %Ret.Hub{hub_bindings: []}), do: true
 
-  # Creators of unbound hubs can perform special actions
+  # Unbound hubs - Creators can perform special actions
```
gfodor: s/Creator/Owner

lib/ret/hub.ex
```diff
@@ -298,41 +353,70 @@ defmodule Ret.Hub do
...
   def can?(%Ret.Account{account_id: account_id}, action, %Ret.Hub{created_by_account_id: account_id})
-      when account_id != nil and action in [:update_hub, :close_hub, :embed_hub, :kick_users, :mute_users],
+      when account_id != nil and action in @special_actions,
       do: true
 
+  # Unbound hubs - Object permissions for regular users are based on perms settings
+  def can?(_account, action, hub) when action in @object_actions do
```
gfodor: I think I mentioned this before, but I feel like this module is getting long enough we may want to explicitly check for no bindings in the guard clause. I'm nervous about the inadvertent fallthrough risk.

lib/ret/hub.ex
```diff
@@ -279,14 +288,60 @@ defmodule Ret.Hub do
...
+    invalid_perms = perms |> Map.drop(@hub_perms_keys) |> Map.keys()
+
+    if invalid_perms |> Enum.count() > 0 do
+      raise ArgumentError, "Invalid permissions #{invalid_perms |> Enum.join(", ")}"
+    end
+
+    @hub_perms |> Enum.reduce(0, fn {val, perm}, acc -> if(perms[perm], do: 1, else: 0) * val + acc end)
+  end
+
+  def has_perm!(%Hub{permissions: hub_perms_bit_field}, perm) do
```
gfodor: i find this function name fairly confusing, since the `!` implies to me it will throw if the permission is not granted -- I'd name it `has_perm?`. i think bang is reserved for error conditions, not argument validation (where, for example, you'd typically match on err for the non-bang version)

also if we decide to rename the column, I'd call this `has_member_perm?` or something aligned with the new column name

lib/ret/hub.ex
```diff
@@ -279,14 +288,60 @@ defmodule Ret.Hub do
...
+    0x0000_0008 => :pin_objects
+  }
+
+  @hub_perms_keys @hub_perms |> Map.values()
+
+  def hub_perms_to_int!(%{} = perms) do
+    invalid_perms = perms |> Map.drop(@hub_perms_keys) |> Map.keys()
+
+    if invalid_perms |> Enum.count() > 0 do
+      raise ArgumentError, "Invalid permissions #{invalid_perms |> Enum.join(", ")}"
```
gfodor: I'd drop the `!` here since it's argument validation that throws, not some kind of error

lib/ret_web/channels/hub_channel.ex
```diff
@@ -140,12 +140,32 @@ defmodule RetWeb.HubChannel do
     {:noreply, socket}
   end
 
-  def handle_in("naf" = event, %{"data" => %{"isFirstSync" => true}} = payload, socket) do
-    data =
-      payload["data"] |> Map.put("creator", socket.assigns.session_id) |> Map.put("owner", socket.assigns.session_id)
+  def handle_in("naf" = event, %{"data" => %{"isFirstSync" => true, "template" => naf_template}} = payload, socket) do
```
gfodor: i'd add a comment noting that this match captures all inbound NAF messages that result in spawned objects

lib/ret_web/channels/hub_channel.ex
```diff
@@ -140,12 +140,32 @@ defmodule RetWeb.HubChannel do
...
+
+    should_broadcast =
+      case naf_template do
+        "#interactable-media" -> account |> can?(spawn_and_move_media(hub))
+        "#static-media" -> account |> can?(spawn_and_move_media(hub))
+        "#static-controlled-media" -> account |> can?(spawn_and_move_media(hub))
+        "#interactable-camera" -> account |> can?(spawn_camera(hub))
+        "#interactable-drawing" -> account |> can?(spawn_drawing(hub))
+        "#pen-interactable" -> account |> can?(spawn_drawing(hub))
+        _ -> true
```
gfodor: hmm. one failure mode here is if a developer renames a template or introduces a new one (similar to what happened say when we split out static/interactable media) then it's going to suddenly "disable" the security checks.

maybe the right balance to minimize the need to deploy reticulum when refactoring/adding templates is to make this suffix based. If it's a recognized suffix, then do the permission check. If it's not, then deny. so something like:

`-media` - Media permission
`-camera` - Camera permission
`-drawing` or `-pen` - Drawing permission, and we'll need to refactor the client to have `interactable-pen` as the template

lib/ret_web/channels/hub_channel.ex
```diff
@@ -249,15 +312,23 @@ defmodule RetWeb.HubChannel do
         socket
       ) do
     with_account(socket, fn account ->
-      perform_pin!(object_id, gltf_node, account, socket)
-      Storage.promote(file_id, file_access_token, promotion_token, account)
-      OwnedFile.set_active(file_id, account.account_id)
+      hub = socket |> hub_for_socket
+
+      if account |> can?(pin_objects(hub)) do
```
johnshaughnessy: Will we need `can?(unpin_objects(hub))` below in `handle_in("unpin" ...`?

priv/repo/migrations/20190613223618_add_permissions_to_hubs.exs
brianpeiris: nice catch

lib/ret_web/channels/hub_channel.ex
brianpeiris: Nice catch! Thanks.

lib/ret/hub.ex
```diff
@@ -85,6 +87,14 @@ defmodule Ret.Hub do
     |> HubSlug.maybe_generate_slug()
   end
 
+  def add_perms_to_changeset(changeset, attrs) do
```
gfodor: s/perms/member_perms

lib/ret_web/channels/hub_channel.ex
```diff
@@ -145,15 +146,79 @@ defmodule RetWeb.HubChannel do
...
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+    hub = socket |> hub_for_socket
+
+    should_broadcast =
+      cond do
+        naf_template |> String.ends_with?("-media") -> account |> can?(spawn_and_move_media(hub))
+        naf_template |> String.ends_with?("-camera") -> account |> can?(spawn_camera(hub))
+        naf_template |> String.ends_with?("-drawing") -> account |> can?(spawn_drawing(hub))
+        naf_template |> String.ends_with?("-pen") -> account |> can?(spawn_drawing(hub))
+        true -> true
```
gfodor: i think this should yield `false` in the non-match case and we should add suffix checks for all the other known templates (and return `true` for those regardless of permissions.) the argument I made in the previous review was that this suffix based approach + fallthrough deny seemed like the right balance between reticulum needing to be updated in response to client changes and not having to do so if you do minor refactoring.

the way to think about it is each suffix defines a "class" of templates that are, by convention, under a single authorization policy/permission, and so adding a new suffix in the client would then warrant a server change to incorporate the relevant policy for that new class of objects. going forward we should read NAF templates in this way: not as a random identifier, but whose suffix actually has semantic meaning wrt security policy.

priv/repo/migrations/20190613223618_add_permissions_to_hubs.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Ret.Repo.Migrations.AddPermissionsToHubs do
+  use Bitwise
+  use Ecto.Migration
+
+  def change do
+    alter table("hubs") do
+      add(:member_permissions, :integer, default: (1 <<< 4) - 1)
```
gfodor: this one's a little magical :) perhaps a comment and/or refactor. isn't this 0x07? you can probably reference the constants from here, so maybe just write out the expression?

lib/ret_web/channels/hub_channel.ex
gfodor: oops, missed this!

lib/ret_web/channels/hub_channel.ex
```diff
@@ -402,6 +484,19 @@ defmodule RetWeb.HubChannel do
...
 
+  defp should_broadcast(%{"networkId" => network_id, "template" => naf_template}, socket) do
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+    hub = socket |> hub_for_socket
+    is_creator = socket.assigns.created_objects |> Enum.member?(network_id)
+
+    cond do
+      naf_template |> String.ends_with?("-media") -> is_creator or account |> can?(spawn_and_move_media(hub))
+      naf_template |> String.ends_with?("-camera") -> account |> can?(spawn_camera(hub))
+      naf_template |> String.ends_with?("-pen") -> account |> can?(spawn_drawing(hub))
```
gfodor: this means you can't use another users pen or camera if you don't have spawn permissions either right? that seems OK but we'll want to make the client updated.

gfodor:
We should disable moving of pinned objects in pause mode if you do not have object pinning permission

gfodor:
I turned off the client UX checks to see if there were any gaps in server enforcement. Here's what I found if permissions are completely denied:

- Creating a pen still has side effects on other clients, with error spam saying drawing isn't subscribed to

- Taking pictures and screenshots does not spawn the media remotely, but it does show up in chat along with a link to the photo taken. (We probably need to filter the relevant hub channel messages for photos/videos based upon camera permissions)

lib/ret_web/channels/hub_channel.ex
```diff
@@ -145,22 +159,129 @@ defmodule RetWeb.HubChannel do
...
+
+      {:noreply, socket}
+    else
+      {:noreply, socket}
+    end
+  end
+
+  # Captures inbound NAF update messages
+  def handle_in("naf" = event, %{"dataType" => "u"} = payload, socket) do
+    if payload["data"] |> should_broadcast_object_manipulation(:update, socket) do
```
gfodor: `should_broadcast_object_manipulation?`

lib/ret_web/channels/hub_channel.ex
```diff
@@ -402,6 +542,66 @@ defmodule RetWeb.HubChannel do
...
+    end
+  end
+
+  defp should_broadcast_object_manipulation(%{"networkId" => network_id}, type, socket) do
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+    hub = socket |> hub_for_socket
+
+    created_object = socket.assigns.created_objects |> Enum.find(&(&1.network_id == network_id))
+
+    if created_object == nil do
```
gfodor: maybe extract to more overloads (particularly if we introduce a `CreatedObject` struct and call them here to avoid the if/else

```
broadcast_object_manipulation(_socket, _type, nil), do: false
broadcast_object_manipulation(socket, type, %CreatedObject{}) do ...
```

lib/ret_web/channels/hub_channel.ex
```diff
@@ -402,6 +542,66 @@ defmodule RetWeb.HubChannel do
...
+      # It seems we've received an object manipulation message for an object that has not received a first sync yet, 
+      # or was denied creation, so just ignore it.
+      false
+    else
+      is_creator = created_object.creator == socket.assigns.session_id
+      template = created_object.template
+
+      cond do
```
gfodor: similar to above maybe worth extracting this to a `Ret.Hub#allow_object_manipulation?(hub, template, type, account)`

lib/ret_web/channels/hub_channel.ex
```diff
@@ -145,22 +159,129 @@ defmodule RetWeb.HubChannel do
...
+    end
 
-    payload = payload |> Map.put("data", data)
-    broadcast_from!(socket, event, payload)
     {:noreply, socket}
   end
 
+  # Fallthrough for all other dataTypes
   def handle_in("naf" = event, payload, socket) do
     broadcast_from!(socket, event, payload)
```
gfodor: what ends up going here? perhaps we should rip this out (and whitelist if needed) to minimize fallthrough risk if there are only a few places?

lib/ret_web/channels/hub_channel.ex
```diff
@@ -145,22 +159,129 @@ defmodule RetWeb.HubChannel do
...
     {:noreply, socket}
   end
 
-  def handle_in("message" = event, payload, socket) do
-    broadcast!(socket, event, payload |> Map.put(:session_id, socket.assigns.session_id))
+  def handle_in("message" = event, %{"type" => type} = payload, socket) do
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+    hub = socket |> hub_for_socket
+
+    if type != "photo" or account |> can?(spawn_camera(hub)) do
```
gfodor: also should add `type != "video"` (coming w video camera)

lib/ret_web/channels/hub_channel.ex
```diff
@@ -402,6 +542,66 @@ defmodule RetWeb.HubChannel do
...
 
+  defp maybe_store_created_object(socket, payload) do
+    case payload do
+      %{"data" => %{"isFirstSync" => true, "creator" => creator, "template" => template, "networkId" => network_id}} ->
+        created_object = socket.assigns.created_objects |> Enum.find(&(&1.network_id == network_id))
+
+        if created_object == nil do
+          socket
+          |> assign(:created_objects, [
+            %{creator: creator, network_id: network_id, template: template} | socket.assigns.created_objects
```
gfodor: this seems like it might be a small hole (though i can't think of an exploit) -- there's no way to be sure here that the `creator` set here is the NAF client id for this connection, but I believe it should be (mod `:scene` owned objects). i forget if the phoenix session ID is now the NAF client id, but it might be so we could just use that (and drop `creator` and maybe change this to have a `scene_owned: true/false` flag for the remaining use of the `creator` field

priv/repo/migrations/20190613223618_add_permissions_to_hubs.exs
```diff
@@ -0,0 +1,19 @@
...
+    alter table("hubs") do
+      add(:member_permissions, :integer,
+        default:
+          %{
+            spawn_and_move_media: true,
+            spawn_camera: true,
+            spawn_drawing: true,
+            pin_objects: true
+          }
+          |> Ret.Hub.member_permissions_to_int()
```
gfodor: 👍

lib/ret_web/channels/hub_channel.ex
```diff
@@ -392,8 +547,68 @@ defmodule RetWeb.HubChannel do
   end
 
   def handle_out("naf" = event, payload, socket) do
+    sender_account = Account |> Ret.Repo.get(payload["sender_account_id"])
```
gfodor: It looks like this is going to do a database read on every NAF message, which isn't gonna work :(

lib/ret_web/channels/hub_channel.ex
```diff
@@ -392,8 +547,68 @@ defmodule RetWeb.HubChannel do
...
+    hub = socket |> hub_for_socket
```
gfodor: this too

lib/ret_web/channels/hub_channel.ex
```diff
@@ -402,6 +617,121 @@ defmodule RetWeb.HubChannel do
...
+        %{creator: creator, network_id: network_id, template: template} | socket.assigns.secure_scene_objects
+      ])
+    else
+      socket
+    end
+  end
+
+  defp authorize_object_manipulation(%{"networkId" => network_id} = data, socket) do
+    account = Guardian.Phoenix.Socket.current_resource(socket)
+    hub = socket |> hub_for_socket
```
gfodor: this is going to be problematic too, since we are going to be fetching from the db for every NAF message

lib/ret_web/channels/hub_channel.ex
```diff
@@ -392,8 +547,68 @@ defmodule RetWeb.HubChannel do
...
+            end
+
+          {payload, sanitized} =
+            if !authorized and secure_scene_object != nil and
+                 secure_scene_object.template |> String.ends_with?("-media") do
+              payload =
+                payload
+                |> Map.put(
+                  "data",
+                  payload["data"] |> sanitize_with_authorized_schemas(secure_scene_object.template, socket)
```
gfodor: this looks like a security hole -- the authorized schemes here being used as far as i can tell are those registered on the receiver's socket, so by sending different authorized schemas on connect the receiver can grant themselves access

lib/ret/hub.ex
```diff
@@ -300,41 +360,70 @@ defmodule Ret.Hub do
 end
 
 defimpl Canada.Can, for: Ret.Account do
+  alias Ret.{Hub}
```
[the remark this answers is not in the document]
brianpeiris: Hmm, I think I wanted to allow for future actions that are not 1-1 with permissions.

lib/ret/hub.ex
```diff
@@ -300,41 +360,70 @@ defmodule Ret.Hub do
...
+  @object_actions [:spawn_and_move_media, :spawn_camera, :spawn_drawing, :pin_objects]
```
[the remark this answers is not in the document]
brianpeiris: I don't really have a strong preference, but maybe I'm not sure what you had in mind exactly for the DRY'ed version. Since you're looking at this code for multiple owners, feel free to change it.

brianpeiris:
Superseded by #201

PR #200, opened 2019-07-14
Feature/video tweets

gfodor:
Adds the necessary plumbing for tweeting videos, and also adds support for mp4 conversion via Speelycaptor.

### DIFF
--- a/lib/ret/speelycaptor.ex
+++ b/lib/ret/speelycaptor.ex
@@ -1,15 +1,7 @@
 defmodule Ret.Speelycaptor do
   import Ret.HttpUtils
 
-  def convert(%Plug.Upload{path: path, content_type: content_type}, convert_to_content_type) do
-    if content_type |> String.downcase() |> String.starts_with?(convert_to_content_type) do
-      nil
-    else
-      convert(path, convert_to_content_type)
-    end
-  end
-
-  def convert(path, "video/mp4") do
+  def convert(%Plug.Upload{path: path, content_type: "video/" <> _tail}, "video/mp4") do
     with speelycaptor_endpoint when is_binary(speelycaptor_endpoint) <- module_config(:speelycaptor_endpoint) do
       case retry_get_until_success("#{speelycaptor_endpoint}/init") do
         :error ->
--- a/lib/ret/speelycaptor.ex
+++ b/lib/ret/speelycaptor.ex
@@ -4,7 +4,7 @@ defmodule Ret.Speelycaptor do
   def convert(%Plug.Upload{path: path, content_type: "video/" <> _tail}, "video/mp4") do
     with speelycaptor_endpoint when is_binary(speelycaptor_endpoint) <- module_config(:speelycaptor_endpoint) do
       case retry_get_until_success("#{speelycaptor_endpoint}/init") do
-        %HTTPoison.Response{body: body} = resp ->
+        %HTTPoison.Response{body: body} ->
           resp_body = body |> Poison.decode!()
           upload_url = resp_body |> Map.get("uploadUrl")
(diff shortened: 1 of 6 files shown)


lib/ret/speelycaptor.ex
```diff
@@ -0,0 +1,52 @@
...
+              nil
+          end
+
+        :error ->
+          nil
+      end
+    else
+      _ -> nil
+    end
+  end
```
brianpeiris: Would be nice to extract some functions here. The nesting is a bit much.

lib/ret/twitter_client.ex
```diff
@@ -3,6 +3,7 @@ defmodule Ret.TwitterClient do
 
   @twitter_api_base "https://api.twitter.com"
   @twitter_upload_api_base "https://upload.twitter.com"
+  @max_video_upload_size 64 * 1024 * 1024
```
brianpeiris: Is the lambda substrate fast enough to transcode 64MB within the 30 second cap?

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -10,29 +10,48 @@ defmodule RetWeb.Api.V1.MediaController do
...
+    store_and_render_upload(conn, upload, content_type, convert_to_content_type, promotion_token)
   end
 
-  def create(conn, %{"media" => %Plug.Upload{filename: filename, content_type: "application/octet-stream"} = upload}) do
-    store_and_render_upload(conn, upload, MIME.from_path(filename))
+  defp promotion_token_for_params(%{"promotion_mode" => "with_token"}), do: SecureRandom.hex()
+  defp promotion_token_for_params(_params), do: nil
+
+  defp store_and_render_upload(conn, upload, content_type, convert_to_content_type, promotion_token)
+       when is_nil(convert_to_content_type) do
```
brianpeiris: Hmm, why use a guard here instead of just matching `nil = convert_to_content_type`?

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -10,29 +10,48 @@ defmodule RetWeb.Api.V1.MediaController do
...
+        converted_upload = %Plug.Upload{
+          path: converted_path,
+          filename: upload.filename,
+          content_type: convert_to_content_type
+        }
+
+        store_and_render_upload(conn, converted_upload, convert_to_content_type, promotion_token)
+
+      _ ->
+        store_and_render_upload(conn, upload, content_type, promotion_token)
```
brianpeiris: Wouldn't it be better to error here, since we were asked to convert the upload?

lib/ret/twitter_client.ex
gfodor: i tested with 60s videos, i don't think they made it up to 64MB, but it seems highly likely. it's pretty fast (except the cold start)

lib/ret_web/controllers/api/v1/media_controller.ex
gfodor: haha oops

lib/ret_web/controllers/api/v1/media_controller.ex
gfodor: I don't think so, because (despite the probably poorly named API) the contract is "try to convert" -- for example, if you don't have the lambda configured, I want that to just gracefully degrade to not conversion. as a general facility this may not make sense, but for the specific thing we're using it for (the camera videos) it is def what we want.

gfodor: i'll rename the arg to `desired_format`

PR #203, opened 2019-07-23
Do not authorize persistent first syncs

brianpeiris:
Persistent first syncs don't need to be authorized, since they should not result in instantiated entities.

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -1,4 +1,4 @@
-defmodule RetWeb.CreatedEntity do
+defmodule RetWeb.SecuredEntity do
   @enforce_keys [:creator, :template]
   defstruct [:creator, :template]
 end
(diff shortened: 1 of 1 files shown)


lib/ret_web/channels/hub_channel.ex
```diff
@@ -152,7 +152,11 @@ defmodule RetWeb.HubChannel do
   def handle_in("events:end_streaming", _payload, socket), do: socket |> set_presence_flag(:streaming, false)
 
   # Captures all inbound NAF messages that result in spawned objects.
-  def handle_in("naf" = event, %{"data" => %{"isFirstSync" => true, "template" => template}} = payload, socket) do
+  def handle_in(
+        "naf" = event,
+        %{"data" => %{"isFirstSync" => true, "persistent" => false, "template" => template}} = payload,
+        socket
+      ) do
```
gfodor: since we're no longer dealing with persistent entities here, I'd suggest we re-rename the `created_entities` map to `secured_entities` since the persistent entities will no longer be registered

gfodor:
lgtm, one suggestion

PR #209, opened 2019-07-31
Client-side update authorization

brianpeiris:
- Broadcasts user permissions in presence.
- Removes server side authorization and state related to NAF remove messages in favor of client authorization.

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -27,7 +27,7 @@ defmodule RetWeb.HubChannel do
   alias RetWeb.{Presence, SecuredEntity}
   alias RetWeb.Api.V1.{HubView}
 
-  intercept(["mute", "add_owner", "remove_owner", "naf"])
+  intercept(["mute", "add_owner", "remove_owner", "hub_refresh", "naf"])
 
   @hub_preloads [
     scene: [:model_owned_file, :screenshot_owned_file, :scene_owned_file],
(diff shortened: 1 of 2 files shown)


gfodor:
I'm on the fence on this one -- having the server authorize instantiations seems like a nice feature (though obviously warranted a lot of server complexity.) By having the server authorize instantiations, we can have a higher level of certainty that any client-side race conditions or bugs around presence/etc will not end up allowing an attacker to spawn objects. (Though obviously mutating the objects would be possible if we have such a flaw.) I consider it a secondary layer of security -- having more layers can be a good thing.

Any other tradeoffs? I think its probably fine to get rid of this, but want to make sure if there's any major benefit we're leaving on the table we recognize it. Perhaps maybe we could get some of the benefits here by removing the SecuredEntity structure and the socket state around secured entities and push just removal authorization exclusively onto the client, but retain both client and server authorization of instantiations? I think that would result in just a little bit of code in the hub channel (and some additional performance overhead in processing these messages, but they're rare enough it won't matter.)

gfodor:
Oh! I just mis-read the PR. Great minds think alike :)

lib/ret_web/channels/hub_channel.ex
```diff
@@ -507,6 +486,16 @@ defmodule RetWeb.HubChannel do
     {:noreply, socket}
   end
 
+  def handle_out("hub_refresh" = event, %{stale_fields: stale_fields} = payload, socket) do
+    push(socket, event, payload)
+
+    if stale_fields |> Enum.member?("member_permissions") do
```
gfodor: This was a bit confusing at first glance, maybe add comment:
`# If hub member permissions change, everyone should flush their new permissions into presence`

PR #215, opened 2019-08-13
Avatar Remixing

netpro2k:
This contains the necessary backend bits to turn on avatar remixing, namely:

- Allowing duplication of thumbnails when creating an avatar with a `parent_avatar_listing_id`
- Allow searching `avatar_listings` by tag, primarily for base, to be used in
  the updated avatar picker and default to be used to remove client baked-in
  avatars
- Bake account `into avatar_listings` instead of associating through avatars. When
  deleting avatars also delist their associated listings, keeping the account
  association for use down the road.
- Rework GLTF material resolving to have a traversal based fallback for finding
  skinnable material.

TODO:

- [x] fix `with_buffer_override` to include buffer size (unrelated but touching related code)
- [x] ~~decide if we want to keep traversal based material resolving or bake something into the GLTF on upload. If we keep traversal it should probably be cleaned up a bit.~~
   Decided to have the client handle renaming materials if a valid one is not found. The plan is to transition this to some meta information in the gltf eventually, but for now this lets us isolate the special cases to one spot.
- [x] write unit test for avatar deletion and delisting
- [x] base.gltf doesnt work correctly when the parent_avatar_listing has maps
- [x] You can currently not get a gltf for a delisted listing

### DIFF
--- a/lib/ret/gltf_utils.ex
+++ b/lib/ret/gltf_utils.ex
@@ -30,17 +30,16 @@ defmodule Ret.GLTFUtils do
     ])
   end
 
-  @primary_material_name "Bot_PBS"
   def with_default_material_override(gltf, image_files) do
-    material_idx = gltf["materials"] |> Enum.find_index(&(&1["name"] == @primary_material_name))
-    gltf |> with_material_override(material_idx, image_files)
+    gltf |> with_material_override("Bot_PBS", image_files)
   end
 
   def with_material_override(gltf, nil, _image_files) do
     gltf
   end
 
-  def with_material_override(gltf, mat_index, image_files) do
+  def with_material_override(gltf, mat_name, image_files) do
+    mat_index = gltf["materials"] |> Enum.find_index(&(&1["name"] == mat_name))
     material = gltf |> get_in(["materials", Access.at(mat_index)])
 
     image_files
--- a/lib/ret/gltf_utils.ex
+++ b/lib/ret/gltf_utils.ex
@@ -30,40 +30,10 @@ defmodule Ret.GLTFUtils do
     ])
   end
 
-  def reduce(%{"children" => children} = node, {nodes, acc}, fun) do
-    acc = fun.(node, acc)
-    children |> Enum.map(&Enum.at(nodes, &1)) |> Enum.reduce({nodes, acc}, &reduce(&1, &2, fun))
-  end
-
-  def reduce(node, {nodes, acc}, fun) do
-    {nodes, fun.(node, acc)}
-  end
-
-  def reduce(%{"scenes" => scenes, "scene" => scene, "nodes" => nodes} = gltf, acc, fun) do
-    {nodes, acc} =
-      List.first(scenes)["nodes"]
-      |> Enum.map(&Enum.at(nodes, &1))
-      |> Enum.reduce({nodes, acc}, &reduce(&1, &2, fun))
-
-    acc
-  end
-
   @primary_material_name "Bot_PBS"
   def with_default_material_override(gltf, image_files) do
-    material_to_replace =
-      case gltf["materials"] |> Enum.find_index(&(&1["name"] == @primary_material_name)) do
-        nil ->
-          # TODO this currently traverses the whole GLTF, and should early out instead
-          first_material =
-            gltf
-            |> reduce([], &(&2 ++ materials_for_node(gltf, &1)))
-            |> List.first()
-        idx ->
-          idx
-
-      end
-
-    gltf |> with_material_override(material_to_replace, image_files)
+    material_idx = gltf["materials"] |> Enum.find_index(&(&1["name"] == @primary_material_name))
+    gltf |> with_material_override(material_idx, image_files)
   end
 
   def with_material_override(gltf, nil, _image_files) do
--- a/lib/ret/gltf_utils.ex
+++ b/lib/ret/gltf_utils.ex
@@ -24,7 +24,8 @@ defmodule Ret.GLTFUtils do
     gltf
     |> Map.put("buffers", [
       %{
-        "uri" => bin_file |> OwnedFile.uri_for() |> URI.to_string()
+        "uri" => bin_file |> OwnedFile.uri_for() |> URI.to_string(),
+        "byteLength" => bin_file.content_length
       }
     ])
   end
--- a/lib/ret/gltf_utils.ex
+++ b/lib/ret/gltf_utils.ex
@@ -11,7 +11,16 @@ defmodule Ret.GLTFUtils do
     ]
   }
 
-  def with_buffer(gltf, bin_file) do
+  def materials_for_node(gltf, node) do
+    with mesh_idx when not is_nil(mesh_idx) <- node["mesh"],
+         primitives <- get_in(gltf, ["meshes", Access.at(node["mesh"]), "primitives"]) do
+      for prim <- primitives, not is_nil(prim["material"]), do: prim["material"]
+    else
+      _ -> []
+    end
+  end
+
+  def with_buffer_override(gltf, bin_file) do
     gltf
     |> Map.put("buffers", [
       %{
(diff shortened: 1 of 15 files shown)


lib/ret/storage.ex
```diff
@@ -245,4 +255,30 @@ defmodule Ret.Storage do
 
     [path, meta_file_path, blob_file_path]
   end
+
+  def duplicate(%OwnedFile{owned_file_uuid: id, key: key}, %Account{} = account) do
+    {:ok, meta, source_stream} = fetch_blob(id, key, @owned_file_path)
+    %{"content_type" => content_type, "content_length" => content_length, "promotion_token" => promotion_token} = meta
```
gfodor: you can destructure this in one expression if you prefer

lib/ret/gltf_utils.ex
```diff
@@ -20,28 +29,64 @@ defmodule Ret.GLTFUtils do
...
+
+  def with_material_override(gltf, nil, _image_files) do
+    gltf
+  end
+
+  def with_material_override(gltf, mat_index, image_files) do
+    material = gltf |> get_in(["materials", Access.at(mat_index)])
+
+    image_files
+    |> Enum.filter(fn {_, v} -> v end)
```
gfodor: can do `Enum.map(&elem(&1, 1))` if you prefer

lib/ret/gltf_utils.ex
```diff
@@ -20,28 +29,64 @@ defmodule Ret.GLTFUtils do
...
+    |> Enum.flat_map(fn {col, file} ->
+      Enum.map(@texture_paths[col], fn path ->
+        texture_index = material |> get_in(path ++ ["index"])
+        image_index = gltf |> get_in(["textures", Access.at(texture_index), "source"])
+        {image_index, file}
+      end)
+    end)
+    |> Enum.reduce(gltf, fn {index, file}, gltf ->
```
gfodor: is it possible for a texture to not have an image? i am not familiar enough w gltf but thought I'd mention it (eg if the texture is just a pure metal, etc)

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -172,8 +191,20 @@ defmodule RetWeb.Api.V1.AvatarController do
     conn |> send_resp(401, "You do not own this avatar")
   end
 
-  defp delete(conn, %Avatar{} = avatar, %Account{}) do
-    case Repo.delete(avatar) do
+  defp delete(conn, %Avatar{} = avatar, %Account{} = account) do
```
gfodor: I'd suggest moving this into the `avatar_listing.ex` schema and also writing a quick unit test

gfodor: (for the test, important to add additional untouched avatars + listings to ensure no unintentional side effects)

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -22,7 +25,23 @@ defmodule RetWeb.Api.V1.AvatarController do
   end
 
   def create(conn, %{"avatar" => params}) do
-    create_or_update(conn, params, %Avatar{})
+    avatar =
+      with %{"parent_avatar_listing_id" => sid} when not is_nil(sid) and sid != "" <- params,
+           parent_listing <-
```
gfodor: probably worth extracting this somewhere else, particularly if it can get out of the controller.

lib/ret/storage.ex
```diff
@@ -245,4 +255,30 @@ defmodule Ret.Storage do
...
+
+    new_key = SecureRandom.hex()
+    new_promotion_token = SecureRandom.hex()
```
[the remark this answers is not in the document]
netpro2k: Ah actually its the lines after this that are debug code, meaning we end up re-encrypting with the same key/token we had before.

lib/ret/gltf_utils.ex
netpro2k: Unless I am missing something that would be doing something different. The intent here is to filter out tuples of `{column, file}` where `file` is `nil`.

lib/ret/gltf_utils.ex
netpro2k: If its a valid gltf texture should always have a `source` which is an index of an image node.

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Hmm the primary thing being "deleted" here is an avatar, it just happens to also touch listings so I think I will put it in avatar.ex

lib/ret/storage.ex
```diff
@@ -17,28 +19,36 @@ defmodule Ret.Storage do
...
-          {:ok, uuid}
+  # Given a stream, a content-type, an optional encryption key, and a storage subpath, returns an id
+  # that can be used to fetch a stream to the uploaded file after this call.
+  def store_stream(source_stream, source_size, content_type, key, promotion_token, subpath) do
+    with storage_path when is_binary(storage_path) <- module_config(:storage_path) do
+      uuid = Ecto.UUID.generate()
+      [file_directory, meta_file_path, blob_file_path] = paths_for_uuid(uuid, subpath)
+
+      File.mkdir_p!(file_directory)
+      source_stream |> write_stream_to_file(source_size, blob_file_path, key)
```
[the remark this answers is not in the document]
netpro2k: Also renamed the decrypt functions to match

lib/ret/storage.ex
```diff
@@ -245,4 +255,30 @@ defmodule Ret.Storage do
 
     [path, meta_file_path, blob_file_path]
   end
+
+  def duplicate(%OwnedFile{owned_file_uuid: id, key: key}, %Account{} = account) do
```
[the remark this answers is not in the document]
netpro2k: We use `id` everywhere else in the file so I just went with it.

lib/ret/media_search.ex
```diff
@@ -595,6 +595,7 @@ defmodule Ret.MediaSearch do
       name: avatar_listing.name,
       description: avatar_listing.description,
       attributions: avatar_listing.attributions,
+      allow_remixing: not is_nil(avatar_listing.avatar) and avatar_listing.avatar.allow_remixing,
```
gfodor: any reason to not use `!== nil` ?

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -2,45 +2,32 @@ defmodule RetWeb.Api.V1.AvatarController do
...
-  defp preload(nil = _avatar) do
-    nil
+  defp preload(a) do
+    preload(a, [])
   end
 
-  defp preload(%Avatar{} = a) do
-    a |> Repo.preload([Avatar.file_columns() ++ [:parent_avatar, :parent_avatar_listing, :account]])
+  defp preload(%Avatar{} = a, preloads) do
+    a |> Repo.preload([Avatar.file_columns() ++ [:parent_avatar, :parent_avatar_listing, :avatar_listings]])
```
gfodor: did you intentionally drop `account` here?

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -2,45 +2,32 @@ defmodule RetWeb.Api.V1.AvatarController do
...
+  defp get_avatar(avatar_sid, preloads \\ []) do
     avatar_sid
     |> Avatar.avatar_or_avatar_listing_by_sid()
-    |> preload()
+    |> preload(preloads)
   end
 
-  defp preload(nil = _avatar) do
-    nil
+  defp preload(a) do
```
gfodor: do you still need to cover the `nil` case?

lib/ret_web/controllers/api/v1/avatar_controller.ex
```diff
@@ -140,15 +127,26 @@ defmodule RetWeb.Api.V1.AvatarController do
   end
 
   def show_base_gltf(conn, %{"id" => avatar_sid}) do
-    conn |> show_gltf(avatar_sid |> get_avatar(), false)
+    case avatar_sid |> get_avatar([:parent_avatar_listing]) do
+      # For avatars with a parent, base is the same as the parents collapsed avatar plus gltf override
+      %{parent_avatar_listing: parent_listing} = avatar when not is_nil(parent_listing) ->
```
gfodor: `avatar` binding looks unused here

lib/ret_web/views/api/v1/avatar_view.ex
```diff
@@ -23,7 +23,8 @@ defmodule RetWeb.Api.V1.AvatarView do
       # Only include account id on your own avatars
       account_id: account && avatar.account_id == account.account_id && avatar.account_id |> Integer.to_string(),
       allow_remixing: avatar.allow_remixing,
-      allow_promotion: avatar.allow_promotion
+      allow_promotion: avatar.allow_promotion,
+      has_listings: length(avatar.avatar_listings) > 0
```
gfodor: i think this needs to filter out delisted avatars, there's probably a fancy ecto way to create a `visible_avatar_listings` association or something we can preload

priv/repo/migrations/20190819204535_add_avatar_listings_removed_state.exs
```diff
@@ -0,0 +1,8 @@
+defmodule Ret.Repo.Migrations.AddAvatarListingsRemovedState do
+  use Ecto.Migration
+  @disable_ddl_transaction true
+
+  def change do
+    Ecto.Migration.execute("ALTER TYPE avatar_listing_state ADD VALUE IF NOT EXISTS 'removed'")
```
gfodor: should we update the state of the existing records here to go from `delisted` to `removed`?

lib/ret/gltf_utils.ex
gfodor: ah yep

priv/repo/migrations/20190819204535_add_avatar_listings_removed_state.exs
netpro2k: Hmm yeah that probably makes sense

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Yep. We only ever look at the avatar's account_id so no need to load the whole association.

lib/ret_web/controllers/api/v1/avatar_controller.ex
netpro2k: Should be handled by the fallthrough to
```
  defp preload(_avatar, _preloads), do: nil
```

lib/ret_web/views/api/v1/avatar_view.ex
netpro2k: Just did a simple inline filter

PR #230, opened 2019-09-05
Feature/cors proxying

gfodor:
Adds CORS proxying to reticulum, with the same API as cors-anywhere and our cloudflare worker.

This takes advantage of a fork of `ReverseProxyPlug` I created which introduces the ability to add CORS headers and CORS proxy redirecting.

Also fixes the formatter 🤕

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -223,7 +223,8 @@ defmodule RetWeb.PageController do
   defp cors_proxy(%Conn{request_path: "/" <> url, query_string: qs} = conn), do: cors_proxy(conn, "#{url}?#{qs}")
 
   defp cors_proxy(conn, url) do
-    opts = ReverseProxyPlug.init(upstream: url)
+    allowed_origins = Application.get_env(:ret, RetWeb.Endpoint)[:allowed_origins] |> String.split(",")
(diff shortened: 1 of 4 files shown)


gfodor:
(have one issue I just realized, please hold off reviewing until this comment is removed)

gfodor:
Forked code is here: https://github.com/tallarium/reverse_proxy_plug/compare/master...MozillaReality:reticulum/master

lib/ret_web/controllers/page_controller.ex
```diff
@@ -214,6 +219,24 @@ defmodule RetWeb.PageController do
     |> send_resp(200, chunks)
   end
 
+  defp cors_proxy(%Conn{request_path: "/" <> url, query_string: ""} = conn), do: cors_proxy(conn, url)
+  defp cors_proxy(%Conn{request_path: "/" <> url, query_string: qs} = conn), do: cors_proxy(conn, "#{url}?#{qs}")
+
+  defp cors_proxy(conn, url) do
+    allowed_origins = Application.get_env(:ret, RetWeb.Endpoint)[:allowed_origins] |> String.split(",")
+    opts = ReverseProxyPlug.init(upstream: url, allowed_origins: allowed_origins, proxy_url: RetWeb.Endpoint.url())
```
gfodor: note we can't use this plug as a plug because the upstream URL is completely dynamic, not fixed. an alternative approach to this would be to just not use this plug library and lift most of the code into reticulum proper, but most of the code in there can be re-used and so it seemed best to work against a fork of it.

PR #239, opened 2019-10-02
Update the readme to something partially useful

InfiniteLee:


netpro2k:
Nice. I think you are also supposed to run `npm install` from within the assets directory. But unclear if that actually maters since we don't really use any "real" phoenix pages.

PR #243, opened 2019-10-09
Allow scene owned files to be updated

netpro2k:
Previously we were updating the id directly rather than doing it through the association. I "fixed" this in my last PR but did not update the model to allow updates, so re-publishing scenes regressed.


gfodor:
lgtm!

PR #245, opened 2019-10-15
Application configs

brianpeiris:
Adds application config capability, with a model, db-storage and API for self-hosting. This API will be used by the Hubs admin interface.
The config is JSON encoded and injected into the page as a script along with the page's meta tags.
This PR also includes changes to proxy the admin console server and proxy a new app-config-schema.toml file from the hubs server.


gfodor:
so one thing I caught today since I was working on making the build time configs dynamic is that injecting a script tag will violate the current CSP. it seems ideal that we not allow dynamic script injection if we can avoid it. so, I'd suggest we pack this data into meta tags or `data-` attributes somewhere.

lib/ret/app_config.ex
```diff
@@ -0,0 +1,30 @@
...
+    # We wrap the config value in an outer %{value: ...} map because we want to be able to accept primitive
+    # value types, but store them as json.
+    attrs = attrs |> Map.put("value", %{value: attrs["value"] |> Poison.decode!()})
+
+    app_config
+    |> cast(attrs, [:key, :value])
+    |> unique_constraint(:key)
+  end
+
+  def get_config() do
```
gfodor: might be premature optimization, but putting this behind a cachex key seems wise

lib/ret_web/controllers/api/v1/app_config_controller.ex
```diff
@@ -0,0 +1,16 @@
+defmodule RetWeb.Api.V1.AppConfigController do
+  use RetWeb, :controller
+  alias Ret.{Repo, AppConfig}
+
+  def create(conn, %{"app_config" => app_config_params}) do
```
gfodor: i'm not familiar with the APIs yet, but I'm guessing we should make this basically the same API as `ita` (or vice versa)

gfodor: (and/or we may want to have this API proxy to ita for configs that are managed there)

lib/ret_web/controllers/page_controller.ex
```diff
@@ -139,17 +139,40 @@ defmodule RetWeb.PageController do
...
+    index_meta_tags =
+      Phoenix.View.render_to_string(RetWeb.PageView, "index-meta.html",
+        app_config_script: {:safe, "<script>#{app_config_script}</script>"}
+      )
 
     chunks =
       chunks_for_page("index.html", :hubs)
       |> List.insert_at(1, index_meta_tags)
 
+    app_config_csp = "'sha256-#{:crypto.hash(:sha256, app_config_script) |> :base64.encode()}'"
```
gfodor: nice one!

gfodor: also be sure to test in both ff and chrome, i believe there can be weirdness w the expected base64 RFC

lib/ret_web/controllers/page_controller.ex
```diff
@@ -139,17 +139,40 @@ defmodule RetWeb.PageController do
...
 
+  defp append_csp(conn, directive, source) do
+    csp_header = conn.resp_headers |> Enum.find(fn {key, _value} -> key == "content-security-policy" end)
+
+    new_directive = "#{directive} #{source}"
+
+    csp_value =
+      case csp_header do
+        nil -> new_directive
+        {_csp, csp_value} -> csp_value |> String.replace(directive, new_directive)
```
gfodor: for completeness i think this needs to handle the case where the CSP has no `script-src` directive already

lib/ret_web/router.ex
```diff
@@ -69,6 +69,12 @@ defmodule RetWeb.Router do
     forward("/", RetWeb.Plugs.PostgrestProxy)
   end
 
+  scope "/api/ita/configs" do
+    pipe_through([:secure_headers, :admin_required])
+    post("/app-config", RetWeb.Api.V1.AppConfigController, :create)
```
gfodor: sorry I poorly phrased my comment wrt ita, what I was getting at is the json protocol for the app configs should be the same as ita's. for example, if we had a hubs config called foo.bar it should be POSTed in the form:
```
{
hubs: {
   foo: {
      bar: "value"
   }
}
```

i didn't mean we should embed the app config REST under the `/api/ita` path -- i think `/api/ita` should strictly be a proxy to the underlying `ita` service. (whose `POST` API for configs is the same as our app level one as above)

lib/ret_web/controllers/page_controller.ex
```diff
@@ -96,6 +96,9 @@ defmodule RetWeb.PageController do
 
   def render_for_path("/hub.service.js", _params, conn), do: conn |> render_page("hub.service.js")
 
+  def render_for_path("/app-config-schema.toml", _params, conn),
```
gfodor: i think we should expose these as `/hubs/schema.toml` and `/spoke/schema.toml`, and they can be served up simliar to the web manifest as a whitelisted file.

the fields can have categories similar to the current ita ones and be presented that way in the admin console

gfodor: i do agree conceptually that we serve up the full "app config" which includes spoke and hubs as you are doing now though

lib/ret_web/controllers/page_controller.ex
```diff
@@ -139,17 +142,55 @@ defmodule RetWeb.PageController do
...
+    app_config =
+      if module_config(:skip_cache) do
+        Ret.AppConfig.get_config()
+      else
+        {:ok, app_config} = Cachex.get(:app_config, :app_config)
+        app_config
+      end
+
+    app_config_json = app_config |> Poison.encode!()
+    app_config_script = "window.APP_CONFIG = JSON.parse('#{app_config_json |> String.replace("'", "\\'")}')"
```
gfodor: `window.APP = {}; window.APP.config = ... ` maybe?

lib/ret_web/controllers/page_controller.ex
```diff
@@ -227,10 +268,17 @@ defmodule RetWeb.PageController do
     "text/html; charset=utf-8"
   end
 
-  defp render_chunks(conn, chunks, content_type) do
+  defp render_chunks(conn, chunks, content_type, include_newlines) do
+    resp =
+      if include_newlines do
```
gfodor: any reason we would ever want to not do this? seems more correct since iirc the chunks were split on newline (i'm surprised we didn't have bugs unless i'm misunderstanding this code)

lib/ret/app_config.ex
```diff
@@ -0,0 +1,64 @@
...
+    AppConfig
+    |> Repo.all()
+    |> Map.new(fn app_config -> expand_key(app_config.key, app_config.value["value"], true) end)
+  end
+
+  defp expand_key(key, val, first) do
+    if key |> String.contains?("|") do
+      [head, tail] = key |> String.split("|", parts: 2)
```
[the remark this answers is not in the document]
brianpeiris: Can't use `.` or `_`, unfortunately, since I want to be able to handle keys like `"translations|en|home.hero_blurb"`. In toml, this is represented as `translations.en."home.hero_blurb"`, so it wouldn't be possible to use `.` without more intelligent parsing.

These transforms I'm doing are kinda unfortunate. I was wondering if I should just store the config as a single json blob record in the DB (or on disk?) instead. I don't have a strong opinion about it though.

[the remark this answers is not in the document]
gfodor: ah understood -- uniblob isn't a bad idea, but i think the current schema is better than a blob because it lets us fit in the stored file stuff nicely wrt referential integrity.

lib/ret_web/controllers/page_controller.ex
gfodor: ignore my second comment, didn't think it through (and we talked about what i think is ideal)

lib/ret/app_config.ex
```diff
@@ -0,0 +1,64 @@
...
+
+      if first do
+        {head, expand_key(tail, val, false)}
+      else
+        %{head => expand_key(tail, val, false)}
+      end
+    else
+      if first do
+        {key, val}
```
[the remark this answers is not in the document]
brianpeiris: I was passing the output of this directly into `Map.new`, so the first level needed to be tuples. But I realized the implementation was broken in the first place since it didn't handle deep merging correctly, so that's fixed now and it's cleaner.

lib/ret_web/controllers/page_controller.ex
brianpeiris: This gets put into your `configs` object anyway, so I don't think the intermediate name matters so much.

lib/ret_web/controllers/page_controller.ex
brianpeiris: Yeah, maybe I can just add the newlines always. Will test it out to see if anything breaks.

PR #249, opened 2019-10-16
"Similar to" search for avatar listings

netpro2k:
The current "algorithm" is super simple and just returns avatars with the same `parent_avatar_listing_id` as the avatar sid specified. An immediate improvement would be to include anything with a shared ancestor, but the query for that is a bit trickier to write, so starting with this. Over time this could actually use some sort of more interesting metric of similarity.

### DIFF
--- a/lib/ret/media_search.ex
+++ b/lib/ret/media_search.ex
@@ -1,6 +1,6 @@
 defmodule Ret.MediaSearchQuery do
   @enforce_keys [:source]
-  defstruct [:source, :type, :user, :filter, :q, :cursor, :locale]
+  defstruct [:source, :type, :user, :filter, :q, :similar_to, :cursor, :locale]
 end
 
 defmodule Ret.MediaSearchResult do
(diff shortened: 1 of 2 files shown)


lib/ret/media_search.ex
```diff
@@ -507,6 +514,21 @@ defmodule Ret.MediaSearch do
   defp add_tag_to_listing_search_query(query, nil), do: query
   defp add_tag_to_listing_search_query(query, tag), do: query |> where(fragment("tags->'tags' \\? ?", ^tag))
 
+  defp add_similar_to_to_listing_search_query(query, nil), do: query
+
+  defp add_similar_to_to_listing_search_query(query, similar_sid) do
+    case Avatar.avatar_or_avatar_listing_by_sid(similar_sid) do
```
netpro2k: There is probably some way to do this without the additional query, but not sure how. That said this probably needs to eventually walk the hierarchy to find listings with any common ancestors, and I think that would need an additional query anyway.

lib/ret_web/controllers/api/v1/media_search_controller.ex
```diff
@@ -29,25 +29,16 @@ defmodule RetWeb.Api.V1.MediaSearchController do
     conn |> render("index.json", results: results)
   end
 
-  def index(conn, %{"source" => "avatar_listings", "filter" => filter} = params) do
```
netpro2k: If there was a reason we had these all being handled individually let me know, but it seems better handled at the `MediaSearchQuery` level

gfodor: no, probably just a spot we missed when refactoring

PR #253, opened 2019-10-29
More app config

brianpeiris:
Additions to app config functionality
- Allow app config to accept file upload values, which are then promoted to owned files.
- Inject app config into the rest of the entry page metas.
- Rename app-config-schema.toml to schema.toml

Misc. improvements
- Have meta.ex use phx_host from config/dev.exs if available
- Rename start.sh to run.sh to better match run-local-reticulum.sh in the hubs repo


lib/ret_web/controllers/api/v1/app_config_controller.ex
```diff
@@ -1,18 +1,31 @@
...
-      (AppConfig |> Repo.get_by(key: key) || %AppConfig{})
-      |> AppConfig.changeset(%{"key" => key, "value" => val})
-      |> Repo.insert_or_update!()
+      app_config = AppConfig |> Repo.get_by(key: key) || %AppConfig{}
+
+      app_config =
+        case val do
+          %{"file_id" => file_id, "meta" => %{"access_token" => access_token, "promotion_token" => promotion_token}} ->
+            {:ok, owned_file} = Storage.promote(file_id, access_token, promotion_token, account)
+            file_uri = owned_file |> OwnedFile.uri_for() |> URI.to_string()
```
gfodor: i think we should defer URL generation to the file at app config GET time, not here, since that URL structure may change. the 'value' of the config is the file itself semantically.

lib/ret_web/controllers/page_controller.ex
```diff
@@ -205,23 +200,43 @@ defmodule RetWeb.PageController do
...
       |> List.insert_at(1, hub_meta_tags)
 
     conn
+    |> append_csp("script-src", app_config_csp)
     |> put_resp_header("content-type", "text/html; charset=utf-8")
     |> send_resp(200, chunks)
   end
 
+  defp generate_app_config() do
+    app_config =
```
gfodor: I'd move this to `Ret.AppConfig`

brianpeiris: Huh, I thought exactly the opposite, since the method deals with caching (skip_cache particularly), CSP and HTML. Why do you prefer it to be in AppConfig?

gfodor: ah, I guess i just have a general aversion to any 'controller' method that theoretically could be stuffed somewhere else. i think you're right though that the caching bits don't belong there. it was the script/csp generation that caught my eye. i don't mind if you keep it here.

PR #255, opened 2019-11-07
Add search for sketchfab collections and tweak featured search params.

robertlong:


lib/ret/media_search.ex
```diff
@@ -62,38 +63,89 @@ defmodule Ret.MediaSearch do
     favorites_search(cursor, type, account_id, q)
   end
 
+  def search(%Ret.MediaSearchQuery{source: "sketchfab", cursor: cursor, filter: nil, collection: nil, q: q}) do
+    if q == nil || q == "" do
```
gfodor: this should be a separate match

lib/ret/media_search.ex
```diff
@@ -62,38 +63,89 @@ defmodule Ret.MediaSearch do
...
     query =
       URI.encode_query(
         type: :models,
         downloadable: true,
         count: @page_size,
         max_face_count: @max_face_count,
+        max_filesizes: "gltf:#{@max_file_size_bytes}",
         processing_status: :succeeded,
+        sort_by:
+          if q == nil || q == "" do
```
gfodor: if its a collection, should we care about the query here?

robertlong: Yeah, you can search within a collection, which can be useful for larger collections.

lib/ret/media_search.ex
robertlong: How can I match on nil and ""?

gfodor: you can use `when` https://hexdocs.pm/elixir/master/guards.html

lib/ret/media_search.ex
gfodor: sorry, what i meant is if the `sort_by` should be `nil` if we have a query. i guess that seems right assuming sketchfab will sort by relevance.

robertlong: Yeah, I'm pretty sure it does.

PR #256, opened 2019-11-07
Configurable static assets

brianpeiris:
- favicon.ico and hub-preview.png (renamed to app-thumbnail.png) are now pulled from the client and are configurable via the app config
- The PWA manifest.webmanifest is no longer pulled from the client, but is generated by reticulum with values from the app config


lib/ret_web/controllers/page_controller.ex
```diff
@@ -119,12 +121,77 @@ defmodule RetWeb.PageController do
...
+  def render_for_path("/app-thumbnail.png", _params, conn) do
+    thumbnail = get_configurable_asset(:app_config_app_thumbnail, "images|app_thumbnail", "app-thumbnail.png")
+
+    conn
+    |> put_resp_header("content-type", "image/png")
+    |> send_resp(200, thumbnail)
+  end
+
+  defp get_configurable_asset(cache_key, config_key, fallback_file) do
+    case Cachex.get(:assets, cache_key) do
```
gfodor: any reason not to use `Cache.fetch` and use the fallback approach /w skip_cache  (similar to HEAD with app_config's caching)

lib/ret_web/templates/page/manifest.webmanifest.eex
```diff
@@ -0,0 +1,40 @@
...
+      "src": "<%= @app_icon %>",
+      "sizes": "512x512",
+      "type": "image/png"
+    }
+  ],
+  "scope": "/",
+  "start_url": "/",
+  "display": "standalone",
+  "orientation": "portrait",
+  "theme_color": "#000",
```
gfodor: we can eventually get smarter here, but i don't think we should worry about dark vs light just yet (i expect us to have more themes)

lib/ret_web/controllers/page_controller.ex
brianpeiris: I wanted the cache to contain either the appconfig asset or the fallback asset from the page warmer.

PR #259, opened 2019-11-09
Self hosting defaults

brianpeiris:
- Inject meta information about available integrations into script tag for twitter and search UI
- Default to restrictive room permissions settings with `permissive_rooms` config


lib/ret/hub.ex
```diff
@@ -50,6 +50,13 @@ defmodule Ret.Hub do
     pin_objects: true
   }
 
+  @default_restrictive_member_permissions %{
```
gfodor: ah this is an interesting approach! i was thinking we would expose all the permissions individually as on/off switches, but i think the approach you've taken here may be smarter, because it means as we add more permissions we can 'do the right thing' based upon the users meta-preference of if rooms should be 'open' or 'closed'. i think given our current permissions, it seems unlikely that a mixed default (eg media but no pen) probably isn't worth supporting in lieu of this.

lib/ret_web/controllers/page_controller.ex
```diff
@@ -200,13 +200,18 @@ defmodule RetWeb.PageController do
 
   def render_hub_content(conn, hub, _slug) do
     hub = hub |> Repo.preload(scene: [:screenshot_owned_file], scene_listing: [:screenshot_owned_file])
+
     {app_config_script, app_config_csp} = generate_app_config()
 
+    {available_integrations_script, available_integrations_csp} =
```
gfodor: Ah, I was assuming we'd put this into the `/api/v1/meta` API, is there a reason injecting it at page load time seems better or worse? i guess since we already have the pattern for configs, these are kind of equivalent in terms of impact, so i'm learning towards your approach.

brianpeiris: Just felt right to guarantee this information was available before the rest of the javascript run, without having to wait for another request. I don't see a strong reason to put in in the API for this purpose, but we can always also put it there if we need it later.

lib/ret/hub.ex
brianpeiris: Yeah I prefer this approach, especially since there is logic involved with the current permission settings -- namely that you cannot toggle `spawn_camera` and `pin_objects` independently of `spawn_and_move_media`.

PR #260, opened 2019-11-12
Scene Remixing and Project publish endpoint

netpro2k:
This adds the ability to clone a scene or scene listing into "my scenes". A project is also created for the scene, though we may want to split that off into its own step.

To make creating publishing projects simpler, a new `/projects/:id/publish` endpoint has been added that will automatically create a scene if one does not exist and associate it with the project, then update its files, all in a single transaction.

Changes were also made to the responses for Scenes, SceneListings, and Projects ass well as the associated media search responses to facilitate the new client features this was built for.

Also small side fix for importing attributions from remote scenes/avatars.

### DIFF
--- a/lib/ret/project.ex
+++ b/lib/ret/project.ex
@@ -14,7 +14,13 @@ defmodule Ret.Project do
     belongs_to(:created_by_account, Ret.Account, references: :account_id)
     belongs_to(:project_owned_file, Ret.OwnedFile, references: :owned_file_id)
     belongs_to(:thumbnail_owned_file, Ret.OwnedFile, references: :owned_file_id)
-    many_to_many(:assets, Ret.Asset, join_through: Ret.ProjectAsset, join_keys: [project_id: :project_id, asset_id: :asset_id], on_replace: :delete)
+
+    many_to_many(:assets, Ret.Asset,
+      join_through: Ret.ProjectAsset,
+      join_keys: [project_id: :project_id, asset_id: :asset_id],
+      on_replace: :delete
+    )
+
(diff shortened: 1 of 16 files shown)


lib/ret/project.ex
```diff
@@ -14,36 +14,57 @@ defmodule Ret.Project do
...
   def project_by_sid_for_account(project_sid, account) do
     from(p in Project,
       where: p.project_sid == ^project_sid and p.created_by_account_id == ^account.account_id,
-      preload: [:created_by_account, :project_owned_file, :thumbnail_owned_file, assets: [:asset_owned_file, :thumbnail_owned_file]])
-    |> Repo.one
+      preload: [
+        :created_by_account,
+        :project_owned_file,
+        :thumbnail_owned_file,
+        :scene,
```
robertlong: This should be:
```
 scene: [:account, :model_owned_file, :screenshot_owned_file, :scene_owned_file],
```

lib/ret/project.ex
```diff
@@ -14,36 +14,57 @@ defmodule Ret.Project do
...
   end
 
   def projects_for_account(account) do
-    Repo.all from p in Project,
-      where: p.created_by_account_id == ^account.account_id,
-      preload: [:project_owned_file, :thumbnail_owned_file]
+    Repo.all(
+      from(p in Project,
+        where: p.created_by_account_id == ^account.account_id,
+        preload: [:project_owned_file, :thumbnail_owned_file, :scene]
```
robertlong: This should be:
```
scene: [:account, :model_owned_file, :screenshot_owned_file, :scene_owned_file]
```

PR #262, opened 2019-11-14
Increase maximum sketchfab collection file size / polycount

robertlong:
There are a few models in the Sketchfab collections I'd like to feature that are over the filesize / polycount limit. Because we check everything that goes into these collections, I think it's fine to increase these limits to a reasonably large upper bound.

### DIFF
--- a/lib/ret/media_search.ex
+++ b/lib/ret/media_search.ex
@@ -27,9 +27,9 @@ defmodule Ret.MediaSearch do
   # HACK for now to reduce page size for scene listings -- real fix will be to expose page_size to API
   @scene_page_size 23
   @max_face_count 60000
-  @max_collection_face_count 200000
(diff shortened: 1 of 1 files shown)


lib/ret/media_search.ex
```diff
@@ -27,7 +27,9 @@ defmodule Ret.MediaSearch do
   # HACK for now to reduce page size for scene listings -- real fix will be to expose page_size to API
   @scene_page_size 23
   @max_face_count 60000
+  @max_collection_face_count 200000
   @max_file_size_bytes 20 * 1_048_576
+  @max_collection_file_size_bytes 100 * 1_048_576
```
gfodor: nit but always can do `1024 * 1024` :)

robertlong: Yeah, I'll do that 👍

PR #265, opened 2019-11-19
Added disk quota

gfodor:
Adds the `quota_gb` config which will block subsequent attempts to store files if the disk the storage path is on has more than `quota_gb` GB in-use.

### DIFF
--- a/lib/ret/storage.ex
+++ b/lib/ret/storage.ex
@@ -21,13 +21,17 @@ defmodule Ret.Storage do
   # Given a path to a file, a content-type, and an optional encryption key, returns an id
   # that can be used to fetch a stream to the uploaded file after this call.
   def store(path, content_type, key, promotion_token) do
-    case File.stat(path) do
-      {:ok, %{size: source_size}} ->
-        source_stream = path |> File.stream!([], @chunk_size)
-        store_stream(source_stream, source_size, content_type, key, promotion_token, @expiring_file_path)
+    if in_quota?() do
+      case File.stat(path) do
+        {:ok, %{size: source_size}} ->
+          source_stream = path |> File.stream!([], @chunk_size)
+          store_stream(source_stream, source_size, content_type, key, promotion_token, @expiring_file_path)
 
-      {:error, _reason} = err ->
-        err
+        {:error, _reason} = err ->
+          err
+      end
+    else
+      { :error, :quota }
     end
   end
 
@@ -281,10 +285,6 @@ defmodule Ret.Storage do
     Ret.Crypto.encrypt_stream_to_file(source_stream, source_size, destination_path, key |> Ret.Crypto.hash())
   end
 
-  defp module_config(key) do
-    Application.get_env(:ret, __MODULE__)[key]
-  end
-
   defp paths_for_uuid(uuid, subpath) do
     path = "#{module_config(:storage_path)}/#{subpath}/#{String.slice(uuid, 0, 2)}/#{String.slice(uuid, 2, 2)}"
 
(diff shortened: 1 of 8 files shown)


lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -62,6 +62,9 @@ defmodule RetWeb.Api.V1.MediaController do
           meta: %{access_token: access_token, promotion_token: promotion_token, expected_content_type: content_type}
         )
 
+      {:error, :quota} ->
+        conn |> send_resp(400, "Unable to store additional content.")
```
brianpeiris: Maybe 413 (payload too large) would be a better response than 400 (bad request)?

lib/ret/storage_used.ex
```diff
@@ -0,0 +1,17 @@
...
+  use Retry
+
+  def interval, do: :timer.minutes(5)
+
+  def execute(_state) do
+    storage_path = Application.get_env(:ret, Ret.Storage)[:storage_path]
+    
+    case System.cmd("df", ["--output=used", storage_path]) do
+      { lines, 0 } ->
+        used = lines |> String.split("\n") |> Enum.at(1) |> Integer.parse |> elem(0)
```
brianpeiris: Is it safe to assume only one mounted filesystem and that the one we care about will always be the first one?

lib/ret/storage_used.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Ret.StorageUsed do
+  use Cachex.Warmer
+  use Retry
+
+  def interval, do: :timer.minutes(5)
+
+  def execute(_state) do
+    storage_path = Application.get_env(:ret, Ret.Storage)[:storage_path]
+    
+    case System.cmd("df", ["--output=used", storage_path]) do
```
brianpeiris: The output flag is a GNU extension, so this is not POSIX compliant, but I guess that doesn't matter if we assume we control the infrastructure.

lib/ret/storage.ex
```diff
@@ -350,4 +350,19 @@ defmodule Ret.Storage do
       owned_file
     end)
   end
+
+  def in_quota?() do
+    with storage_path when is_binary(storage_path) <- module_config(:storage_path),
+         quota_gb when is_integer(quota_gb) and quota_gb > 0 <- module_config(:quota_gb) do
```
brianpeiris: Is this going to be configurable via ita? I wonder if we should make this configurable with the app config.

lib/ret/storage_used.ex
gfodor: passing the path to `ld` here should return the mount the path exists on. the main weirdness is we are going to cap storage to that entire volume -- so if its a volume shared with other data, it'll quota out early.

lib/ret/storage_used.ex
gfodor: interesting -- this mode is pretty nice since it avoids most of the weird parsing needed, but i'll see if i can do something more compliant. (i'm assuming the habitat coreutils has the GNU variant, which limits some of the risk here insofar as we expect people to use habitat)

lib/ret/storage.ex
gfodor: yup, it's part of the stack configs

gfodor: https://github.com/mozilla/hubs-ops/blob/master/cloudformation/stack.yaml#L139

lib/ret/storage_used.ex
gfodor: updated to use `ld -k -T`

PR #266, opened 2019-11-20
Configurable meta tags, etc.

brianpeiris:
- Configurable meta tags and titles in entry pages
- Injected app name into hub.service.js
- Parameterized email content
- Removed unused spoke image assets
- Removed static fallback for app icons and thumbnail. They must be configured in the DB now.
- Removed old spoke landing page

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -107,10 +107,10 @@ defmodule RetWeb.PageController do
   def render_for_path("/whats-new/", _params, conn),
     do: conn |> render_page("whats-new.html", :hubs, "whats-new-meta.html")
 
-  def render_for_path("/hub.service.js", _params, conn), do: conn |> render_asset("hub.service.js")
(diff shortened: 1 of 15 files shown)


lib/ret_web/controllers/page_controller.ex
```diff
@@ -198,38 +194,49 @@ defmodule RetWeb.PageController do
     end
   end
 
-  defp get_configurable_asset(cache_key, config_key, fallback_file) do
-    case Cachex.get(:assets, cache_key) do
-      {:ok, nil} ->
-        app_config = AppConfig |> Repo.get_by(key: config_key) |> Repo.preload(:owned_file)
+  defp respond_with_configurable_asset(conn, cache_key, config_key, content_type) do
```
gfodor: as a sanity check, did you verify we do no database queries when the cache is warm?

brianpeiris: yup, no queries when the cache is warm.

PR #267, opened 2019-11-21
Add hubs header to index HEAD response so that hubs servers can ident…

brianpeiris:
…ify eachother

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -230,15 +230,15 @@ defmodule RetWeb.PageController do
     end
   end
 
+  defp render_index(conn) do
+    method = conn |> get_req_header("x-original-method") |> Enum.at(0)
+    conn |> render_index(method)
+  end
+
   defp render_index(conn, "HEAD") do
     conn
     |> put_hub_headers("hub")
     |> send_resp(200, "")
-  end
-
-  defp render_index(conn) do
-    method = conn |> get_req_header("x-original-method") |> Enum.at(0)
-    conn |> render_index(method)
   end
 
   defp render_index(conn, _method) do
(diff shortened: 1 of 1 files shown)


lib/ret_web/controllers/page_controller.ex
```diff
@@ -229,25 +229,32 @@ defmodule RetWeb.PageController do
   end
 
   def render_index(conn) do
-    {app_config, app_config_script, app_config_csp} = generate_app_config()
+    if conn |> get_req_header("x-original-method") |> Enum.at(0) === "HEAD" do
```
gfodor: haha, this hack lives. i'd probably refactor this into two functions matched on the header value.

lib/ret_web/controllers/page_controller.ex
```diff
@@ -229,25 +229,32 @@ defmodule RetWeb.PageController do
...
+      conn
+      |> put_resp_header("hubs-server", "true")
```
gfodor: given that we're calling these servers "hubs", maybe this is redundant and should be "Is-Hub" or something

lib/ret_web/controllers/page_controller.ex
brianpeiris: heh, some day I'll learn to resist `if` statement muscle memory

lib/ret_web/controllers/page_controller.ex
gfodor: maybe `Hubs-Origin`?

gfodor: what about a header akin to `Content-Type`, but identifies URLs as the root homepage (the 'hub'), avatars, scenes, or rooms?

`Entity-Type`?

brianpeiris: Yeah, "is-hub" seems okay. I was considering "is-moz-hub" or "is-hub-compatible" to help reduce the chance of future collision with the wider web, but I'm not sure. I guess we probably want to avoid any branding though.

PR #268, opened 2019-11-22
Add dynamic robots.txt

gfodor:
Adds the ability to disallow crawlers, which will serve up a robots.txt denying all crawling other than twitter or facebook image unfurling bots.

Note this will be off by default in `ita` for self hosted instances.

### DIFF
--- a/lib/ret_web/templates/page/robots.txt.eex
+++ b/lib/ret_web/templates/page/robots.txt.eex
@@ -1,0 +1,13 @@
+<%=if @allow_crawlers do %>
+User-agent: *
+Allow: /
+<% else %>
+User-agent: *
+Disallow: /
+
+User-agent: Twitterbot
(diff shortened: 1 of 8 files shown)


lib/ret_web/templates/page/robots.txt.eex
```diff
@@ -0,0 +1,13 @@
...
+Allow: /
+<% else %>
+User-agent: *
+Disallow: /
+
+User-agent: Twitterbot
+Allow: /
+
+User-agent: facebookexternalhit
+Allow: /
```
brianpeiris: Do we need to allow slack and discord crawlers as well?

gfodor: afaict, no, slack explicitly says they ignore it, couldn't find anything on discord.

PR #269, opened 2019-11-25
Fix scene preloads

netpro2k:
Fix scene preload issues introduced from scene remixing change, and centralize them so that we are less likely to break them in the future.


robertlong:
I think we also need to preload here: https://github.com/mozilla/reticulum/blob/master/lib/ret/media_search.ex#L478

PR #283, opened 2019-12-11
Feature/dynamic csp

gfodor:
Updates reticulum to compute the CSP. Moves the previous config into a new key (`csp`) so all existing stacks will by default begin generating the CSP.


gfodor:
Note that our prod CSP is weird enough I believe we should continue manually maintaining it. It does mean a CSP change now warrants changes in both the vault configs and also this file. (And thankfully this will let me remove the CSPs written in arbortect and the cloudformation template)

habitat/config/config.toml
```diff
@@ -42,6 +42,14 @@ host = "{{ cfg.phx.url_host_prefix }}{{ cfg.phx.cors_proxy_url_host }}"
...
 {{/if}}
 
+{{#if cfg.phx.thumbnail_url_host }}
+[ret."Elixir.RetWeb.Endpoint".thumbnail_url]
+host = "{{ cfg.phx.url_host_prefix }}{{ cfg.phx.thumbnail_url_host }}"
+{{#if cfg.phx.thumbnail_url_port }}
+port = "{{ cfg.phx.thumbnail_url_port }}"
+{{/if}}
+{{/if}}
+
```
brianpeiris: Is this change intentionally in the PR?

brianpeiris: Ah, nvm, I see it's used below

PR #296, opened 2020-01-04
Pass account along to scene rendering

gfodor:
This fixes a bug where when we render the `scene` for a hub when connecting to a room, we fail to pass the `account` for the user along to the renderer, which results in a `null` `account_id` in the response regardless of if the user who is logged in created the scene.

Without this PR, the scene link in the room and scene info dialog will be disabled, even if logged in as the scene creator.


emstanley:
Not a high priority. May close as "won't fix."

PR #303, opened 2020-01-15
Improve Quality of 360 Youtube Videos

gfodor:
Updates the media resolver to properly deal with youtube 360 videos. Particularly:

- Changes the boolean `low_resolution` flag to now be a `quality` measure, which can be `low`, `high`, `low_360`, or `high_360`
- Updates the API so the client now can specify quality, instead of deriving it from the user agent
- `low_360` and `high_360` now perform two `youtube-dl` requests in order to extract the video and audio streams separately. this is done because it was determined that youtube serves these up separately for 360 videos, there is no way to get an all-in-one stream.
- if there's a separate audio stream, it is returned in `origin_audio`
- deals with a change in youtube where you can't get equirectangular 360 videos unless you pass an empty user-agent string (youtube dl server has been patched to support this) -- see https://github.com/ytdl-org/youtube-dl/issues/15267#issuecomment-370122336

### DIFF
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -1,11 +1,11 @@
 defmodule Ret.ResolvedMedia do
   @enforce_keys [:uri]
-  defstruct [:uri, :meta]
+  defstruct [:uri, :audio_uri, :meta]
 end
 
 defmodule Ret.MediaResolverQuery do
   @enforce_keys [:url]
-  defstruct [:url, supports_webm: true, low_resolution: false, version: 1]
+  defstruct [:url, supports_webm: true, quality: :high, version: 1]
 end
 
 defmodule Ret.MediaResolver do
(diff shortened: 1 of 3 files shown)


lib/ret/media_resolver.ex
```diff
@@ -426,46 +461,61 @@ defmodule Ret.MediaResolver do
...
 
   defp module_config(key) do
     Application.get_env(:ret, __MODULE__)[key]
   end
 
-  defp ytdl_resolution(%MediaResolverQuery{low_resolution: true}), do: "480"
-  defp ytdl_resolution(_query), do: "720"
+  defp ytdl_resolution(%MediaResolverQuery{quality: :low}), do: "[height<=480]"
+  defp ytdl_resolution(%MediaResolverQuery{quality: :low_360}), do: "[height<=1440]"
+  defp ytdl_resolution(%MediaResolverQuery{quality: :high_360}), do: "[height<=2160]"
```
netpro2k: Hmm, would this just return no results if you requested a `high_360` but only 1440 was available? Bit confusing how this interacts with `best` and `bestvideo`

gfodor:
Ah yep, so the reason I decided to only scope the dual-stream approach to 360 videos for now was because it does result in two round trips to ytdl (and crawls) which is materially slower. Eventually it would probably be smart to change the way we interface w ytdl to grab the full `info` blob and make a good choice then.

lib/ret/media_resolver.ex
gfodor: It's `<=`, so it will prefer the highest resolution video that is at most 2160 I assume.

PR #304, opened 2020-01-17
Public Rooms

netpro2k:
Adds ability to make rooms public, which means they will show up on the homepage. By default only admins can do this, but it can be enabled for all room owners through a config flag. 

Hubs changes: https://github.com/mozilla/hubs/pull/2059

todo:
- [x] Add support for "default room" which takes over the homepage
- [x] PR Feedback


lib/ret/enums.ex
```diff
@@ -9,3 +9,4 @@ defenum(Ret.SceneListing.State, :scene_listing_state, [:active, :delisted], sche
 defenum(Ret.Avatar.State, :avatar_state, [:active, :removed], schema: "ret0")
 defenum(Ret.AvatarListing.State, :avatar_listing_state, [:active, :delisted, :removed], schema: "ret0")
 defenum(Ret.Asset.Type, :asset_type, [:image, :video, :model], schema: "ret0")
+defenum(Ret.Hub.Privacy, :hub_privacy, [:public, :private, :invite_only], schema: "ret0")
```
gfodor: hmm, this name is OK, I was thinking `access_policy` though.

lib/ret/media_search.ex
```diff
@@ -69,6 +69,10 @@ defmodule Ret.MediaSearch do
     favorites_search(cursor, type, account_id, q)
   end
 
+  def search(%Ret.MediaSearchQuery{source: "public_rooms", cursor: cursor, q: q}) do
```
gfodor: I'd refactor this API to be just `room` search, and for now its an implicit filter of only public rooms. (We could revisit this if/when we add more favorite types, and then could roll favorites into being a filter here and elsewhere, etc)

lib/ret/enums.ex
gfodor: upon reading further, I forgot about `entry_mode` -- perhaps we should consolidate these states (ie, add `closed` to the new enum and update references to `entry_mode`)

lib/ret/media_search.ex
```diff
@@ -654,8 +676,8 @@ defmodule Ret.MediaSearch do
       id: hub.hub_sid,
       url: hub |> Hub.url_for(),
       type: :hub,
+      participant_count: RetWeb.Presence.list("hub:#{hub.hub_sid}") |> Enum.count(),
```
gfodor: hmm. this feels a little fishy to me to be grabbing presence here. i think we should push this functionality into the `Ret.Hub` module, and only expose the count as a return value to minimize our interactions with presence, given its sensitivity

lib/ret_web/channels/hub_channel.ex
```diff
@@ -372,14 +379,18 @@ defmodule RetWeb.HubChannel do
     if account |> can?(update_hub(hub)) do
       name_changed = hub.name != payload["name"]
       member_permissions_changed = hub.member_permissions != payload |> Hub.member_permissions_from_attrs()
+      can_change_privacy = account |> can?(update_hub_privacy(hub))
+      privacy_changed = can_change_privacy and hub.privacy != payload["privacy"]
```
gfodor: will `hub.privacy` be a string or an atom here? I forget how it comes out of the database into the record (i'm guessing you tested this but just checking)

lib/ret_web/channels/hub_channel.ex
```diff
@@ -372,14 +379,18 @@ defmodule RetWeb.HubChannel do
...
 
       stale_fields = []
       stale_fields = if name_changed, do: ["name" | stale_fields], else: stale_fields
       stale_fields = if member_permissions_changed, do: ["member_permissions" | stale_fields], else: stale_fields
+      stale_fields = if privacy_changed, do: ["privacy" | stale_fields], else: stale_fields
 
       hub
       |> Hub.add_name_to_changeset(payload)
       |> Hub.add_member_permissions_to_changeset(payload)
+      |> add_privacy_to_changeset_if_allowed(can_change_privacy, payload)
```
gfodor: i think the idiom we use here is `try_add_privacy_to_changeset` -- also I would push the permission check into that function (and pass along account, hub) instead of passing in the result of the check

lib/ret_web/controllers/api/v1/media_search_controller.ex
```diff
@@ -2,6 +2,18 @@ defmodule RetWeb.Api.V1.MediaSearchController do
   use RetWeb, :controller
   use Retry
 
+  def index(conn, %{"source" => "public_rooms"} = params) do
```
gfodor: same as above, I'd call this `rooms`

priv/repo/migrations/20200115013245_add_privacy_to_hubs.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Ret.Repo.Migrations.AddPrivacyToHubs do
+  use Ecto.Migration
+
+  def change do
+    Ret.Hub.Privacy.create_type()
+
+    alter table(:hubs) do
+      add(:privacy, :hub_privacy, null: false, default: "private")
```
gfodor: after this runs, as a sanity check make sure this lands in the `ret0` schema. I made some changes to the migration tools that should do this automatically now, but in the past we had to specify it in the migration.

gfodor:
One more thing — we should add a integration test on the API just to make sure we don’t ever start accidentally leaking non-public rooms

lib/ret/media_search.ex
netpro2k: Changed the source to "rooms" and added an explicit "public" filter. Doing it implicitly feels like it will be more of a pain to rollback later

lib/ret_web/channels/hub_channel.ex
```diff
@@ -390,7 +390,7 @@ defmodule RetWeb.HubChannel do
       hub
       |> Hub.add_name_to_changeset(payload)
       |> Hub.add_member_permissions_to_changeset(payload)
-      |> add_privacy_to_changeset_if_allowed(can_change_privacy, payload)
+      |> try_add_privacy_to_changeset(account, hub, payload)
```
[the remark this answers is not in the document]
netpro2k: image
Where did I get that idea :thinking:? lol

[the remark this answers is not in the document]
gfodor: ha ha. I guess my brain remembered `maybe_` as `try_` when I wrote that. go go gadget dad-brain

PR #305, opened 2020-01-27
Feature/hubs page resolution

gfodor:
Adds ability for reticulum media resolver to properly crawl hubs entities (rooms, avatars, scenes) and return the appropriate thumbnail + content subtype (`hubs-room`, `hubs-scene`, `hubs-avatar`)


gfodor:
Actually reconsidering this change.

PR #312, opened 2020-02-07
Add support for feature flag to disable sign up

gfodor:
As per https://github.com/mozilla/hubs-cloud/issues/27#issuecomment-582513027 this PR adds the ability for a feature flag to be used to disable account sign up. (A subsequent PR will add additional APIs for administrators to manage accounts.)

This PR also finally adds our first channel test! In general, we should try to get full code coverage on changes going forward now that hubs cloud will increase the risk of reticulum pushes.

Note to self: when pushing, need to ensure emails continue to send in production since that code path is dark in test.


gfodor:
You can hold on reviewing this atm, further work made me realize this should probably become permission checking

gfodor:
OK, done changing this. There is now a `create_account` permission.

lib/ret/hub.ex
```diff
@@ -511,6 +511,9 @@ defimpl Canada.Can, for: Ret.Account do
     hub |> Hub.has_member_permission?(action) or hub |> Ret.Hub.is_owner?(account_id)
   end
 
+  def can?(%Ret.Account{is_admin: true}, :create_account, _), do: true
+  def can?(_account, :create_account, _), do: !AppConfig.get_cached_config_value("features|disable_sign_up")
```
brianpeiris: I think this is the first time we're adding a permission that is independent of a room. If we see more of those, it may make more sense to move all the canada permissions into account.ex instead of hub.ex, or start splitting the permissions into other files.

lib/ret_web/channels/auth_channel.ex
```diff
@@ -22,17 +23,27 @@ defmodule RetWeb.AuthChannel do
     if !Map.get(socket.assigns, :used) do
       socket = socket |> assign(:used, true)
 
-      # Create token + send email
-      %LoginToken{token: token, payload_key: payload_key} = LoginToken.new_login_token_for_email(email)
+      if can?(nil, create_account(nil)) || Account.exists_for_email?(email) do
```
brianpeiris: Hmm, seems a bit odd to shoe-horn this in to canada if it's really just an app config check. Maybe we should just do something like `AppConfig.feature_off(:disable_sign_up)`, if we end up doing more of this in the future.

test/ret_web/channels/auth_channel_test.exs
```diff
@@ -0,0 +1,51 @@
...
+    {:ok, %{session_id: _session_id}, socket} = subscribe_and_join(socket, "auth:testauth", %{})
+    {:ok, socket: socket}
+  end
+
+  test "Login token is created for an email", %{socket: socket} do
+    refute login_token_for_email_exists?(@test_email)
+    refute Account.exists_for_email?(@test_email)
+
+    push(socket, "auth_request", %{"email" => @test_email, "origin" => "test"})
+    :timer.sleep(500)
```
brianpeiris: Is this sleep really necessary?

lib/ret/hub.ex
gfodor: ah yep. i think in general, we probably should eventually do this for all the places where we are currently checking admin role explicitly instead of permissions.

lib/ret_web/channels/auth_channel.ex
gfodor: I had originally had this as a config check, but it did seem to me to be about access control. The permission basically encapsulates security policy defined by the app configuration. I think probably we may need to carve out these kinds of configs separate from `feature` configs, or at least subcategorize them.

Another reason I moved away from the config check is because the next step here is going to be writing an API for managing accounts (the full use case is locking down the IEEEVR server.) From that perspective, the effective check would need to be in two places (unless we want to add a 'model tier') so it felt like a permission check in those two places was more reasonable than a config check to enforce authorization.

gfodor: Oh, also one more thing. The check here is actually a pre-emptive check that doesn't actually enforce policy per se: it just doesn't send the email as a "convenience", since if a user got an email, they'd be denied the ability to create the account if a new account was needed, even if they had a valid login token. So technically we could remove this check with no loss of security.

gfodor: (in that sense, a config check would be less weird to put here vs the check later during account creation, since it's more of a "user interface" mask, not a policy enforcement check)

test/ret_web/channels/auth_channel_test.exs
gfodor: Yep, unfortunately the channel messaging going on here is not request/response, we currently (perhaps erroneously) do not send a reply when the email is sent. I don't remember if there was an intentional reason to do this, but without a reply the only way to test the side effects of the message is to just wait long enough.

PR #314, opened 2020-02-07
Add feature flag to require account when joining a room

gfodor:
Adds a feature flag which will force users to be signed into an account to be in a room.

Also fixes a longstanding bug where bad tokens being passed to the auth channel caused error spam.


gfodor:
Yeah fair enough, I have been a little scared to start adding the elusive hub channel test, but I will do it.

brianpeiris:
👍 lgtm

PR #315, opened 2020-02-12
Basic v1 account creation API

gfodor:
Adds a basic API for creating accounts and looking up accounts by email, accessible to administrator accounts only.

The create API attempts to be a first stab at a idiomatic public API. It has a few characteristics:

- Allows single or multiple record posting, with multiple records being handled via an HTTP 207 response
- Uses JSON schema for validation
- Has fairly robust error handling (at least compared to what we've had in the past :))

Note the lookup API uses a POST because otherwise emails would risk being logged somewhere.

### DIFF
--- a/lib/ret_web/api_helpers.ex
+++ b/lib/ret_web/api_helpers.ex
@@ -1,0 +1,85 @@
+defmodule RetWeb.ApiHelpers do
+  import Plug.Conn
+
+  # In an API create routine, call this with a handler.
+  # The handler will get passed the (validated) record and source and should return either:
+  #   { :ok, { http_status, result } }
+  #   { :error, [ { code, details, source } ] }
+  # TODO dialyzer this
+  def exec_api_create(conn, %{"data" => data}, schema, handler), do: process_create_records(conn, data, schema, handler)
+
+  def exec_api_create(conn, _invalid_params, _schema, _handler) do
+    conn |> send_error_resp([{:MALFORMED_REQUEST, "Missing 'data' property in request.", nil}])
+  end
+
+  defp process_create_records(conn, record, schema, handler) when is_map(record) do
+    case ExJsonSchema.Validator.validate(schema, record, error_formatter: Ret.JsonSchemaApiErrorFormatter) do
+      :ok ->
+        case handler.(record, "data") do
+          {:ok, {status, result}} ->
+            conn |> send_resp(status, %{"data" => result} |> Poison.encode!())
+
+          {:error, errors} ->
+            conn |> send_error_resp(errors)
+        end
+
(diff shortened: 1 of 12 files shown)


lib/ret_web/controllers/api/v1/account_search_controller.ex
```diff
@@ -0,0 +1,16 @@
...
+  alias Ret.{Account}
+  alias RetWeb.Api.V1.{AccountView}
+
+  # Account lookup, is a POST because lookup contains sensitive information
+  def create(conn, %{"email" => email}) do
+    with %Account{} = account <- Account.account_for_email(email) do
+      record = Phoenix.View.render(AccountView, "show.json", account: account)
+      conn |> send_resp(200, %{records: [record]} |> Poison.encode!())
+    else
+      _ -> conn |> put_status(:not_found) |> halt()
```
brianpeiris: Would it make sense to respond with an `{errors: ...}` json in this case as well? It's generally nicer for an API to always return json if it normally returns json.

lib/ret_web/controllers/api/v1/account_controller.ex
```diff
@@ -0,0 +1,34 @@
+defmodule RetWeb.Api.V1.AccountController do
+  use RetWeb, :controller
+  import RetWeb.ApiHelpers
+
+  alias Ret.{Account}
+  alias RetWeb.Api.V1.{AccountView}
+
+  # TODO move to a file
```
brianpeiris: Remove/resolve TODO

lib/ret_web/api_helpers.ex
```diff
@@ -0,0 +1,86 @@
+defmodule RetWeb.ApiHelpers do
+  import Plug.Conn
+
+  # In an API create routine, call this with a handler.
+  # The handler will get passed the (validated) record and source and should return either:
+  #   { :ok, { http_status, result } }
+  #   { :error, [ { code, details, source } ] }
+  # TODO dialyzer this
```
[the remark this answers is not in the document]
gfodor: ah that was more for longer term once we start caring about dialyzer. will take out for now

lib/ret_web/controllers/api/v1/account_search_controller.ex
gfodor: i think for the 404's this seems right. we're already `halt()`ing for 401s in other places, but i'll update the 404 paths.

lib/ret_web/controllers/api/v1/account_controller.ex
gfodor: this was also longer term (ie when we want to have swagger docs, etc). will remove

PR #317, opened 2020-02-13
Add account identities

gfodor:
Includes https://github.com/mozilla/reticulum/pull/312 https://github.com/mozilla/reticulum/pull/313 https://github.com/mozilla/reticulum/pull/314 https://github.com/mozilla/reticulum/pull/315 

Adds a new `Identity` concept which can optionally be attached (in a `has_one`) to an account. If an account has an identity (which has a name), the identity name will be broadcast in presence as `identityName` (in contrast to `displayName`) when the account is connected to a room. The `communityIdentifier` value in presence is deprecated -- the discord integration will set the `identityName` to the discord user's identity in the case of a bound room.

Also renames `records` -> `data` in the API which was making things confusing between single and multiple result sets.

Note that these new identities can only currently be set by admins, either through the account API or the admin console.


lib/ret/identity.ex
```diff
@@ -0,0 +1,24 @@
...
+    belongs_to(:account, Ret.Account, references: :account_id)
+
+    timestamps()
+  end
+
+  def changeset_for_new(account, params \\ %{}) do
+    %Identity{}
+    |> cast(params, [:name])
+    |> validate_required([:name])
+    |> validate_length(:name, min: 2, max: 64)
```
brianpeiris: `min: 2` seems kinda arbitrary, why not `min: 1`?

gfodor: it is definitely arbitrary :) no strong opinion, I can make it 1

PR #323, opened 2020-02-22
Add per-room caps to room settings and entry

netpro2k:
Goes with https://github.com/mozilla/hubs/pull/2156, adds `member_cap` as a property to rooms that can be updated by owners. Also fixes issues where the `entering` state was not being cleared after the user entered the room. The maximum cap is hardcoded to be 64 currently and the client has a hardcoded default of 24. We likely want to expose these as settings in the admin console.

Currently the limits are only enforced client side. We would ideally also prevent ever "entering" the room if it is full, and also prevent sending any `naf` traffic unless you were in the entered state. Janus should probably also enforce that only people who have "entered" can create outgoing streams.

There is also a tiny unrelated change to add some spacing to the sign in email, purely to make the link easier to click in the server logs when running locally.

- [x] ~~Decide if/how we want to expose default/maximum room cap to admin console (maximum should probably line up with actual Janus room caps~~ We do, exposed in the admin panel under a new "Rooms" category in app settings.
- [x] ~~Decide if/how we want to do server side enforcement of room caps.~~ Not in scope for this PR.


lib/ret/hub.ex
```diff
@@ -117,11 +119,13 @@ defmodule Ret.Hub do
...
 
-  def add_meta_to_changeset(changeset, attrs) do
+  def add_attrs_to_changeset(changeset, attrs) do
     changeset
-    |> cast(attrs, [:name, :description])
+    |> cast(attrs, [:name, :description, :member_cap])
     |> validate_required([:name])
     |> validate_length(:name, max: 64)
+    |> validate_length(:description, max: 64_000)
+    |> validate_number(:member_cap, greater_than_or_equal_to: 0, less_than_or_equal_to: 64)
```
gfodor: I'd argue to bump the max to 100

priv/repo/migrations/20200220012135_add_cap_to_rooms.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Ret.Repo.Migrations.AddCapToRooms do
+  use Ecto.Migration
+
+  def change do
+    alter table("hubs") do
+      add(:member_cap, :integer, null: false, default: 0)
```
gfodor: we should run an `UPDATE` to set the member_cap properly on existing rows to 64.

gfodor: you can either split this migration into a `up/down` and do the alter manually, or leave this one and have a separate migration with an `up/down` that just runs the update in the `up`

lib/ret/hub.ex
```diff
@@ -83,6 +83,8 @@ defmodule Ret.Hub do
 
     field(:allow_promotion, :boolean)
 
+    field(:member_cap, :integer)
```
gfodor: maybe `max_members`?

gfodor: also terminology is a bit rough. the list of people in the app is the 'member list', which includes lobby. maybe we should just call these "lobby members" and "room members"? (and hence, call this `max_room_members`)?

netpro2k: We also have already `max_occupant_count`, which is the stats field used to track the maximum number of people who ever entered...

lib/ret/hub.ex
netpro2k: Seems fine. Probably worth just exposing this to the admin console as well since we are going to expose a default anyway.

priv/repo/migrations/20200220012135_add_cap_to_rooms.exs
netpro2k: Hmm, there are no existing room counts. My intent was that 0 means "server default", but maybe that is too weird.

lib/ret/hub.ex
gfodor: ah that makes sense. room_member_cap then?

priv/repo/migrations/20200220012135_add_cap_to_rooms.exs
gfodor: oh, sorry, duh. i think it should be nullable if we want to have semantics around a default

lib/ret/hub.ex
netpro2k: sticking with `member_cap` since we don't prefix any other room properties with `room` except for the global `room_id`

lib/ret/hub.ex
```diff
@@ -245,13 +253,29 @@ defmodule Ret.Hub do
...
 
-  defp presence_entry_to_member_count(%{metas: [%{context: %{"discord" => true}}]}), do: 0
-  defp presence_entry_to_member_count(_presence_entry), do: 1
+  def room_size_for(%Hub{} = hub) do
+    hub.room_size || AppConfig.get_cached_config_value("features|default_room_size")
+  end
+
+  def room_size_for(%Hub{} = hub) do
+    hub.room_size || AppConfig.get_cached_config_value("features|default_room_size")
+  end
```
brianpeiris: This is repeated.

netpro2k: woops good catch, merge snafu

PR #324, opened 2020-02-26
Feature/custom csp rules

gfodor:
Goes with https://github.com/MozillaReality/ita/pull/8

Adds support for administrator set CSP rules, HTTP headers, and custom scripts. Individual headers and scripts can be set for the homepage, room, scene, and avatar pages.

Did a bunch of refactoring in page controller to make it so scripts are slightly more nicely handled.

I also realized a fairly major problem: we were not doing any kind of string escaping in the template injection in habitat. The net result is you could break reticulum in the admin console by putting a string with a quote in it in one of the configs. Fortunately, there's a simple fix: use the `toToml` helper, which conveniently injects habitat configs as toml. This also came in quite handy for the new CSP/http/script features. In general, we should probably use `toToml` for all new configs -- I was nervous about paving over *all* of them since the current semantics we have now always add quotes. So, if a value was registered as `null` in the habitat ring, it would appear as an empty string in the generated config.toml, whereas if we converted it via `toToml`, I suspect it would end up as a `null` (and `nil` in Elixir.)

Note that when pushing this out we'll need to do a bunch of refactoring of our own configs for the CSP in hubs.mozilla.com configs -- this drops the legacy "full CSP override" mechanism which was not very useful, since the server generates the CSP now. The new method is to add additional rules beyond the ones the server generates.

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -42,14 +42,15 @@ defmodule RetWeb.PageController do
   end
 
   defp render_scene_content(%t{} = scene, conn) when t in [Scene, SceneListing] do
-    {app_config, app_config_script, app_config_csp} = generate_app_config()
+    {app_config, app_config_script} = generate_app_config()
 
     scene_meta_tags =
       Phoenix.View.render_to_string(RetWeb.PageView, "scene-meta.html",
         scene: scene,
         ret_meta: Ret.Meta.get_meta(include_repo: false),
         translations: app_config["translations"]["en"],
-        app_config_script: {:safe, app_config_script}
+        app_config_script: {:safe, app_config_script |> with_script_tags},
+        extra_script: {:safe, get_extra_script(:scene) |> with_script_tags}
       )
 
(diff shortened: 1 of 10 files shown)


gfodor:
Actually hold on reviewing, I'm also going to add custom header tags.

gfodor:
OK this is ready to go. Added custom HTML as well.

lib/ret_web/controllers/page_controller.ex
```diff
@@ -589,7 +605,44 @@ defmodule RetWeb.PageController do
...
+
+  defp extra_header_to_tuple(header) do
+    [name | rest] = header |> String.split(~r/:\s*/) |> Enum.map(&String.trim/1)
+
+    name =
+      name
+      |> String.downcase()
+      |> String.replace(~r/[^a-z0-9_-]/, "")
+
+    value = rest |> Enum.join(":")
```
brianpeiris: The split and join here could convert values from "foo: bar : baz" to "foo:bar:baz". Might be less surprising if you used the `parts` option on `split` to just extract the name.

PR #329, opened 2020-03-04
Selectively filter packets to reduce reliance upon intercepts

gfodor:
It turns out that `handle_out` is quite expensive, given that it always results in N re-encodings of each message, where N is the number of subscribers to the topic. We are using `handle_out` for a variety of purposes, but the vast majority of our messages are `naf` (and now `nafr`) and hence the `handle_out` we are using there specifically leads to a ton of overhead under load.

We use `handle_out` over NAF messages to deal with two concerns:
- If a user is blocked, or is blocking people, we intercept outgoing messages to prevent delivery in both directions.
- If the user has requested (such as in an iframe) to `block_naf` messages overall. In practice, we only do this for hubs which are embedded in iframes - we run with this assumption for now in the interest of optimization.

It turns out, these are the minority of connections typically, so if we avoid the use of `handle_out` in the majority case, the performance concerns should go away.

To do so, we now no longer intercept `naf` and `nafr` directly, but instead intercept two new event types, `maybe-nafr` and `maybe-naf`, which are only used internally. When a message comes in, if the socket is blocked, has blockers, or the hub is embedded, we send the message with the filtered variant, meaning all receivers will have an opportunity to filter it.

For the blocking case, this means if user A is blocking B, user A's messages will be flagged as filterable when sent and so will user B's, meaning anytime a block happens all of user A and B's 
messages will incur the overhead of filtering (since they may be the blockee, or may be the blocker.) We could reduce this overhead somewhat by only filtering messages from the blocker (to ensure they are not delivered to blocked users for privacy) and updating the client to drop packets for blocked users on the floor, but this doesn't seem worth the extra complexity.

For the `block_naf` case, the condition is that if *anyone* in the room is blocking NAF, it means all messages must be sent over the filtered event type, to ensure those sockets blocking NAF will have them filtered. Unfortunately, the only way to have global awareness of if anyone is blocking NAF would be to register it into presence, and this seems like a bad idea given the costs of checking presence on every incoming message. However, since we only currently use the NAF blocking feature for iframe embeds, for now we can use that as the way to determine if all messages should be intercepted or not for a topic. So, if a room is embedded (meaning it was ever seen in an iframe, even if not being viewed in an iframe in the current socket connection) then we just start filtering all packets. This does mean that if we start using `block_naf` in other contexts, it will simply not work mysteriously, and we'll need to come up with some kind of method to know up-front if messages for that room need to be globally intercepted for filtering.

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -969,7 +969,7 @@ defmodule RetWeb.HubChannel do
       end
 
     # Each channel connection needs to be aware if there are, or ever have been,
-    # embeddings of this hub (see should_intercept_naf?)
+    # embeddings of this hub (see internal_naf_event_for/2)
     socket = socket |> assign(:has_embeds, hub.embedded)
 
     push_subscription_endpoint = params["push_subscription_endpoint"]
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -66,6 +66,7 @@ defmodule RetWeb.HubChannel do
     |> assign(:blocked_session_ids, %{})
     |> assign(:blocked_by_session_ids, %{})
     |> assign(:has_blocks, false)
+    |> assign(:has_embeds, false)
     |> perform_join(
       hub,
       context,
@@ -518,6 +519,8 @@ defmodule RetWeb.HubChannel do
     {:noreply, socket}
   end
 
+  # NOTE: block_naf will only work if the hub is embedded. We *only* enable packet filtering
+  # (and therefore, only respect block_naf) when a hub is embedded (or if there are blocks on the socket.)
   def handle_in("block_naf", _payload, socket), do: {:noreply, socket |> assign(:block_naf, true)}
   def handle_in("unblock_naf", _payload, socket), do: {:noreply, socket |> assign(:block_naf, false)}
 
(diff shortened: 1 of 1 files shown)


lib/ret_web/channels/hub_channel.ex
```diff
@@ -476,13 +489,21 @@ defmodule RetWeb.HubChannel do
...
     broadcast_from!(socket, event, payload |> payload_with_from(socket))
     {:noreply, socket}
   end
 
   def handle_in("unblock" = event, %{"session_id" => session_id} = payload, socket) do
-    socket = socket |> assign(:blocked_session_ids, socket.assigns.blocked_session_ids |> Map.delete(session_id))
+    socket =
+      socket
+      |> assign(:blocked_session_ids, socket.assigns.blocked_session_ids |> Map.delete(session_id))
+      |> assign(:has_blocks, true)
```
brianpeiris: Why are we setting `has_blocks` to true on `unblock`? Same for line 609 below. Shouldn't we instead set it to false if `blocked_session_ids` is empty?

gfodor: nice catch. bug on my part.

gfodor: PR updated. we need to actually set it properly when either map is non-empty, across all calls

robertlong:
Notes from @gfodor when we merge:

```
For that one, once it is on smoke, please verify in the iframe test you don’t see any incoming NAF messages when on the iframe landing page even if others are in the room
The test page is linked in the smoke test room
```

PR #334, opened 2020-03-10
Add special handling for offline twitch streams

netpro2k:
Returns the same response as if we were to screenshot the webpage (which is what the current fallback is) except it skips screenshotting, uses a static "stream offline" image, and adds the `offline_stream: true` meta tag to the media response so the client can deal with it in a different way if it chooses (for example exposing different refresh behavior). The caching semantics have not been changed in this PR.


gfodor:
Might be more future proof to make it a enum-like thing, eg ‘media_status:
offline_stream’ or ‘stream_status: offline’

.gitignore
```diff
@@ -20,8 +20,7 @@ priv/server.key
 priv/server.pem
 
 # generated static files
-priv/static/js/
-priv/static/css/
+priv/static/
```
johnshaughnessy: Do we need to retain https://github.com/mozilla/reticulum/blob/master/priv/static/favicon-spoke.ico and https://github.com/mozilla/reticulum/blob/master/priv/static/spoke-preview.png ?

netpro2k: From what I can tell these are all generated files so I don't think so, but leaving them in place just in case.

PR #336, opened 2020-03-11
Add update hub route

robertlong:
This PR adds the update route for hubs and the `scene_id` field to hubs views and media search results. This allows us to manage existing rooms via http.

### DIFF
--- a/lib/ret_web/controllers/api/v1/hub_controller.ex
+++ b/lib/ret_web/controllers/api/v1/hub_controller.ex
@@ -45,6 +45,32 @@ defmodule RetWeb.Api.V1.HubController do
     end
   end
 
+  def update(conn, %{"id" => hub_sid, "hub" => %{"scene_id" => scene_id}} = params) do
+    account = Guardian.Plug.current_resource(conn)
+    scene = 
+    case Scene.scene_or_scene_listing_by_sid(scene_id) do
+      nil -> conn |> send_resp(422, "scene not found")
+      scene ->
+        case Hub |> Repo.get_by(hub_sid: hub_sid) do
+          %Hub{} = hub -> 
+            if account |> can?(update_hub(hub)) do
+              hub
+              |> Hub.add_attrs_to_changeset(params["hub"])
+              |> Hub.changeset_for_new_scene(params["hub"])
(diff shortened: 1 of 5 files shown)


lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -45,6 +45,52 @@ defmodule RetWeb.Api.V1.HubController do
...
+    
+    if not is_nil(scene) do
+      changeset |> Hub.changeset_for_new_scene(scene)
+    end
+
+    if not is_nil(hub_params["member_permissions"]) do
+      changeset |> Hub.add_member_permissions_to_changeset(hub_params)
+    end
+
+    if not is_nil("allow_promotion") do
```
brianpeiris: Is this supposed to be `hub_params["allow_promotion"]`?

lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -45,6 +45,52 @@ defmodule RetWeb.Api.V1.HubController do
     end
   end
 
+  def update(conn, %{"id" => hub_sid, "hub" => hub_params}) do
+    account = Guardian.Plug.current_resource(conn)
+
+    case Hub |> Repo.get_by(hub_sid: hub_sid) |> Repo.preload([:created_by_account, :hub_bindings, :hub_role_memberships]) do
+      %Hub{} = hub -> 
+        if account |> can?(update_hub(hub)) do
+          update_with_hub(conn, account, hub, hub_params)
```
brianpeiris: nit: generally nicer to use the `conn |> ...` form where possible.

robertlong: Is it better here? We're not chaining any methods.

brianpeiris: After having gotten used to our elixir code, it's just a little surprising to see the traditional method call form, especially with operations related to `conn`s. More a matter of consistency than technical correctness. I also take it to mean that "we're piping a conn through this function and it likely returns a conn" as opposed to a method that produces some side effect and doesn't return a conn. Anyway, it was just a nit.

PR #341, opened 2020-03-12
Add more tests and fix update route

robertlong:


lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -72,17 +72,23 @@ defmodule RetWeb.Api.V1.HubController do
 
   defp update_with_hub_and_scene(conn, account, hub, scene, hub_params) do
     changeset = hub |> Hub.add_attrs_to_changeset(hub_params)
-    
-    if not is_nil(scene) do
-      changeset |> Hub.changeset_for_new_scene(scene)
+
+    changeset = if not is_nil(scene) do
```
robertlong: I couldn't find a way to do this with pattern matched functions. I also found my half-working attempt to be hard to read and of little benefit. So I chose to stick with these conditionals.

robertlong: Nvm, @brianpeiris helped me out. Thanks!

PR #343, opened 2020-03-14
add flying to member_permissions

InfiniteLee:
see: https://github.com/mozilla/hubs/pull/2224


gfodor:
This seems a bit weird to me since it’s the second security through
obscurity permission (the first I think is “camera mode”) - to actually do
this properly clients would need to reject locomotion if it’s off nav-mesh.
(Not gonna block, but highlighting we may want to reconsider since anything
like this could be bypassed w a hacked client, unlike most other
permissions)

InfiniteLee:
I also agree it's weird- but arguably the range of abuse from someone flying is far less than the other cases (media spawning).

PR #345, opened 2020-03-20
Account disabling

brianpeiris:
- Adds a `state` enum field to accounts with "disabled" and "enabled" values
- Denies all permissions to disabled accounts, including room joins
- Denies access to most API routes, including hub creation
- Prevents disabled accounts from signing in


lib/ret_web/channels/hub_channel.ex
```diff
@@ -876,6 +876,18 @@ defmodule RetWeb.HubChannel do
...
   end
 
+  defp join_with_hub(
+         %Hub{},
+         %Account{},
+         _socket,
+         _context,
+         %{
+           hub_requires_oauth: false,
+           account_can_join: false
```
johnshaughnessy: Are there any cases where `account_can_join` is `false` but the join is _not_ denied? Wondering how this and the `join_with_hub` case directly below relate to the `can?(_account, :join_hub ...` checks in hub.ex

brianpeiris: `account_can_join` is derived from the `:join_hub` permission check, so it should be impossible for account_can_join to be false if the join permission is denied. The case below only covers Discord-bound hubs.

lib/ret_web/channels/auth_channel.ex
```diff
@@ -23,7 +23,9 @@ defmodule RetWeb.AuthChannel do
     if !Map.get(socket.assigns, :used) do
       socket = socket |> assign(:used, true)
 
-      if can?(nil, create_account(nil)) || Account.exists_for_email?(email) do
+      account_disabled = Account.exists_for_email?(email) && Account.account_for_email(email).disabled
```
gfodor: `exists_for_email?(email)` round trips to db, probably should just do it once.

gfodor:
consider making this an enum - we may have varying levels of account access denial (eg perhaps an account could be on probation, indicating they can access their own data but cannot join rooms)

lib/ret_web/channels/hub_channel.ex
johnshaughnessy: What I mean to ask is whether a match on `account_can_join: false` is a sufficient condition for us to want to `deny_join`, as in: 
```
  defp join_with_hub(
         %Hub{},
         %Account{},
         _socket,
         _context,
         %{
           account_can_join: false
         }
       ),
       do: deny_join()
```
I'm trying to understand whether it's possible for `account_can_join` to be `false` but still fall through to the final `join_with_hub`.

gfodor:
Maybe the account lookup functions (exists_for_email, etc) should just return values as though the account doesn’t exist if disabled?

lib/ret_web/channels/hub_channel.ex
brianpeiris: Ah, I see. I don't think we can do that because we want to reply with `require_oauth` in another scenario below.

brianpeiris:
Returning `nil` on account lookups doesn't really pan out since we do sometimes want to retrieve the account to check other things in addition to checking if the account is indeed disabled. It also would essentially make the user "anonymous" in some scenarios, which is probably not the right behavior for a disabled account.
I've changed the field to an enum with "enabled" and "disabled" values, and I've added a ForbidDisabledAccounts plug to most of the api routes.
Also added tests.

gfodor:
Lgtm!

PR #349, opened 2020-03-30
CircleCI test run and code format check

brianpeiris:
See runs at https://circleci.com/gh/mozilla/reticulum


gfodor:
This is great, lgtm. I wonder if there is a way to get it to fail if there are build warnings?

Issue #346, opened 2020-03-21
Migrations can fail to run when server under load


gfodor:
Migrations will fail if the server is getting hit when starting up and errors are flying out because database is out of date.

probably boils down to integrating something like https://github.com/mspanc/ecto_boot_migration

blairmacintyre:
It would be REALLY good if there was a way in the admin panel to "close" the site temporarily, esp for hubs-cloud.  Something like shutdown on unix:  enter a reasons (WHATEVER REASON) and time (default 60, could be 0, seconds), and everyone starts getting "system shutting down (WHATEVER REASON) in 60, 59, ...) then everyone is kicked.

Landing page should change immediately to a message, and all new entry prohibited.

gfodor:
You can do this - in the stack, flip the site to offline mode. It can
redirect to a url.

PR #351, opened 2020-03-31
Migrate database before server starts

gfodor:
Fixes https://github.com/mozilla/reticulum/issues/346

Migrates database on startup, before server begins listening. Tested locally in dev, will do a test deploy to a hubs cloud stack just in case.


gfodor:
hah jinx, I thought of that too, just pushed a new commit to fix it.

PR #352, opened 2020-04-02
Feature/youtube rate limiting

gfodor:
Adds mitigations for yt-dl rate limiting:

- Each server can only do one youtube.com resolution every 4 seconds (via a mutex)
- We now read the `expire` query var from the response, and use that to set the cache ttl. We expire URLs 60 seconds before they are due to expire on the youtube side.

Before this rolls out, yt-dl should be moved to a locally running service on the reticulum nodes.


lib/ret/media_resolver.ex
```diff
@@ -45,6 +48,40 @@ defmodule Ret.MediaResolver do
...
+
+      {:commit, %Ret.ResolvedMedia{uri: %URI{query: youtube_query}} = resolved_media} = res
+
+      # YouTube returns a 'expire' which has timestamp of expiration.
+      resolved_media =
+        with parsed_youtube_query <- URI.decode_query(youtube_query),
+             expire when is_binary(expire) <- Map.get(parsed_youtube_query, "expire"),
+             {expire_s, _} <- Integer.parse(expire),
+             # Don't use custom TTL if expires in < 2 minutes
+             ttl_s when ttl_s > 120 <- expire_s - System.system_time(:second) do
```
netpro2k, line 66: Did you actually see this? In the videos I tested expire always seemed to be 6 hours. In any case, probably makes sense to use the ttl regardless no? Since the link will literally be broken after it expires, so caching it is wrong.

lib/ret/media_resolver.ex
```diff
@@ -45,6 +48,40 @@ defmodule Ret.MediaResolver do
...
+          # Expire a minute early
+          resolved_media |> Map.put(:ttl, ttl_s * 1000 - 60000)
+        else
+          _ -> resolved_media
+        end
+
+      request_duration_ms = System.system_time(:millisecond) - now
+      remaining_duration_ms = @youtube_ms_per_request_rate_limit - request_duration_ms
+
+      if remaining_duration_ms > 0 do
```
netpro2k, line 76: Not fully following how the mutex API works, but doesn't this add delay to every request, even if no other request is actually pending? Seems like we would only want to wait on getting the mutex in a subsequent request, but not delay the first request that came in. (basically the piece I am missing here is when does the outer block actually return a result? Is it before or after this sleep?

lib/ret/media_resolver.ex
gfodor: thats a good point, i may have had the wrong mental model. i thought that the URL returned by youtube-dl would be the same up until expiration, so eventually the window would tighten. beyond that tho the other point you raise is right, i'm not totally sure what i thought i was accomplishing here- probably in part due to paternity brain

netpro2k: No I think youtube-dl will give you a new url each time, just the way youtube's CDN and such work is that the signed url it generates is only valid for a certain window (seemingly always 6 hours right now).

lib/ret/media_resolver.ex
gfodor: it'd be after the sleep, so good point. i think to do what you want we need to grab the mutex, do the request, and then spawn a process in the critical section to free the mutex after 4 seconds (and return immediately once the work is done.) A little hairy, but much better than stalling the requests.

PR #353, opened 2020-04-02
Scene deletion WIP

robertlong:
To Do:
- [ ] Decide if we are deleting or delisting the associated scene listing
- [ ] Decide what to do about scenes remixed from the scene / scene listing
- [ ] Decide if the scene deletion API should also work when a scene listing id is provided
- [ ] Write integration tests for the success and two error cases

### DIFF
--- a/lib/ret/scene.ex
+++ b/lib/ret/scene.ex
@@ -9,6 +9,7 @@ end
 defmodule Ret.Scene do
   use Ecto.Schema
   import Ecto.Changeset
+  import Ecto.Query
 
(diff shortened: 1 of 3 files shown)


robertlong:
I would prefer not to delete remixed scenes when you delete the parent. We could do this by adding a `:deleted` state to `Scene` and `SceneListing`. We would delete owned files and all other metadata when a user deletes the scene/scene listing, but the remixed scenes would maintain the reference. We would show a "Deleted Scene" page when the user tried to navigate to that page.

robertlong:
We need to revisit this and deleting assets entirely.

PR #357, opened 2020-04-08
Add allow_local_origins flag

robertlong:
Mobile development is currently a pain because dev.reticulum.io only whitelists `hubs.local` and not local ip addresses like `192.168.1.5`. This PR adds the allow_local_origins flag which lets you allow local ips to access the API and use the CORS proxy.

To Do:
- [ ] Test more
- [ ] Merge https://github.com/MozillaReality/reverse_proxy_plug/pull/1


lib/ret_web/endpoint.ex
```diff
@@ -7,11 +7,35 @@ defmodule RetWeb.Endpoint do
   def get_cors_origins, do: Application.get_env(:ret, RetWeb.Endpoint)[:allowed_origins] |> String.split(",")
   def get_cors_origin_urls, do: get_cors_origins() |> Enum.filter(&(&1 != "*")) |> Enum.map(&URI.parse/1)
 
+  def local_address_ranges(), do: [CIDR.parse("192.168.0.0/16"), CIDR.parse("10.0.0.0/8"), CIDR.parse("172.16.0.0/12")]
```
robertlong, line 10: Can I make this a constant or memoize it somehow? Parsing these on every request is probably going to get expensive.

mix.exs
```diff
@@ -51,7 +51,7 @@ defmodule Ret.Mixfile do
       {:httpoison, "~> 1.5"},
       {:poison, "~> 3.1"},
       {:ecto_autoslug_field, "~> 2.0"},
-      {:cors_plug, "~> 2.0"},
+      {:cors_plug, "~> 2.0.2"},
```
robertlong, line 54: cors_plug 2.01 added the ability to make `origin` a function, so I had to upgrade it.

mix.exs
```diff
@@ -76,12 +76,13 @@ defmodule Ret.Mixfile do
       {:ua_parser, "~> 1.5"},
       {:download, git: "https://github.com/gfodor/download.git", branch: "reticulum/master"},
       {:reverse_proxy_plug,
-       git: "https://github.com/mozillareality/reverse_proxy_plug.git", branch: "reticulum/master"},
+       git: "https://github.com/robertlong/reverse_proxy_plug.git", branch: "feature/allow_origin"},
```
robertlong, line 79: We still need to merge this.

habitat/config/config.toml
```diff
@@ -22,6 +22,7 @@ keyfile = "{{ pkg.svc_files_path }}/ssl.key"
 
 [ret."Elixir.RetWeb.Endpoint"]
 allowed_origins = "{{ cfg.security.cors_origins }}"
+allow_local_origins = {{ cfg.security.allow_local_origins }}
```
robertlong, line 25: I'll need to configure this on dev.reticulum.io. That's done in Jenkins, right?

PR #359, opened 2020-04-14
Feature/coturn

gfodor:
Adds support for TURN authentication and coturn based secret managment/rotation.

### DIFF
--- a/lib/ret/application.ex
+++ b/lib/ret/application.ex
@@ -26,15 +26,16 @@ defmodule Ret.Application do
           Ecto.Adapters.SQL.query!(Ret.Repo, "CREATE SCHEMA IF NOT EXISTS coturn")
         end
 
+        schemas =
+          if coturn_enabled do
+            ["ret0", "coturn"]
+          else
+            ["ret0"]
+          end
+
         Ecto.Adapters.SQL.query!(
           Ret.Repo,
-          "ALTER DATABASE #{db_name} SET search_path TO ret0#{
-            if coturn_enabled do
-              ", coturn"
-            else
-              ""
-            end
-          }"
+          "ALTER DATABASE #{db_name} SET search_path TO #{schemas |> Enum.join(",")}"
         )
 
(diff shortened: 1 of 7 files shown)


lib/ret/application.ex
```diff
@@ -14,9 +14,29 @@ defmodule Ret.Application do
...
+
+        Ecto.Adapters.SQL.query!(
+          Ret.Repo,
+          "ALTER DATABASE #{db_name} SET search_path TO ret0#{
+            if coturn_enabled do
+              ", coturn"
+            else
+              ""
+            end
+          }"
```
brianpeiris, line 37: nit: this would be a bit easier to read if you had a list of schemas and joined them with a comma.

lib/ret/coturn.ex
```diff
@@ -0,0 +1,51 @@
...
+      {:commit, nil}
+    end
+  end
+
+  def enabled? do
+    !!realm()
+  end
+
+  defp realm do
+    Application.get_env(:ret, __MODULE__)[:realm]
```
brianpeiris, line 49: What purpose does it serve to be able to change the realm in the config?

priv/repo/migrations/20200413202224_add_coturn_schema.exs
```diff
@@ -0,0 +1,23 @@
+defmodule Ret.Repo.Migrations.AddCoturnSchema do
+  use Ecto.Migration
+
+  def up do
+    execute("create schema if not exists ret0")
```
brianpeiris, line 5: typo?
```suggestion
    execute("create schema if not exists coturn")
```

lib/ret/coturn.ex
gfodor: coturn reads from these tables to check credentials, and is also configured with a realm. so the realms need to match so it looks up the right rows

priv/repo/migrations/20200413202224_add_coturn_schema.exs
gfodor: indeed!

PR #365, opened 2020-04-22
Feature/coturn april update

gfodor:
First deals with some issues related to shutdown:

First, the graceful connection draining logic broke with Phoenix 1.4 due to unsupported behavior in the the_end dependency. We now use a fork that has the necessary one line change to fix it. (The upstream project seems dead.)

Next, when we start reticulum on Hubs Cloud nodes, it uses the burned in AMI version to start. However, if there is a new version, habitat will SIGTERM that process and then install + run the new version. This is problematic when a new system is coming online, since it may receive the SIGTERM during database migrations, and leave the database in an inconsistent state. (Ecto properly runs each migration in a transaction unless we opt-out of it, but it sadly does not transactionally update the schema versions table in that transaction, so a poorly timed SIGTERM can leave the database in a state where a migration could be attempted a second time, which breaks things.)

This adds a new SIGTERM handler which is temporarily registered during application startup during migrations. If the signal is seen during this time, it is ignored, and the application shuts down after the migrations have finished. After migrations finish, if no signal was seen, the default signal handler is re-registered, so there are no long term side effects of this temporary override during startup.

Also, in Hubs Cloud, another common failure was that because the repo migration on AWS is happening in the context of a `transaction` scope `pgbouncer` connection, it was easy for the `search_path` to be lost during migrations (which led to us having to be explicit about `ret0` everywhere it was not being injected by the migration runner automatically.) This is now fixed: there is a separate `SessionLockRepo` that can be used to run SQL in the context of a session (instead of a transaction) scoped pgbouncer connection, where it is necessary to retain connection environment variables across several statements. This is now done for migrations.

Drops the unneeded `coturn` user from our migration chain - this ended up being unnecessary and clunky to deal with in Hubs Cloud. Since this migration already ran on some Hubs Cloud servers (without TURN) that user is on those systems, but doesn't have any purpose given that TURN is not on those servers.

The TURN secrets are now rotated on startup, because otherwise there is a catch-22: the cron won't add the first secret right away *and* it is also skipped if nobody is connected. So there's no way to get the initial secret.

Drops the `transport` field from the TURN response from reticulum, it's now assumed that TURN is available over both UDP/DTLS and TCP/TLS.


lib/ret/application.ex
```diff
@@ -11,39 +11,54 @@ defmodule Ret.Application do
...
+          #
+          # If application is killed mid-migration, then it's possible for schema migrations
+          # table to not accurately reflect the migrations which have ran.
+          Ret.DelayStopSignalHandler.delay_stop()
+
+          try do
+            Ecto.Migrator.run(Ret.SessionLockRepo, priv_path, :up, all: true, prefix: "ret0")
+          after
+            Ret.DelayStopSignalHandler.allow_stop()
           end
```
brianpeiris, line 49: I believe we won't return `repos_pids` if there's an error in the `try`. Should we also `stop_repos` in the `after` clause?

lib/ret/application.ex
```diff
@@ -11,39 +11,54 @@ defmodule Ret.Application do
...
+      end)
 
-        priv_path = Path.join(["#{:code.priv_dir(:ret)}", "repo", "migrations"])
-        Ecto.Migrator.run(Ret.Repo, priv_path, :up, all: true, prefix: "ret0")
-      end
-    end)
+    if repos_pids do
+      # Ensure there are some TURN secrets in the database, so that if system is idle
+      # the cron isn't indefinitely skipped and nobody can join rooms.
+      Ret.Coturn.rotate_secrets(true, Ret.SessionLockRepo)
```
brianpeiris, line 58: Why is `rotate_secrets` dependent on `repos_pids`?

lib/ret/delay_stop_signal_handler.ex
```diff
@@ -0,0 +1,39 @@
...
+      )
+  end
+
+  def allow_stop do
+    :gen_event.call(:erl_signal_server, Ret.DelayStopSignalHandler, {:allow_stop, self()})
+
+    :ok =
+      :gen_event.swap_sup_handler(
+        :erl_signal_server,
+        {Ret.SignalHandler, []},
```
brianpeiris, line 27: What is `Ret.SignalHandler`?

priv/repo/migrations/20200413202224_add_coturn_schema.exs
```diff
@@ -17,35 +15,6 @@ defmodule Ret.Repo.Migrations.AddCoturnSchema do
...
-      do 
-      $$ 
-      begin
-        execute format('grant connect on database %I to coturn', current_database());
-      end;
-      $$;
-      """)
-
-      execute("grant usage on schema coturn to coturn;")
-      execute("grant all privileges on all tables in schema coturn to coturn;")
```
brianpeiris, line 47: Don't we still have to grant access to the coturn schema to some user?

lib/ret/application.ex
gfodor: `after` is like `finally`, if it throws in try this will throw, and hence stop starting the application altogether. (since this is the application start process)

lib/ret/application.ex
gfodor: its a bit roundabout - `repos_pids` is null when mix env is test, and we don't want to rotate the secrets on startup in test. i can refactor even tho its redundant.

lib/ret/delay_stop_signal_handler.ex
gfodor: 😱 
typo. i tested this in production and it seemed to work because migrations were not stomped, but obviously was timing related. nice catch. will need to re-do all the integration tests.

priv/repo/migrations/20200413202224_add_coturn_schema.exs
gfodor: no - the relevant user is now just the superuser `postgres` - we have the same security contract for coturn as we do ret's db access. both are internal servers accessing the db. the fine grained security policy we have is postgrest, which is a SQL proxy, so we have to manage it directly w access control whitelisting and a dedicated user.

lib/ret/application.ex
brianpeiris: Gotcha. Maybe just add this explanation to the comments.

PR #373, opened 2020-04-30
Update hub-meta.html.eex

chamberlain-link:
Removed 'in VR' from "Join others at <%= @hub.name %>, right in your browser."


gfodor:
Thanks for the contribution @chamberlain-link !

PR #375, opened 2020-05-06
Update README.md

shawticus:
Improved the DX with more specific instructions for postgres, setting up loopback alias properly (host and alias were flipped), setting up admin panel for development mode.


shawticus:
Only problem with this repo is the admin panel doesn't work locally, I assumed it's tied to some of the cloud infrastructure. Would be handy to add more tutorial for deving it.

robertlong:
Thanks @shawticus!

PR #379, opened 2020-06-09
Add dev-janus.reticulum.io to dev config

johnshaughnessy:
When running hubs and reticulum locally, hubs fails to connect to the dev-janus server via `https` because its only entries in the CSP are `wss`. This adds the `https` entry for dev-janus.reticulum.io to the CSP so that the fetch to it from hubs succeeds. (https://github.com/mozilla/hubs/blob/dbc29edb69b277471b3b1ae66f0e0a1536923293/src/hub.js#L558)

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -160,7 +160,7 @@ config :ret, RetWeb.Plugs.AddCSP,
   font_src: asset_hosts,
   style_src: asset_hosts,
   connect_src:
-    "https://#{host}:8080 https://sentry.prod.mozaws.net #{asset_hosts} #{websocket_hosts} https://www.mozilla.org",
+    "https://#{host}:8080 https://sentry.prod.mozaws.net #{asset_hosts} #{websocket_hosts} https://www.mozilla.org https://dev-janus.reticulum.io",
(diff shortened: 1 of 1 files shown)


config/dev.exs
```diff
@@ -153,7 +153,7 @@ asset_hosts =
 websocket_hosts =
   "https://localhost:4000 https://localhost:8080 wss://localhost:4000 " <>
     "https://#{host}:4000 https://#{host}:8080 wss://#{host}:4000 wss://#{host}:8080 wss://#{host}:8989 wss://#{host}:9090 " <>
-    "wss://#{dev_janus_host} wss://prod-janus.reticulum.io wss://#{host}:4000 wss://#{host}:8080 https://#{host}:8080 https://hubs.local:8080 wss://hubs.local:8080"
+    "wss://prod-janus.reticulum.io wss://#{host}:4000 wss://#{host}:8080 https://#{host}:8080 https://hubs.local:8080 wss://hubs.local:8080"
```
johnshaughnessy, line 156: How come `prod-janus.reticulum.io` is hard coded here? Is it safe to remove?

PR #381, opened 2020-06-16
Slackbot oauth functionality added

robin-k-wilson:


lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+
+@permissions %{
+  manage_channels: false,
+  kick_members: false,
+  view_channel: false
+}
+
+  def has_permission?(nil, _, _), do: true
```
brianpeiris, line 54: I think we want to default to false here, for anonymous users.

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  def has_permission?(%Ret.OAuthProvider{} = oauth_provider, %Ret.HubBinding{} = hub_binding, permission) do
+    oauth_provider.provider_account_id |> has_permission?(hub_binding, permission)
+  end
+
+  def has_permission?(provider_account_id, %Ret.HubBinding{} = hub_binding, permission) do
```
brianpeiris, line 60: The discord client has a guard here, checking `is_binary(provider_account_id)`. Wouldn't the same restriction apply here?

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+
+    %{@permissions |
+      manage_channels: is_member, # change scene
+      kick_members: is_member, # moderate room
+      view_channel: is_member # can access room
+    }
+  end
+
+  defp is_member_in_channel(channel_id, provider_account_id) do
+    %{"members" => members} = "#{@slack_api_base}/api/conversations.members?" <> URI.encode_query(%{token: module_config(:bot_token), channel: channel_id})
```
brianpeiris, line 78: Should these API calls be cached? Not sure what Slack's rate limits are, but we'll want to make sure we're not abusing the API with tons of requests in a short period.

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+    permissions = get_permissions(hub_binding.channel_id, provider_account_id)
+
+    permissions[permission]
+  end
+
+  defp get_permissions(channel_id, provider_account_id) do
+    # Specific channel permissions
+    is_member = is_member_in_channel(channel_id, provider_account_id)
+
+    %{@permissions |
```
brianpeiris, line 70: nit: probably don't need to use this update syntax, since you're updating all the keys and not just some of them. It also means that you don't need `@permissions` at all.

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+      |> Poison.decode!()
+
+      name
+  end
+
+  # not necessary for slack, only discord
+  # Todo remove references for slack fetching_community_identifier
+  def fetch_community_identifier(%Ret.OAuthProvider{source: _type, provider_account_id: _provider_account_id}) do
+    "1234"
+  end
```
brianpeiris, line 103: Still need some cleanup here? I don't think this is unnecessary for Slack. Slack does have a concept of "display name" and "full name". It seems right to use "full name" for the community identifier here, since it would allow users to differentiate people who happen to have the same display name.

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  def api_request(path) do
+    "#{@slack_api_base}#{path}"
+    |> Ret.HttpUtils.retry_get_until_success([{"authorization", "Bot #{module_config(:bot_token)}"}])
+    |> Map.get(:body)
+    |> Poison.decode!()
+  end
```
brianpeiris, line 110: Technically this function is not in use right now, though it ought to be used via the cache mechanism. We should probably be caching some, if not all, of the slack API requests.

lib/ret_web/channels/hub_channel.ex
```diff
@@ -830,7 +828,9 @@ defmodule RetWeb.HubChannel do
     oauth_provider =
       account.oauth_providers |> Enum.filter(fn provider -> hub_binding.type == provider.source end) |> Enum.at(0)
 
+    # when oauth_provider.type == :discord do
     assigns |> override_display_name_via_binding(oauth_provider, hub_binding)
+    # end
```
brianpeiris, line 833: Delete this commented code?

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,117 @@
...
+  @slack_api_base "https://slack.com/"
+
+  def get_oauth_url(hub_sid) do
+    authorize_params = %{
+    response_type: "code",
+    client_id: module_config(:client_id),
+    user_scope: @oauth_scope,
+    state: Ret.OAuthToken.token_for_hub(hub_sid),
+    redirect_uri: get_redirect_uri()
+    }
```
brianpeiris, line 12: Please run these changed files through `mix format` before merging.

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -1,24 +1,30 @@
...
 
-  alias Ret.{Repo, OAuthToken, OAuthProvider, DiscordClient, TwitterClient, Hub, Account, PermsToken}
+  alias Ret.{Repo, OAuthToken, OAuthProvider, DiscordClient, SlackClient, TwitterClient, Hub, Account, PermsToken}
   import Canada, only: [can?: 2]
 
   plug(RetWeb.Plugs.RateLimit when action in [:show])
 
-  def show(conn, %{"type" => "discord", "state" => state, "code" => code}) do
-    %{claims: %{"hub_sid" => hub_sid}} = OAuthToken.peek(state)
+  # handle twitter oauth to allow users hubs to tweet
```
brianpeiris, line 9: nit: Generally, I think comments like these are not particularly useful, since you can just read the code and see what it's doing. This comment might also not be true in the future, if we decide to use twitter oauth for other purposes.

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -29,7 +35,31 @@ defmodule RetWeb.Api.V1.OAuthController do
...
+  # def handle_chat_error(params, conn) do
+  #   case
+
+
+  # def handle_oauth("slack", params, conn) do
+  #   case params do
+  #     %{"error" => _} -> handle_chat_error(params)
+  #     _ -> handle_chat_oauth(params)
+  #   end
+  # end
```
brianpeiris, line 60: I would just delete all this commented code. IMO, it's more distracting when reading this code than it is helpful and the code is straightforward enough to reproduce in the future if we need it.

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -29,7 +35,31 @@ defmodule RetWeb.Api.V1.OAuthController do
     end
   end
 
-  def show(conn, %{"type" => "discord", "error" => "access_denied", "state" => state}) do
+  # handle the chat app oauth information: discord, slack etc.
+  def show(conn, %{"type" => _type} = params) do
```
brianpeiris, line 39: You could make this code more explicit and self-documenting by using a `when` guard:
```elixir
def show(conn, %{"type" => type} = params) when type in ["discord", "slack"] do
  handle_chat_oauth(params, conn)
end
```

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -121,9 +157,9 @@ defmodule RetWeb.Api.V1.OAuthController do
     end
   end
 
-  # If an oauthprovider exists for the given discord_user_id, return the associated account, updating the email
+  # If an oauthprovider exists for the given discord_user_id (chat_user_id), return the associated account, updating the email
```
brianpeiris, line 160: Might as well stick with chat_user_id, since we've removed all references to discord_user_id.
```suggestion
  # If an oauthprovider exists for the given chat_user_id, return the associated account, updating the email
```

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -85,17 +121,17 @@ defmodule RetWeb.Api.V1.OAuthController do
...
       hub
       |> Hub.perms_for_account(oauth_provider)
-      |> Map.put(:oauth_account_id, discord_user_id)
-      |> Map.put(:oauth_source, :discord)
+      |> Map.put(:oauth_account_id, chat_user_id)
+      |> Map.put(:oauth_source, source)
       |> PermsToken.token_for_perms()
 
-    conn |> put_short_lived_cookie("ret-oauth-flow-perms-token", perms_token)
+    conn |> put_short_lived_cookie("ret-oauth-flow-perms-token", perms_token) # Todo should this cookie support both discord and slack?
```
brianpeiris, line 134: I think we resolved this todo.

lib/ret_web/controllers/api/v1/bot_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+
+    # TODO check if there's any double spaces
+    args = String.split(text, " ")
+    handle_command(List.first(args), channel_id, args, team_id)
+
+    conn
+    |> send_resp(200, "")
+  end
+
+  defp handle_command("help" = _command, channel_id, _args, _team_id) do
```
brianpeiris, line 36: nit: If you didn't know, the `= _command` isn't actually necessary.
```suggestion
  defp handle_command("help", channel_id, _args, _team_id) do
```

lib/ret_web/controllers/api/v1/bot_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+    else
+      # environmentURL = if (Enum.at(args, 1)) === nil, do "", else Enum.at(args, 1)
+      # name = if (Enum.at(args, 2)) === nil, do channelName, else Enum.at(args, 2)
+      params = %{
+        name: channel_name
+        # default_environment_gltf_bundle_url: sceneurl,
+        # scene_id: sceneid
+      }
+
+      {:ok, %{hub_sid: hub_sid} = new_hub} = HubController.create_new_room(params)
```
brianpeiris, line 66: It's a bit odd for controllers to call functions in other controllers. I would move `create_new_room` to this controller, or move it into the Hub module and call it from there.

lib/ret_web/controllers/api/v1/bot_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+      update_topic(channel_id, add_hub_topic(topic, Hub.url_for(new_hub)))
+
+      RetWeb.Api.V1.HubBindingController.bind_hub(%{
```
brianpeiris, line 69: Same here, move `bind_hub` into this controller, or into hub_binding.ex. In fact, it doesn't really make sense to create this struct to hold all the properties. You could just pass them directly to the function. That would also solve the awkward `"hub_id" => hub_sid` bit.

lib/ret_web/controllers/api/v1/bot_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+  defp handle_command(_command, channel_id, _args, _team_id) do
+    send_message_to_channel(channel_id, "Type \"/hubs help\" if you need the list of available commands")
+  end
+
+  defp has_hubs_url(topic) do
+    String.match?(topic, get_hub_url_regex())
+  end
+
+  defp remove_hub_topic(topic) do
+    new_topic = Regex.replace(~r/[<>]/, topic, "")
```
brianpeiris, line 100: It's a bit odd for this regex to remove all occurrences of "<" and ">". What if the topic already happen to have those characters in it?

lib/ret_web/controllers/api/v1/bot_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+    |> Map.get(:body)
+    |> Poison.decode!()
+  end
+
+  defp send_message_to_channel(channel_id, message) do
+    ("#{@slack_api_base}/api/chat.postMessage?" <>
+       URI.encode_query(%{token: module_config(:bot_token), channel: channel_id, text: message}))
+    |> Ret.HttpUtils.retry_get_until_success()
+    |> Map.get(:body)
+    |> Poison.decode!()
```
brianpeiris, line 129: Probably don't need to do `get(:body)` and `Poison.decode!()` for this function and for update_topic, since we don't actually do anything with the response.

lib/ret_web/router.ex
```diff
@@ -98,6 +98,10 @@ defmodule RetWeb.Router do
       resources("/ret_notices", Api.V1.RetNoticeController, only: [:create])
     end
 
+    scope "/v1", as: :api_v1 do
+      resources("/slack", Api.V1.BotController, only: [:create, :index])
```
brianpeiris, line 102: Don't need the `:index` method, I think.

PR #382, opened 2020-06-22
Add rooms query API

johnshaughnessy:


lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -1,16 +1,101 @@
...
   plug(RetWeb.Plugs.HeaderAuthorization when action in [:delete])
 
+  @show_request_schema %{
+                         "type" => "object",
+                         "properties" => %{
+                           "created_by_account_email" => %{
+                             "type" => "string",
+                             "format" => "email"
+                           },
+                           "hub_sids" => %{
```
robertlong, line 23: Let's change this to `room_ids` since I think with this API we are remapping the term `hub` to `room`.

lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -1,16 +1,101 @@
...
   # Limit to 1 TPS
   plug(RetWeb.Plugs.RateLimit)
 
   # Only allow access to remove hubs with secret header
   plug(RetWeb.Plugs.HeaderAuthorization when action in [:delete])
 
+  @show_request_schema %{
+                         "type" => "object",
+                         "properties" => %{
+                           "created_by_account_email" => %{
```
robertlong, line 19: This should only be available if the user is an admin.

lib/ret_web/controllers/api/v1/hub_controller.ex
```diff
@@ -1,16 +1,101 @@
...
+      Ret.Hub
+      |> filter_by_created_by_account_email(params)
+      |> filter_by_hub_sids(params)
+      |> Ret.Repo.all()
+      |> filter_by_can_join(account)
+
+    views =
+      hubs
+      |> Enum.map(fn hub ->
+        Phoenix.View.render(RetWeb.Api.V1.HubView, "show.json", %{hub: hub}) |> Map.get(:hubs) |> hd
```
robertlong, line 93: I think we should do a new view for this route and map `hub` to `room`. As we make more APIs public we can continue to make this change.

lib/ret_web/router.ex
```diff
@@ -120,6 +120,7 @@ defmodule RetWeb.Router do
       resources("/scenes", Api.V1.SceneController, only: [:create, :update])
       resources("/avatars", Api.V1.AvatarController, only: [:create, :update, :delete])
       resources("/hubs", Api.V1.HubController, only: [:update])
+      get("/rooms", Api.V1.HubController, :show)
```
robertlong, line 123: Let's just move Rooms to it's own controller. I expect we'll run both the HubController and RoomController for a while where RoomController is the public API and HubController will be deprecated.

lib/ret_web/controllers/api/v1/hub_controller.ex
robertlong: Also we need to put a limit on the number of rooms you can query for, otherwise this will be abused.

lib/ret_web/controllers/api/v1/hub_controller.ex
johnshaughnessy: Removed for now after conversation with Liv -- we can start with an API that gives people access to the rooms that are already visible to them under normal conditions (public, created, favorited) and add more as we get our heads around authentication (via scopes / roles / etc)

lib/ret_web/controllers/api/v1/room_controller.ex
```diff
@@ -0,0 +1,153 @@
+# TODO: Should this API be re-using code in media_search.ex?
+# TODO: Resolve this question before deploy
```
johnshaughnessy, line 2: Feedback welcome

lib/ret_web/controllers/api/v1/room_controller.ex
```diff
@@ -0,0 +1,153 @@
...
+defmodule Ret.RoomSearchResult do
+  # TODO: Should :data be called :entries like Ret.MediaSearchResult and the scrivener api?
+  # TODO: Should :data be called :rooms because that's "what is being returned"
+  # TODO: Resolve this question before deploy
```
johnshaughnessy, line 6: Feedback welcome

lib/ret_web/views/api/v1/room_view.ex
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def render("show.json", %{hub: hub}) do
+    hub |> render_with_scene_asset(:gltf_bundle, hub.default_environment_gltf_bundle_url)
+  end
+
+  def render_with_scene(hub, embeddable) do
+    %{
+      # hub_id: hub.hub_sid, # renamed hub_id to room_id
+      room_id: hub.hub_sid,
```
robertlong, line 30: This doesn't need to block this PR necessarily, but we should talk about this `<type>_id` convention in API responses rather than just using `id`. IMO we should be using `id` when it is already explicit what type of record is being returned. Internally, in Reticulum code, I don't see a problem with it though. It helps me keep ids straight.

lib/ret_web/controllers/api/v1/room_controller.ex
```diff
@@ -0,0 +1,153 @@
...
+      |> filter_by_entry_mode("allow")
+      |> maybe_filter_by_only_favorites(params)
+      |> maybe_filter_by_room_ids(params)
+      |> Ret.Repo.paginate(%{page: page_number, page_size: @page_size})
+      |> result_for_page(page_number)
+
+    {:ok, {200, results}}
+  end
+
+  defp include_public_rooms(conditions) do
```
robertlong, line 76: Move these into the Hub model

johnshaughnessy:
Closed in favor of 
https://github.com/mozilla/reticulum/pull/388
https://github.com/mozilla/reticulum/pull/399

There are still some things missing before the above PR's can be deployed (https://github.com/mozilla/hubs/issues/2629#issuecomment-667444491)

PR #383, opened 2020-06-29
Reticulum slackbot V0

robin-k-wilson:
Slack bot V0 managed by reticulum.

The Slack bot:
- Handles Slack OAuth + gives permission in the hubs room depending on slack in-channel permissions
- To setup: you must have a slash command in https://api.slack.com/apps > select bot > slash command
       - /hubs with the endpoint https://[reticulum url]/api/v1/slack?token=[token_in_dev.ex_for_slack_controller]
       - OAuth menu: Bot scopes
           - channels:manage
           - channels:read
           - chat:write
           - commands
           - groups:read
           - groups:write
           - im:read
           - im:write
           - mpim:read
           - mpim:write


lib/ret/slack_client.ex
```diff
@@ -0,0 +1,128 @@
...
+    }
+  end
+
+  defp is_member_in_channel(channel_id, provider_account_id) do
+    %{"members" => members} =
+      ("#{@slack_api_base}/api/conversations.members?" <>
+         URI.encode_query(%{token: module_config(:bot_token), channel: channel_id}))
+      |> Ret.HttpUtils.retry_get_until_success()
+      |> Map.get(:body)
+      |> Poison.decode!()
```
brianpeiris, line 82: Shouldn't we be using the cache here? This will eventually get called when we refresh permissions tokens from the client, so I expect we'll be seeing a lot of calls to this for each user in a slack-bound room.

lib/ret/slack_client.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  def fetch_display_name(
+        %Ret.OAuthProvider{source: :slack, provider_account_id: provider_account_id},
+        %Ret.HubBinding{community_id: _community_id}
+      ) do
+    %{"user" => %{"name" => name}} =
+      case Cachex.fetch(
+             :slack_api,
+             "/api/users.info?" <>
+               URI.encode_query(%{token: module_config(:bot_token), user: provider_account_id})
```
brianpeiris, line 95: Since the cache is backed by `api_request`, I think passing the token here is redundant, since `api_request` passes the token in the headers. Same for the `fetch_community_identifier` function below.

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -38,24 +47,29 @@ defmodule RetWeb.Api.V1.OAuthController do
...
+  def handle_chat_oauth(params, conn) do
+    %{"type" => type, "code" => code, "state" => state} = params
 
+    %{claims: %{"hub_sid" => hub_sid}} = OAuthToken.peek(state)
     hub = Hub |> Repo.get_by(hub_sid: hub_sid)
-    account = Account |> Repo.get_by(account_id: account_id |> Integer.parse() |> elem(0))
+
+    source = String.to_atom(type)
+
+    module =
```
brianpeiris, line 58: nit: maybe rename this to `chat_client`

lib/ret_web/controllers/api/v1/oauth_controller.ex
```diff
@@ -67,13 +81,13 @@ defmodule RetWeb.Api.V1.OAuthController do
   end
 
   # Discord user has a verified email, so we create a Hubs account for them associate it with their discord user id.
```
brianpeiris, line 83:
```suggestion
  # Chat account has a verified email, so we create a Hubs account for them associate it with their chat user id.
```

lib/ret_web/controllers/api/v1/slack_controller.ex
```diff
@@ -0,0 +1,147 @@
...
+               "Rooms created with `/hubs create` will inherit moderation permissions from this Slack channel and only allow Slack users in this channel to join the room.\n" <>
+               "🦆 `/hubs create [name]` - Creates a new room with the name and puts its URL into the channel topic.\n" <>
+               "🦆 `/hubs remove` - Removes the room URL from the topic."
+
+  def create(conn, params) do
+    %{:query_params => %{"token" => reticulum_token}} = conn
+
+    if reticulum_token !== module_config(:token) do
+      conn
+      |> send_resp(404, "Missing reticulum \"token\" in query params.")
```
brianpeiris, line 19: 401 might be a better response status in this case.

lib/ret_web/controllers/api/v1/slack_controller.ex
```diff
@@ -0,0 +1,147 @@
...
+               "🦆 `/hubs help` - Shows the help text you're reading right now.\n" <>
+               "🦆 `/hubs create` - Creates a default Hubs room and puts its URL into the channel topic. " <>
+               "Rooms created with `/hubs create` will inherit moderation permissions from this Slack channel and only allow Slack users in this channel to join the room.\n" <>
+               "🦆 `/hubs create [name]` - Creates a new room with the name and puts its URL into the channel topic.\n" <>
+               "🦆 `/hubs remove` - Removes the room URL from the topic."
+
+  def create(conn, params) do
+    %{:query_params => %{"token" => reticulum_token}} = conn
+
+    if reticulum_token !== module_config(:token) do
```
brianpeiris, line 17: As a precaution, maybe also check that the configured token is not empty.

lib/ret_web/controllers/api/v1/slack_controller.ex
```diff
@@ -0,0 +1,147 @@
...
+
+      String.length(name) > 64 ->
+        send_message_to_channel(
+          channel_id,
+          "Room name is too long (over 64 characters). Please shorten room name."
+        )
+
+      true ->
+        {:ok, %{hub_sid: hub_sid} = new_hub} = Hub.create_new_room(%{"name" => name}, true)
+        update_topic(channel_id, add_hub_topic(topic, Hub.url_for(new_hub)))
```
brianpeiris, line 69: Maybe update the topic after the hub is bound.

lib/ret_web/controllers/api/v1/slack_controller.ex
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  defp has_hubs_url(topic) do
+    String.match?(topic, get_hub_url_regex())
+  end
+
+  defp remove_hub_topic(topic) do
+    # In Slack topic, hub url is surrounded by '<>' in slack ex: <https://etc>
+    # A topic set in slack with <> will not match regex. It's coded with '&gt;&lt;'
+    new_topic = Regex.replace(~r/[<>]/, topic, "")
```
brianpeiris, line 106: I think this will remove all occurrences of "<" and ">", regardless of where they are in the topic, which we probably don't want.

PR #385, opened 2020-07-01
Do not show favorite rooms that have been closed

johnshaughnessy:
Fixes https://github.com/mozilla/hubs/issues/2148

### DIFF
--- a/lib/ret/media_search.ex
+++ b/lib/ret/media_search.ex
@@ -486,17 +486,13 @@ defmodule Ret.MediaSearch do
     {:commit, results}
   end
 
-  defp rooms_must_allow_entry(results, "rooms") do
-    results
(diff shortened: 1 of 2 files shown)


lib/ret/media_search.ex
```diff
@@ -486,14 +486,25 @@ defmodule Ret.MediaSearch do
     {:commit, results}
   end
 
-  defp favorites_search(cursor, _type, account_id, _query, order \\ [desc: :last_activated_at]) do
+  defp rooms_must_allow_entry(results, "rooms") do
```
netpro2k, line 489: Was a bit confused by the need for differentiation on different types, since only rooms can be favorites, but we carry through that assumption through the rest of the flow so this makes sense (though it should be noted that the actual thing doing the rendering `favorite_to_entry` does not)

PR #391, opened 2020-07-20
Fix resolving of youtube live urls

netpro2k:
We were previously assuming all resolved youtube urls would have a querystring with expires params, this is not true for live videos. This was previously throwing an error since we tried to call `URI.decode_query` on `nil`, now handling this sort of error as a 500 instead of a 404 and actually returning the error so that we can tell the difference between internal errors and things that actually couldn't be resolved at the source.

Fixes https://github.com/mozilla/hubs/issues/2431

### DIFF
--- a/lib/ret_web/controllers/api/v1/media_controller.ex
+++ b/lib/ret_web/controllers/api/v1/media_controller.ex
@@ -90,6 +90,7 @@ defmodule RetWeb.Api.V1.MediaController do
       version: version
     }
 
+    # Status can be :commit, :ignore, or :error
     case Cachex.fetch(:media_urls, query) do
       {_status, nil} ->
         conn |> send_resp(404, "")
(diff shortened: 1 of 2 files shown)


lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -101,8 +102,8 @@ defmodule RetWeb.Api.V1.MediaController do
 
         render_resolved_media(conn, resolved_media)
 
-      _ ->
-        conn |> send_resp(404, "")
+      {:error, e} ->
```
johnshaughnessy, line 105: May still need the fallback case below this one in case none of these cases match. (The previous version was a catch-all, which we might prefer to be a 500 error with a fixed response. "Unexpected error in media resolution right at this spot")

netpro2k: I think this case block represents all the valid shapes we expect. If something unexpected does come down the pipe (it shouldn't) the case itself with throw, which seems appropriate.

PR #392, opened 2020-07-21
404 in case where youtube also fails to grab a screenshot

netpro2k:
This was hitting an error and returning 500, not a big deal but investigated to find out why, and its better explicitly handled than happening to fail.

### DIFF
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -88,7 +88,7 @@ defmodule Ret.MediaResolver do
 
           {:commit, resolved_media}
 
-        # Faild to resolve, fall through as a 404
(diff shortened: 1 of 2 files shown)


lib/ret/media_resolver.ex
```diff
@@ -70,24 +70,28 @@ defmodule Ret.MediaResolver do
...
+              # Expire a minute early
+              resolved_media |> Map.put(:ttl, ttl_s * 1000 - 60000)
+            else
+              _ -> resolved_media
+            end
 
-      {:commit, resolved_media}
+          {:commit, resolved_media}
+
+        # Faild to resolve, fall through as a 404
```
johnshaughnessy, line 91: `Faild` -> `Failed`

PR #393, opened 2020-07-22
Special handling of youtube-dl rate limit errors

netpro2k:
Returns an error image similar to what we do for offline twitch streams when youtube-dl fails because of rate limiting. This should hopefully help us separate real resolver issues vs rate limiting which we can't do much about.

Fixes https://github.com/mozilla/hubs/issues/2641


netpro2k:
Not sure why there needs to be 2 copies of the image. Something seems to generate the one in the priv directory from the one in the assets directory, but form my experience last time with the "stream offline" image it does not seem to run as part of the build process, so it needs to be checked in. We should probably try to understand that at some point, but this is just fixing a "paper cut" so not going to worry about it here.

lib/ret/media_resolver.ex
```diff
@@ -103,15 +103,26 @@ defmodule Ret.MediaResolver do
...
+          cond do
+            String.contains?(body, "is offline") ->
+              {:commit,
+               resolved(query.url, %{
+                 expected_content_type: "text/html",
+                 media_status: :offline_stream,
+                 thumbnail: RetWeb.Endpoint.static_url() <> "/stream-offline.png"
+               })}
+
+            String.contains?(body, "HTTPError 429") ->
```
johnshaughnessy, line 116: Probably unnecessary but I'd like to see an real example response from yt-dl when these errors happen (`"is offline"` or `"HTTPError 429"`). Sometime in the future their responses might change shape/form and we might more easily notice that kind of regression if we had full examples.

netpro2k: Yep, as with most stuff with yotuube-dl, this is very fragile to upstream changes because its essentially just scraping. For this one, solong as the HTTP error code doesn't change I suspect this one is pretty safe, since this string we are searching for here comes from the Python http library. The "stream offline" one less so, since I think this is text coming from twitch. 

The full text I observed on a reticulum server that was being rate limited was:
> "error":"ERROR: Unable to download webpage: HTTP Error 429: Too Many Requests (caused by <HTTPError 429: 'Too Many Requests'>); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output."}

To test this I specifically set this string to the body text, since I don't have a good way of actually getting a server into this state.

netpro2k:
> In the quota-error.png image, there is a typo, it says "vidoes" instead of "videos"!

Woops, nice catch!

PR #403, opened 2020-08-07
Add ex_doc for documentation

johnshaughnessy:
Experimental branch to explore api documentation

image

- Adds `ex_doc` dependency for generating documentation (https://github.com/elixir-lang/ex_doc#exdoc)
- Adds `@moduledoc false` to every module (since we do not have meaningful documentation in our modules). 

Issues:
- Not sure if we want to exclude `@moduledoc false` to every module, or just write guides for using the API and group them separately than all the other auto-generated module docs.
- `ex_doc` makes it easy for us to add guides / automatically generated docs, but we might want these to live somewhere else (e.g. https://github.com/MozillaReality/hubs-docs )
- No guides written for using the graphql API yet. 
- No integration with the graphql schema yet.
- Modules that are created by dependent libraries auto-generate docs. If we do not want to include them, we would have to either fork the dependencies or modify the output of the `mix docs` task. Specifically, all of the following modules are generated in some way besides us defining them directly (e.g. via a macro from a library):
```
Ret.Account.State
Ret.Asset.Type
Ret.Avatar.AvatarSlug.Type
Ret.Avatar.State
Ret.AvatarListing.AvatarListingSlug.Type
Ret.AvatarListing.State
Ret.Guardian.Plug
Ret.Hub.EntryMode
Ret.Hub.HubSlug.Type
Ret.HubBinding.Type
Ret.OAuthProvider.Source
Ret.OAuthToken.Plug
Ret.OwnedFile.State
Ret.PermsToken.Plug
Ret.Scene.SceneSlug.Type
Ret.Scene.State
Ret.SceneListing.SceneListingSlug.Type
Ret.SceneListing.State
Ret.Scheduler
Ret.SessionToken.Plug
RetWeb.Router.Helpers 
```


robertlong:
I'm not so certain we should use ex_doc for the GraphQl docs. More likely we should use an existing GraphQL API doc generator and make use of [GraphiQL Explorer](https://github.com/OneGraph/graphiql-explorer) or [GraphQL Playground](https://github.com/prisma-labs/graphql-playground) to make the API explorable for new users.

Ideally this GraphQL API doc generator just spits out Markdown that can go into our existing docs site. Perhaps as an automated build step. This project looks like it'll accomplish just that: https://github.com/edno/docusaurus2-graphql-doc-generator There could also be other options out there.

johnshaughnessy:
Added documentation directly to the graphql objects/fields (https://github.com/mozilla/reticulum/pull/399/commits/57a052f2ee6d088689ffd62a84783150dbcc6a27)

PR #404, opened 2020-08-07
Add Phoenix Live Dashboard with Absinthe metrics

johnshaughnessy:
Experimental

image

- Upgrade phoenix and other dependencies
- Add `phoenix_live_dashboard` 
- Add request logger
- Add absinthe telemetry events for monitoring graphql api performance

Issues: 
- This should be broken into two parts: Upgrade phoenix and other dependencies first, then add phoenix_live_dashboard and absinthe telemetry events.
- Need to make sure elixir/erlang versions running in production are compatible with the upgraded dependencies.


mix.exs
```diff
@@ -34,23 +34,27 @@ defmodule Ret.Mixfile do
...
       {:phoenix_html, "~> 2.13"},
+      {:phoenix_live_view, "~> 0.14.4"},
       {:phoenix_live_reload, "~> 1.2", only: :dev},
+      {:telemetry_metrics, "~> 0.4"},
+      {:telemetry_poller, "~> 0.4"},
+      {:phoenix_live_dashboard, "~> 0.2.7"},
       {:gettext, "~> 0.17"},
-      {:cowboy, "~> 2.6.3"},
-      {:plug_cowboy, "~> 2.0"},
+      {:cowboy, "~> 2.8"},
```
johnshaughnessy, line 56: If we upgrade cowboy to 2.8, then we need to make sure we are running erlang 22+

It looks like we are running 22.0 in habitat deploys:
https://github.com/mozilla/hubs-ops/blob/master/plans/erlang/habitat/plan.sh#L3

Need to confirm this and check terraform / hmc

brianpeiris:
Just noting this here for when we do end up upgrading phoenix. From a discussion in discord:

> there is an ugly bug in the toml lib in elixir that causes the server to not start if you update elixir to the latest - fix is easy just bump the toml lib

johnshaughnessy:
Closing this in favor of https://github.com/mozilla/reticulum/pull/461

PR #406, opened 2020-08-10
Basic Hub Invites

brianpeiris:
Introduces the idea of a `HubInvite`, which allows us to restrict access to rooms when they are in the `:invite` `entry_mode`. Hubs can have multiple invites, but this basic implementation only allows for one. Future enhancements might also include different types of invites, with expiration, usage limits, etc.
Currently, invites do not persist any data on a user's account. They are effectively just used as passwords in this initial implementation. Users must use the invite URL to access an invite-only room.
This PR also adds `hub_channel` handlers to fetch and revoke invites, and room join denial logic to enforce the invites.

### DIFF
--- a/lib/ret/hub_invite.ex
+++ b/lib/ret/hub_invite.ex
@@ -1,0 +1,45 @@
+defmodule Ret.HubInvite do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Ret.{Hub, HubInvite, Repo}
+
+  @schema_prefix "ret0"
+  @primary_key {:hub_invite_id, :id, autogenerate: true}
+
+  schema "hub_invites" do
+    field(:hub_invite_sid, :string)
+    field(:state, HubInvite.State)
+    belongs_to(:hub, Hub, references: :hub_id)
+    timestamps()
+  end
+
+  def invite_for_hub(%Hub{} = hub) do
+    hub_invite = HubInvite |> Repo.get_by(hub_id: hub.hub_id, state: :active)
+
+    if hub_invite == nil do
+      hub_invite_sid = Ret.Sids.generate_sid()
+
+      change(%HubInvite{hub_id: hub.hub_id, hub_invite_sid: hub_invite_sid})
+      |> unique_constraint(:hub_invite_sid)
+      |> Repo.insert!()
+    else
+      hub_invite
+    end
+  end
+
+  def revoke_invite(hub_invite_sid) when is_binary(hub_invite_sid) do
+    hub_invite = HubInvite |> Repo.get_by(hub_invite_sid: hub_invite_sid)
+    change(hub_invite, state: :revoked) |> Repo.update!()
+  end
+
+  def active?(nil = _hub_invite_sid), do: false
+
+  def active?(hub_invite_sid) do
+    case Ret.HubInvite |> Ret.Repo.get_by(hub_invite_sid: hub_invite_sid) do
+      nil -> false
+      %Ret.HubInvite{state: :revoked} -> false
+      %Ret.HubInvite{state: :active} -> true
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/ret/hub_invite.ex
```diff
@@ -0,0 +1,45 @@
...
+  @primary_key {:hub_invite_id, :id, autogenerate: true}
+
+  schema "hub_invites" do
+    field(:hub_invite_sid, :string)
+    field(:state, HubInvite.State)
+    belongs_to(:hub, Hub, references: :hub_id)
+    timestamps()
+  end
+
+  def invite_for_hub(%Hub{} = hub) do
```
netpro2k, line 17: Might not be clear from the name that this can result in an insert. Maybe `get_or_create_invite_for_hub/1`?

lib/ret_web/channels/hub_channel.ex
```diff
@@ -886,6 +921,18 @@ defmodule RetWeb.HubChannel do
        ),
        do: deny_join()
 
+  defp join_with_hub(
```
netpro2k, line 924: This pattern match is the actual thing enforcing the invite permissions right? And I assume the `account_can_update` check is so that if you have permission to update the hub you bypass the invite check altogether?

priv/repo/migrations/20200730155900_add_invite_to_hub_entry_mode_enum.exs
```diff
@@ -0,0 +1,8 @@
+defmodule Ret.Repo.Migrations.AddInviteToHubEntryModeEnum do
+  use Ecto.Migration
+  @disable_ddl_transaction true
+
+  def change do
+    Ecto.Migration.execute("ALTER TYPE ret0.hub_entry_mode ADD VALUE IF NOT EXISTS 'invite'")
```
netpro2k, line 6: Don't think you are allowed to use an `execute` in a `change` migration, since it can't automatically generate `up` and `down` for you. This seems pretty easy to reverse so I would just manually write the up/down migrations.

priv/repo/migrations/20200805190647_create_hub_invites_table.exs
```diff
@@ -0,0 +1,18 @@
...
+    create table(:hub_invites, primary_key: false) do
+      add(:hub_invite_id, :bigint, default: fragment("ret0.next_id()"), primary_key: true)
+      add(:hub_invite_sid, :string, null: false)
+      add(:hub_id, :bigint, null: false)
+      add(:state, :hub_invite_state, null: false, default: "active")
+
+      timestamps()
+    end
+
+    create(index(:hub_invites, [:hub_invite_sid], unique: true))
```
netpro2k, line 16: Should we also index by `hub_id`?

lib/ret_web/channels/hub_channel.ex
brianpeiris: Correct.

priv/repo/migrations/20200805190647_create_hub_invites_table.exs
brianpeiris: I haven't explicitly used a lookup by `hub_id`, so I generally avoid adding indices unless they're evidently needed.

priv/repo/migrations/20200730155900_add_invite_to_hub_entry_mode_enum.exs
brianpeiris: So it turns out that removing a value from a type is pretty complicated, not to mention it's not obvious what we should do with existing data in this scenario. If it were easy, I would have made a change here, but generally I would like us to understand if/why reversible migrations are important to us, because I'm not totally convinced that they are worth the trouble in our typical migration patterns.

PR #415, opened 2020-09-15
Admin only plug returns 401 correctly

robin-k-wilson:
With just put_status, the response that failed on the :admin_only step would receive a 500 response. 

This returns a 402 for unauthorized.

### DIFF
--- a/lib/ret_web/plugs/admin_only.ex
+++ b/lib/ret_web/plugs/admin_only.ex
@@ -12,8 +12,7 @@ defmodule RetWeb.Plugs.AdminOnly do
       conn
     else
       conn
-      |> resp(401, "Not authorized")
-      |> send_resp()
+      |> send_resp(401, "Not authorized")
       |> halt()
     end
   end
--- a/lib/ret_web/plugs/admin_only.ex
+++ b/lib/ret_web/plugs/admin_only.ex
@@ -7,7 +7,6 @@ defmodule RetWeb.Plugs.AdminOnly do
   def call(conn, _opts) do
     # Put account in into assigns for Canary to consume.
     account = Guardian.Plug.current_resource(conn)
-    main_host = RetWeb.Endpoint.config(:url)[:host]
 
     if account && account.is_admin do
       conn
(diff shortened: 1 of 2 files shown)


lib/ret_web/plugs/admin_only.ex
```diff
@@ -12,7 +12,8 @@ defmodule RetWeb.Plugs.AdminOnly do
       conn
     else
       conn
-      |> put_status(:unauthorized)
+      |> resp(401, "Not authorized")
+      |> send_resp()
```
johnshaughnessy, line 16: Not sure what the difference between them is, but I think you can do `|> send_resp(401, "Not authorized")` in one step.

johnshaughnessy, line 16: Not sure what the difference between them is, but I think you can do this in one line with 
`|> send_resp(401, "Not authorized")`

lib/ret_web/plugs/admin_only.ex
```diff
@@ -12,7 +12,8 @@ defmodule RetWeb.Plugs.AdminOnly do
       conn
     else
       conn
-      |> put_status(:unauthorized)
+      |> resp(401, "Not authorized")
```
johnshaughnessy, line 15: Might change the PR title/comment to say "401" instead of "402"

johnshaughnessy (approved):
:+1:

johnshaughnessy (changes requested):
Sorry, I approved this a second ago before I noticed the two failing tests related to non-admin permissions. Should fix  
https://app.circleci.com/pipelines/github/mozilla/reticulum/238/workflows/ea16659c-7e3e-4534-8304-909b571fb696/jobs/240
```
 1) test non-admins cannot search for accounts (RetWeb.AccountControllerTest)
     test/ret_web/controllers/api/account_controller_test.exs:123
     Assertion with === failed
     code:  assert conn.state() === :unset
     left:  :sent
     right: :unset
     stacktrace:
       test/ret_web/controllers/api/account_controller_test.exs:131: (test)
```
```
  2) test non-admins cannot create accounts (RetWeb.AccountControllerTest)
     test/ret_web/controllers/api/account_controller_test.exs:13
     Assertion with === failed
     code:  assert conn.state() === :unset
     left:  :sent
     right: :unset
     stacktrace:
       test/ret_web/controllers/api/account_controller_test.exs:21: (test)
```

johnshaughnessy (approved):

PR #417, opened 2020-09-18
Add revocable API tokens using guardian_db

johnshaughnessy:
Adds revocable API tokens

TODO: Configure prod.exs


priv/session_lock_repo/migrations/20200918180620_guardiandb.exs
```diff
@@ -0,0 +1,17 @@
...
+      add(:typ, :string)
+      add(:iss, :string)
+      add(:sub, :string)
+      add(:exp, :bigint)
+      add(:jwt, :text)
+      add(:claims, :map)
+      timestamps()
+    end
+  end
+end
```
johnshaughnessy, line 17: What are `session_lock_repo` migrations?

lib/ret/api_permissions.ex
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  defp scope_rooms_user do
+    [
+      :rooms_mutation_create_room,
+      :rooms_mutation_update_room,
+      :rooms_query_created_rooms,
+      :rooms_query_favorite_rooms
+    ]
+  end
```
johnshaughnessy, line 38: Is this a reasonable way to be organizing permissions into scopes?

priv/session_lock_repo/migrations/20200918180620_guardiandb.exs
brianpeiris: The SessionLockRepo is a special repo that we created to work around issues with migrations on Hubs Cloud (See #365 if you really want to dive deeper). In short, as a side effect of having SessionLockRepo, `ecto.gen.migration` creates two migration files. The one in the `priv/session_lock_repo` directory is actually redundant and should be deleted. Ideally we can find a way to prevent `ecto.gen.migration` from creating these files in the first place.

johnshaughnessy:
Closing this. I removed the guardian_db dependency and created a new API Token module in the big graphql api: https://github.com/mozilla/reticulum/pull/399/commits/9f23b27b5c3e9d3f215d4098eb8436c7e13dc60e

PR #418, opened 2020-10-08
Able to customize magic link login email message + subject from admin panel

robin-k-wilson:
Added ability to customize email subject + email message for magic link login emails.

### DIFF
--- a/habitat/config/config.toml
+++ b/habitat/config/config.toml
@@ -242,6 +242,8 @@ server = {{ toToml cfg.email.server }}
 port = {{ toToml cfg.email.port }}
 username = {{ toToml cfg.email.username }}
 password = {{ toToml cfg.email.password }}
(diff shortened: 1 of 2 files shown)


robin-k-wilson:
https://github.com/mozilla/hubs-cloud/issues/111

habitat/config/config.toml
```diff
@@ -242,6 +242,8 @@ server = {{ toToml cfg.email.server }}
 port = {{ toToml cfg.email.port }}
 username = {{ toToml cfg.email.username }}
 password = {{ toToml cfg.email.password }}
+custom_email_subject = {{ toToml cfg.email.custom_email_subject }}
```
netpro2k, line 245: I forget what we found out with Slack config stuff, but don't we need to wrap this in a conditional otherwise a blank value will cause reticulum not to start up?

PR #424, opened 2020-10-21
Add steps to setup Reticulum against a local Dialog instance to the README

keianhzo:


robin-k-wilson (approved):

johnshaughnessy:
Thanks for writing this up! I have never set up dialog (and not dialog + reticulum)

README.md
```diff
@@ -101,3 +101,22 @@ Ret.Account |> Ret.Repo.all() |> Enum.at(0) |> Ecto.Changeset.change(is_admin: t
...
+1. Update the Janus port in `dev.exs`:
+```
+config :ret, Ret.JanusLoadStatus, default_janus_host: dev_janus_host, janus_port: 4443
+```
+3. Add the Dialog meta endpoint to the CSP rules in `add_csp.ex`: 
+
+```
+default_janus_csp_rule =
+   if default_janus_host,
+      do: "wss://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port}/meta",
```
netpro2k, line 120: We actually should remove the /meta call from the client. We planned to do this after we transitioned over to dialog as it was only used to detect if the server was still on Janus

PR #426, opened 2020-11-02
OpenID Connect Login

netpro2k:
This implements the basic pieces needed for OpenID Connect login. With this and some client changes we should be able to roll out enough to at least test this as a login method on Hubs Cloud. Before rolling it out more widely we would likely need to figure out what we want to do around edge cases, like switching login providers when accounts exist, how to setup the initial admin account when using a third party provider. (this later part could also simplify the setup process for Hubs Cloud since it would get rid of the email sending requirement, but also probably means a CloudFormation template update).

Some of the layers of security here feel a bit redundant, but maybe that's ok. The session_id on a socket is already secured by a token, which guarantees the only socket getting a token back is the one that started the request, so encoding the state as a JWT is probably overkill, but we already had the mechanism in place, and it does provide an added layer including additional replay protection due to the short lived timestamp on the token. The nonce is part of the openid spec and confirms that the id token you get back is in fact for the same request that started it (since the id server will only accept this nonce once).

Currently we rely on the user manually keeping the public signing keys up to date on their instance. We eventually would want to read-through cache this.

Accompanying client PR: https://github.com/mozilla/hubs/pull/3331

If you like videos, and already understand OAuth pretty well, the last 10 minutes of this talk is a good quick primer on OpenID Connect (the rest of the talk is also a decent into to OAuth) https://youtu.be/996OiexHze0?t=2820

### DIFF
--- a/lib/ret/remote_oidc_token.ex
+++ b/lib/ret/remote_oidc_token.ex
@@ -1,7 +1,7 @@
 defmodule Ret.RemoteOIDCToken do
   @moduledoc """
   This represents an OpenID Connect token returned from a remote service.
-  The public keys will be configured by an admin for a particular setup.
+  These tokens are never created locally, only ever provided externally and verified locally.
   """
   use Guardian,
     otp_app: :ret,
@@ -12,14 +12,27 @@ defmodule Ret.RemoteOIDCToken do
 end
 
 defmodule Ret.RemoteOIDCTokenSecretsFetcher do
+  @moduledoc """
+  This represents the public keys for an OpenID Connect endpoint used to verify tokens.
+  The public keys will be configured by an admin for a particular setup. These can not be used for signing.
+  """
+
   def fetch_signing_secret(_mod, _opts) do
     {:error, :not_implemented}
   end
 
-  def fetch_verifying_secret(mod, token_headers, _opts) do
-    IO.inspect(token_headers)
+  def fetch_verifying_secret(mod, %{"kid" => kid, "typ" => "JWT"}, _opts) do
+    # TODO implement read through cache that hits discovery endpoint instead of hardcoding keys in config
+    case Application.get_env(:ret, mod)[:verification_jwks]
+         |> Poison.decode!()
+         |> Map.get("keys")
+         |> Enum.find(&(Map.get(&1, "kid") == kid)) do
+      nil -> {:error, :invalid_key_id}
+      key -> {:ok, key |> JOSE.JWK.from_map()}
+    end
+  end
 
-    # TODO use KID to look up the key. Do we want to bake it into the config at setup time? Fetch and cache? Should it be optional?
-    {:ok, Application.get_env(:ret, mod)[:verification_key] |> JOSE.JWK.from_pem()} |> IO.inspect()
+  def fetch_verifying_secret(_mod, _token_headers_, _optss) do
+    {:error, :invalid_token}
   end
 end
--- a/lib/ret/remote_oidc_token.ex
+++ b/lib/ret/remote_oidc_token.ex
@@ -22,8 +22,8 @@ defmodule Ret.RemoteOIDCTokenSecretsFetcher do
   end
 
   def fetch_verifying_secret(mod, %{"kid" => kid, "typ" => "JWT"}, _opts) do
-    # TODO implement read through cache that hits discovery endpoint instead of hardcoding keys in config
-    case Application.get_env(:ret, mod)[:verification_jwks]
+    # TODO implement read through cache that hits discovery endpoint instead of hardcoding keys in config as per https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys
+    case Application.get_env(:ret, mod)[:verification_secret_jwk_set]
          |> Poison.decode!()
          |> Map.get("keys")
          |> Enum.find(&(Map.get(&1, "kid") == kid)) do
--- a/lib/ret/remote_oidc_token.ex
+++ b/lib/ret/remote_oidc_token.ex
@@ -1,0 +1,25 @@
+defmodule Ret.RemoteOIDCToken do
+  @moduledoc """
+  This represents an OpenID Connect token returned from a remote service.
+  The public keys will be configured by an admin for a particular setup.
+  """
+  use Guardian,
+    otp_app: :ret,
+    secret_fetcher: Ret.RemoteOIDCTokenSecretsFetcher
+
+  def subject_for_token(_, _), do: {:ok, nil}
+  def resource_from_claims(_), do: {:ok, nil}
+end
+
+defmodule Ret.RemoteOIDCTokenSecretsFetcher do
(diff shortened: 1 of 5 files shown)


netpro2k:
> bit weird that remote_username is sent as "email"

Currently this was just to make the client code simpler since it matches the response for the standard login method (the verify message also ends up being the same here as well, so the only thing that changes is which auth channel to connect to. This is probably not an important property to keep, but just made prototyping it easier)

brianpeiris (commented):
Looking good so far. Would definitely like to do another review when you've got the full flow working.

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,185 @@
...
+
+    Statix.increment("ret.channels.oidc.joins.ok")
+    {:ok, %{session_id: socket.assigns.session_id}, socket}
+  end
+
+  defp module_config(key) do
+    Application.get_env(:ret, __MODULE__)[key]
+  end
+
+  defp get_redirect_uri(), do: RetWeb.Endpoint.url() <> "/verify"
```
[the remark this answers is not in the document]
netpro2k: Yes, this is the verify page on the client used for the normal email flow as well.

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,185 @@
...
+    Process.send_after(self(), :close_channel, 1000 * 5)
+
+    IO.inspect("Verify OIDC auth!!!!!")
+
+    # Slow down token guessing
+    :timer.sleep(500)
+
+    "oidc:" <> expected_topic_key = socket.topic
+
+    # TODO since we already have session_id secured by a token on the other end using a JWT for state may be overkill
```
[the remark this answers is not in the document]
netpro2k: It does, just unclear what you could do with that, since you can't chose your session id, it is assigned to you, so connecting to a different topic key would not work.

lib/ret/remote_oidc_token.ex
```diff
@@ -0,0 +1,25 @@
...
+
+defmodule Ret.RemoteOIDCTokenSecretsFetcher do
+  def fetch_signing_secret(_mod, _opts) do
+    {:error, :not_implemented}
+  end
+
+  def fetch_verifying_secret(mod, token_headers, _opts) do
+    IO.inspect(token_headers)
+
+    # TODO use KID to look up the key. Do we want to bake it into the config at setup time? Fetch and cache? Should it be optional?
```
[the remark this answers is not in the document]
netpro2k: I could not find it formally defined, but looking at how other libraries handle this they seem to either cache on an interval, or re-pulling keys if a key id that is not yet in the cache is used. My plan is to do a Cachex cache storing each key by id, with some long expiriy time (like 30 days) and then purging the cache and refetching if we see a key we don't already have. 

I got the impression this is a pretty rare event for most normal cases. If a user wanted to manually invalidate the cache they could do so by just restarting their reticulum servers, since I dont think we write Cachex caches to disk at all.

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,172 @@
...
+  intercept(["auth_credentials"])
+
+  def join("oidc:" <> _topic_key, _payload, socket) do
+    # Expire channel in 5 minutes
+    Process.send_after(self(), :channel_expired, 60 * 1000 * 5)
+
+    # Rate limit joins to reduce attack surface
+    :timer.sleep(500)
+
+    Statix.increment("ret.channels.oidc.joins.ok")
```
brianpeiris, line 18: What's the intention with this metric? Doesn't seem particularly useful on its own.

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,172 @@
...
+      )
+
+      {:reply, :ok, socket}
+    else
+      # intentionally not exposing the nature of the error, can uncomment this to return more details to the client
+      # {:error, error} ->
+      #   {:reply, {:error, %{message: error}}, socket}
+
+      _ ->
+        {:reply, {:error, %{message: "error fetching or verifying token"}}, socket}
```
brianpeiris, line 105: Maybe add a Statix error metric here?

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,172 @@
...
+    end
+  end
+
+  def handle_in(_event, _payload, socket) do
+    {:noreply, socket}
+  end
+
+  def fetch_oidc_tokens(oauth_code) do
```
brianpeiris, line 113: `defp`?

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,172 @@
...
+    GenServer.cast(self(), :close)
+    {:noreply, socket}
+  end
+
+  def handle_info(:channel_expired, socket) do
+    GenServer.cast(self(), :close)
+    {:noreply, socket}
+  end
+
+  # Only send creddentials back down to the original socket that started the request
```
brianpeiris, line 143: typo
```suggestion
  # Only send credentials back down to the original socket that started the request
```

lib/ret_web/channels/oidc_auth_channel.ex
```diff
@@ -0,0 +1,172 @@
...
+
+    {:noreply, socket}
+  end
+
+  defp broadcast_credentials_and_payload(nil, _user_info, _verification_info, _socket), do: nil
+
+  defp broadcast_credentials_and_payload(identifier_hash, user_info, verification_info, socket) do
+    account_creation_enabled = can?(nil, create_account(nil))
+    account = identifier_hash |> Account.account_for_login_identifier_hash(account_creation_enabled)
+    credentials = account |> Account.credentials_for_account()
```
brianpeiris, line 164: Took me a while to think out the implications of account creation being disabled. I guess the expectation is that users are still able to access the site/room, but are not actually "logged in" if site-wide account creation is disabled?
Might want to explain that out here in a comment.

brianpeiris (commented):
Just a few comments. Looks good, mostly.

tolen23:
May I ask if you can give a rough timeline for this ticket to be finished? It seems that there is not much left to do. Adding OpenID Connect support would be a great enhancement in my opinion.

camelgod:
Is there anything the community can do to speed this up? Let us know!!

rawnsley:
This is the single most important feature request for us as well. I would have rolled up my sleeves and implemented it myself by now if it wasn't for the fact that Elixir is so opaque and reticulum server deployments aren't easy to patch in Hubs Cloud. It really needs somebody from the core Hubs team to take pity on us :-)

Cross linking to [the original issue](https://github.com/mozilla/hubs/issues/2312).

PR #430, opened 2020-11-10
Add PosgreSQL config step for a local dialog

keianhzo:
The Coturn PSQL connection string needs to explicitly specify the `coturn` schema to be able to connect to the Reticulum DB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,3 +120,8 @@ default_janus_csp_rule =
       do: "wss://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port}/meta",
       else: ""
 ```
+
+4. Edit the Dialog ocnfiguration file *turnserver.conf* and update the PosgreSQL database connection string to use the *coturn* schema from the Reticulum database:
+```
+   psql-userdb="host=hubs.local dbname=ret_dev user=postgres password=postgres options='-c search_path=coturn' connect_timeout=30"
+```
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ default_janus_csp_rule =
       else: ""
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,3 +120,8 @@ default_janus_csp_rule =
       do: "wss://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port}/meta",
       else: ""
 ```
+
+4. Update the PosgreSQL database connection string:
```
brianpeiris, line 124: Can you elaborate on where this config needs to be changed? It looks like a coturn server config?
Also, minor typo here, it's "PostgreSQL", not "PosgreSQL".

keianhzo: This is really a Dialog related setting but as there was a  *Run Reticulum against a local Dialog instance* section here I thought it might make sense to have it here. I can move it to the dialog Readme if that makes more sense.

brianpeiris: Leaving it here is fine with me, but it would be useful to describe it more in the README and add context. Mainly we should leave instructions about where this dialog setting is so that readers know where to go to change it.

keianhzo: Right, I've updated to add more context.

README.md
```diff
@@ -120,3 +120,8 @@ default_janus_csp_rule =
       do: "wss://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port} https://#{default_janus_host}:#{janus_port}/meta",
       else: ""
 ```
+
+4. Edit the Dialog ocnfiguration file *turnserver.conf* and update the PosgreSQL database connection string to use the *coturn* schema from the Reticulum database:
```
brianpeiris, line 124:
```suggestion
4. Edit the Dialog configuration file *turnserver.conf* and update the PostgreSQL database connection string to use the *coturn* schema from the Reticulum database:
```

brianpeiris (approved):
Just fixed some typos. LGTM otherwise.

PR #436, opened 2020-12-09
Add (REST) API for managing (GraphQL) API credentials(/tokens)

johnshaughnessy:


lib/ret/api/token_utils.ex
```diff
@@ -20,4 +22,96 @@ defmodule Ret.Api.TokenUtils do
       account_id: account.account_id
     })
   end
+
+  defp ensure_atom(x) when is_atom(x), do: x
+  defp ensure_atom(x) when is_binary(x), do: String.to_atom(x)
+
+  # TODO Should we permit the creation of tokens
+  # on behalf of other accounts like this?
```
johnshaughnessy, line 30: It feels odd to allow this, but it seemed like an easy way for us to do a closed alpha/beta test of the API on HMC. Alternatively, we could add a column (e.g. an integer bitmask) to the database and use it to assign roles, then check these roles to see if the given account has permission to create credentials.

lib/ret/hub.ex
```diff
@@ -897,6 +897,57 @@ end
...
   alias Ret.{Hub, AppConfig}
+  alias Ret.Api.{Scopes, Credentials}
+
+  # Non-admin accounts can create their own account tokens
+  def can?(%Ret.Account{account_id: account_id}, :create_credentials, %{
+        subject_type: :account,
+        scopes: scopes,
+        account_id: account_id
+      }) do
+    # TODO: Is it correct to disallow the create_accounts scope for account type tokens?
```
johnshaughnessy, line 908: I got it into my head that only `:app` credentials would be allowed to have the `create_accounts` scope, but then I don't have a good reason for this. Maybe only admin accounts should be allowed to create `:account` credentials that have the `create_accounts` scope. Then the only important distinction between `:app` credentials and `:account` credentials is the fact that `:app` credentials are accessible / managed / shared by all admin accounts.

lib/ret/hub.ex
```diff
@@ -897,6 +897,57 @@ end
...
+  def can?(%Ret.Account{}, :create_credentials, _params) do
+    false
+  end
+
+  def can?(%Ret.Account{is_admin: is_admin}, :list_credentials, :app) do
+    is_admin
+  end
+
+  def can?(%Ret.Account{}, :list_credentials, :account) do
+    # TODO: Allow admins to disable this in config
```
johnshaughnessy, line 932: I don't know how/where to add config values. I'll have to learn

lib/ret_web/controllers/api/v2/credentials_controller.ex
```diff
@@ -0,0 +1,132 @@
...
+
+  alias Ret.{Repo}
+  alias Ret.Api.Credentials
+  alias Ecto.Changeset
+
+  import Ret.Api.TokenUtils,
+    only: [to_claims: 2, authed_create_credentials: 2, authed_list_credentials: 2, authed_revoke_credentials: 2]
+
+  # Limit to 1 TPS
+  # TODO : I copied this from other controllers. Is this an appropriate limit? -John
```
johnshaughnessy, line 12: Perhaps the same limit should be added to the graphql endpoint

lib/ret_web/controllers/api/v2/credentials_controller.ex
```diff
@@ -0,0 +1,132 @@
...
+  plug(RetWeb.Plugs.RateLimit when action in [:create, :update])
+
+  def index(conn, %{"app" => _anything} = _params) do
+    handle_list_credentials_result(conn, authed_list_credentials(Guardian.Plug.current_resource(conn), :app))
+  end
+
+  def index(conn, _params) do
```
johnshaughnessy, line 19: Should I add a parameter to specify an `account_id` in case an admin account wants to query for the tokens of a non-admin account?

lib/ret_web/views/api/v2/credentials_view.ex
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def render("show.json", %{token: token, credentials: credentials}) do
+    %{
+      credentials: [Map.merge(render_credentials(credentials), %{token: token})]
+    }
+  end
+
+  def render("show.json", %{credentials: credentials}) do
+    render("show.json", %{credentials: credentials, token: nil})
```
brianpeiris, line 30: Bit strange that `token` is on the outer result here, but inside `credentials:` in the other views.

lib/ret/hub.ex
```diff
@@ -897,6 +897,57 @@ end
...
+  # Non-admin accounts can create their own account tokens
+  def can?(%Ret.Account{account_id: account_id}, :create_credentials, %{
+        subject_type: :account,
+        scopes: scopes,
+        account_id: account_id
+      }) do
+    # TODO: Is it correct to disallow the create_accounts scope for account type tokens?
+    # If so, why? Do we need to enforce this everywhere?
+    Scopes.create_accounts() not in scopes
+  end
```
brianpeiris, line 911: Maybe we should only allow admins to create credentials for the first release, at least until we decide on doing roles?

test/ret_web/controllers/api/credentials_controller_test.exs
```diff
@@ -0,0 +1,362 @@
...
+    conn: conn,
+    list: list,
+    create: create
+  } do
+    assert conn
+           |> put_auth_header_for_account(account)
+           |> get(list)
+           |> json_response(200)
+           |> Map.get("credentials")
+           |> Enum.count() === 0
```
brianpeiris, line 91: We do this enough in the tests that it should be a `get_credentials_count` helper.

test/ret_web/controllers/api/credentials_controller_test.exs
```diff
@@ -0,0 +1,362 @@
...
+
+    assert conn
+           |> put_auth_header_for_account(admin_account)
+           |> patch(credentials_path(conn, :update, Map.get(creds, "id"), revoke: true))
+           |> json_response(200)
+           |> Map.get("credentials")
+           |> hd()
+           |> Map.get("is_revoked")
+  end
+end
```
brianpeiris, line 362: :+1: nice tests!

brianpeiris (commented):
Looks good, though we should definitely resolve our approach to this in general, as we discussed.

lib/ret_web/controllers/api/v2/credentials_controller.ex
johnshaughnessy: I can do this in a followup PR if we need it.

lib/ret/hub.ex
```diff
@@ -897,6 +897,57 @@ end
...
+    true
+  end
+
+  def can?(%Ret.Account{is_admin: true}, :revoke_credentials, %Credentials{}) do
+    true
+  end
+
+  def can?(%Ret.Account{}, :revoke_credentials, %Credentials{}) do
+    false
+  end
```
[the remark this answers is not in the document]
johnshaughnessy: I agree, but I don't think I can break the `Canada` `impl` for `Account`s across two files. (I tried doing this for `Atom` but didn't have success, but I might have done something wrong.) Instead I put these `Account` checks for credentials-related actions here. 

I put the `impl` for `Credentials` into its own file (`can_credentials.ex`) because I knew eventually we would have actions that did not have to do with rooms. We could do the same for accounts (e.g. `can_account.ex`). 

All the `Canada` code is sorted by `subject` due to it being a protocol (which matches on the first argument), but it might also make sense to sort functions like these by `action`. Not sure what's better.

[  def can?(subject, action, resource)](https://github.com/jarednorman/canada/blob/b21d17f7b4561aca8461614106595054d4bfd557/lib/canada/can.ex#L5)

lib/ret_web/views/api/v2/credentials_view.ex
johnshaughnessy: I may be misunderstanding your comment - but `token` always ends up as a key _inside_ the `credentials` json, not outside it. This line is just redirecting to the other `render("show.json", ...)` (which does the call to `render_credentials` and merges in the `token`). 

Open to suggestions to make this clearer -- My intention is to return a `token` in the `credentials` object when you first create one, but then have `token: null` every time after that.

test/ret_web/controllers/api/credentials_controller_test.exs
johnshaughnessy: Thanks :)

I have not wrapped my head around or studied up on best practices for testing elixir + phoenix apps. Here I test an interface (`CredentialsController`) and the model (`Credentials`) together, which I could have done this separately. I am happy enough with it as it is, but want to be better about this in the future. This seems like it'll be increasingly important as we have alternative interfaces, transports, and views (REST over HTTP, graphql, websockets passing json).

lib/ret/hub.ex
johnshaughnessy: Added in https://github.com/mozilla/reticulum/pull/436/commits/f52a5c2ddcb702b3d2aa1f1c0b4f401bc7de4009 and https://github.com/mozilla/hubs/pull/3551

lib/ret/hub.ex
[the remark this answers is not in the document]
brianpeiris: Ah, right. I think moving everything into a `can_accounts.ex` is a good idea.

lib/ret_web/views/api/v2/credentials_view.ex
brianpeiris: Ah, my mistake, I misread the code paths in this case.

PR #439, opened 2020-12-16
Enable promotion to staging channel in jenkins build

brianpeiris:
We want to be able to promote habitat builds directly to a staging channel in order to test Hubs Cloud builds. This allows a jenkins job to be configured to skip the usual channel promotion (which are only used for hubs.mozilla.com) and promote to a staging channel instead.


Jenkinsfile
```diff
@@ -37,43 +23,54 @@ pipeline {
...
+              sendSlackMessage(text, "#mr-builds", ":gift:", slackURL);
+            }
 
             // Upload to ret depot after publishing to slack to minimize wait
             sh 'sudo /usr/bin/hab-ret-pkg-upload $(ls -rt results/*.hart | head -n 1)'
+
+            if (onlyPromoteToStage == "true") {
+              sh "sudo /usr/bin/hab-ret-pkg-promote '${packageIdent}' '${stageChannel}'"
+
+              def text = "*${jobName}* promoted ${retVersion} to ${stageChannel}"
```
johnshaughnessy, line 71: It may be helpful to include some of the info / links that we do for the other case as well.
Something like
```
def text = "*${jobName}* promoted ${retVersion} to ${stageChannel}" + 
                "\nJob: *<http://localhost:8080/job/${jobName}/${buildNumber}|#${buildNumber}>* *${jobName}* " +
                "\nPackage: <https://bldr.habitat.sh/#/pkgs/${packageIdent}|${packageIdent}>\n" +
                "\nGit: <https://github.com/mozilla/reticulum/commit/$gitSha|$gitSha> " +
                "\nHead: ```${gitSha} ${gitMessage}```\n"
```

johnshaughnessy (approved):
I checked out the Jenkins job configuration for this. lgtm!

Jenkinsfile
```diff
@@ -37,43 +23,54 @@ pipeline {
...
         script {
+            def poolHost = env.RET_DARK_POOL_HOST
+            def slackURL = env.SLACK_URL
+            def buildNumber = env.BUILD_NUMBER
+            def jobName = env.JOB_NAME
+            def onlyPromoteToStage = env.ONLY_PROMOTE_TO_STAGE
+            def stageChannel = env.STAGE_CHANNEL
+
             // Grab IDENT file and cat it from .hart
             def s = $/eval 'ls -rt results/*.hart | head -n 1'/$
```
johnshaughnessy, line 34: I think this line grabs the _oldest_ `.hart` file whereas I would expect it to grab the _newest_ one. I don't think it matters because we clear `/results` in the `prebuild` step -- but might be worth fixing anyway.

```
 def s = $/eval 'ls -t results/*.hart | head -n 1'/$
 ```

PR #440, opened 2020-12-18
Remove allow_promotion from the GraphQL API

johnshaughnessy:
This removes the `allow_promotion` from the GraphQL API. This property should be handled with care. In the current implementation there were two issues:
- The flag did not undergo a permission check from the `create_room` code path
- If the token failed the permission check in the `update_room` code path, an error message was not generated. Instead, the change simply was not included in the changeset.

Instructions for adding this capability back into the API have been detailed in https://github.com/mozilla/reticulum/issues/444


brianpeiris (approved):
LGTM

PR #425, opened 2020-10-22
Bump hackney to fix TLS related issues

keianhzo:
In some platforms like Mac Catalina there seems to be some TLS related incompatibility between the hackney version  that we were using and the OS. That caused an error when accessing the admin at https://hubs.local:4000/?skipadmin:
```Missing file index.html. Please try again.```

Bumping the hackney version to 1.16.0 seems to fix that issues for Mac and WSL.

PR #447, opened 2021-01-05
added mix.lock with hackney bump

robinkwilson:
To follow up with https://github.com/mozilla/reticulum/pull/425

Added the mix.lock per PR comments.


mix.exs
```diff
@@ -53,6 +53,7 @@ defmodule Ret.Mixfile do
       {:plug_cowboy, "~> 2.0"},
       {:distillery, "~> 2.0"},
       {:peerage, "~> 1.0"},
+      {:hackney, "~> 1.16.0", override: true},
```
netpro2k, line 56: Do we actually need to specify this override in mix.exs? I had this working on my system with just the lockfile changes (resulting from just installing the latest version of the package, not hand editing it or anything)... Not super familiar with standard practice for elixir stuff but "override" feels like a scary thing.

brianpeiris: IMO, it's better to be explicit in the mix.exs file, than to be implicit in the mix.lock file, because we often don't manually read diffs of the mix.lock file.

robinkwilson:
Remade (and merged) here: https://github.com/mozilla/reticulum/pull/451

Issue #435, opened 2020-12-07
Media resolver suggestion: screenshot fallback for failed media resolution


misslivirose:
We occasionally get reports that websites with embedded videos fail to resolve properly and display as a broken media link. While we use the screenshot service if other media types aren't detected, we may want to re-work how videos and general websites "break" if there is an unresolved response. In many cases (e.g. I'm pasting a URL to a website that has an embedded video on it) it would be preferable to see a screenshot than a broken media image.

misslivirose:
Updating this to include other websites as examples where the screenshot code is failing per this issue here: https://github.com/mozilla/hubs/issues/3654

johnshaughnessy:
This PR https://github.com/mozilla/reticulum/pull/452 does not fix 100% of the issue we have been seeing. I have opened a new issue to address its shortcomings. https://github.com/mozilla/reticulum/issues/455

MegaMotion:
I am wondering if this is the same bug I am experiencing... I have a sit that has been up since last June, with many images uploaded using the magic wand tool and others hosted externally. They were all working before, but now a large number of them show broken links, even though the image is fine if I go to "open link" in hubs. Others work fine. Is there any way to reset or fix this? Will this PR do it?

visioninglab:
I am having the same problem - put URL in hub, works fine as a link but shows a broken link image. Ironically the link is https://hubs.mozilla.com/docs/hubs-features.html. Thought it might be an https problem so tried http://hubs.mozilla.com/docs/hubs-features.html. Same problem links fine but shows up as a broken link in the hub. Any tips? It's embarrassing to show a broken link even if it actually works

ConstantChange:
Any attempt to create a link with the magic wand results in a broken in link in our HC deployment.

PR #452, opened 2021-01-12
Fallback to screenshots or opengraph if possible in media resolution

johnshaughnessy:
Fix https://github.com/mozilla/reticulum/issues/435

A lot of screenshots in a hubs room

### DIFF
--- a/lib/ret/http_utils.ex
+++ b/lib/ret/http_utils.ex
@@ -24,6 +24,8 @@ defmodule Ret.HttpUtils do
   end
 
   def retry_until_success(verb, url, body \\ "", headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000) do
+    headers = headers ++ [{"User-Agent", "Mozilla/5.0 (X11; Linux x86_64; rv:84.0) Gecko/20100101 Firefox/84.0"}]
+
     hackney_options =
       if module_config(:insecure_ssl) == true do
         [:insecure]
--- a/lib/ret/http_utils.ex
+++ b/lib/ret/http_utils.ex
@@ -24,7 +24,8 @@ defmodule Ret.HttpUtils do
   end
 
   def retry_until_success(verb, url, body \\ "", headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000) do
-    headers = headers ++ [{"User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:84.0) Gecko/20100101 Firefox/84.0"}]
+    headers =
+      headers ++ [{"User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:84.0) Gecko/20100101 Firefox/84.0"}]
 
     hackney_options =
       if module_config(:insecure_ssl) == true do
--- a/lib/ret/http_utils.ex
+++ b/lib/ret/http_utils.ex
@@ -24,7 +24,7 @@ defmodule Ret.HttpUtils do
   end
 
   def retry_until_success(verb, url, body \\ "", headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000) do
-    headers = headers ++ [{"User-Agent", "Mozilla/5.0 (X11; Linux x86_64; rv:84.0) Gecko/20100101 Firefox/84.0"}]
+    headers = headers ++ [{"User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:84.0) Gecko/20100101 Firefox/84.0"}]
 
     hackney_options =
       if module_config(:insecure_ssl) == true do
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -39,8 +39,16 @@ defmodule Ret.MediaResolver do
     # TODO: We could end up running fallback_to_screenshot_opengraph_or_nothing
     #       twice in a row. These resolve functions can be simplified so that we can
     #       more easily track individual failures and only fallback when necessary.
+    #       Also make sure they have a uniform response shape for indicating an
+    #       error.
     case resolve(query, root_host) do
-      {:error, _} ->
+      :error ->
+        fallback_to_screenshot_opengraph_or_nothing(query)
+
+      {:error, _reason} ->
+        fallback_to_screenshot_opengraph_or_nothing(query)
+
+      {:commit, :error} ->
         fallback_to_screenshot_opengraph_or_nothing(query)
 
       {:commit, nil} ->
@@ -359,8 +367,7 @@ defmodule Ret.MediaResolver do
 
     case uri |> URI.to_string() |> retry_head_then_get_until_success([{"Range", "bytes=0-32768"}]) do
       :error ->
-        # TODO: Track these failures in telemetry
-        {:commit, nil}
+        :error
 
       %HTTPoison.Response{headers: headers} ->
         content_type = headers |> content_type_from_headers
@@ -415,8 +422,7 @@ defmodule Ret.MediaResolver do
   defp opengraph_result_for_uri(uri) do
     case uri |> URI.to_string() |> retry_get_until_success([{"Range", "bytes=0-32768"}]) do
       :error ->
-        # TODO: Track these failures
-        {:commit, nil}
+        :error
 
       resp ->
         # note that there exist og:image:type and og:video:type tags we could use,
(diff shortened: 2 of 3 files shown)


lib/ret/media_resolver.ex
```diff
@@ -34,7 +34,26 @@ defmodule Ret.MediaResolver do
...
     uri = url |> URI.parse()
     root_host = get_root_host(uri.host)
-    resolve(query |> Map.put(:url, uri), root_host)
+    query = Map.put(query, :url, uri)
+
+    # TODO: We could end up running fallback_to_screenshot_opengraph_or_nothing
+    #       twice in a row. These resolve functions can be simplified so that we can
+    #       more easily track individual failures and only fallback when necessary.
+    #       Also make sure they have a uniform response shape for indicating an
+    #       error.
```
johnshaughnessy, line 43: This is a tech debt concern that I do not want to cover in this PR, (and I am not sure I'm right), but it might be worth changing in the future:

I think what we might want to do is:
- Refactor `resolve/2`, `resolve_with_ytdl`, and `resolve_non_video` so that they return `{:ok, %ResolvedMedia{}} | {:error, reason}` instead of a return value for the `Cachex` `fallback` function.
- In the outer function (`resolve/1`), match on the `{:ok, %ResolvedMedia{}} | {:error, reason}` tuples (from the functions above) using `case` to decide 
  - whether to try to fallback to screenshots/opengraph,
  - whether to increment Statix metrics or other telemetry,
  - what we _actually_ want to store in the cache in the event of an error (e.g. details about what went wrong so that we can return that info to the client)
- From the outer function (`resolve/1`), return a tuple for the `Cachex` `fallback` function with a status of `:commit` or `:ignore`:
> A fallback function should return a Tuple consisting of a :commit or :ignore tag and a value. If the Tuple is tagged :commit the value will be placed into the cache and then returned. If tagged :ignore the value will be returned without being written to the cache. If you return a value which does not fit this structure, it will be assumed that you are committing the value.

From https://hexdocs.pm/cachex/Cachex.html#fetch/4

This may not be necessary, but it seems to me like it might be clearer to explicitly return one of:
  - `{:commit, resolved_media}`
  - `{:commit, {:error, reason}}`
  - ~~`{:commit, nil}`~~ (Don't use `nil` to indicate 404 -- instead use something like `{:commit, {:error, 404}}`
  - `{:ignore, some_uncached_value}`
  
  We should do our best differentiate between the return values of `Cachex.fetch(:media_urls, query)` responses. Specifically, 
  ``` elixir
  case Cachex.fetch(:media_urls, query) do
    {:error, _} -> # This is a Cachex error!
    {_status, {:error, _reason} } -> # This is a cached error we put in ourselves
  end
  ```
  See this commit for details https://github.com/mozilla/reticulum/pull/452/commits/ffab45ce844dcd6f0aac6e86a4b93c37e966c244#diff-41751f7931f9a08bf162d96c927b4c6872c21b2d215916a5b0f8041848d08152R119
  
This might be easier to distinguish if we create a `ResolvedMediaError` struct to put in the cache so that we don't mistake the difference between the two.

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -136,4 +113,59 @@ defmodule RetWeb.Api.V1.MediaController do
       :high
     end
   end
+
+  defp maybe_do_telemetry({:commit, nil}), do: Statix.increment("ret.media_resolver.404")
+  defp maybe_do_telemetry({:commit, %Ret.ResolvedMedia{}}), do: Statix.increment("ret.media_resolver.ok")
+  defp maybe_do_telemetry({:error, _reason}), do: Statix.increment("ret.media_resolver.unknown_error")
+  defp maybe_do_telemetry({:commit, :error}), do: Statix.increment("ret.media_resolver.500")
+  defp maybe_do_telemetry({:commit, {:error, _reason}}), do: Statix.increment("ret.media_resolver.500")
+  defp maybe_do_telemetry(_), do: nil
```
johnshaughnessy, line 122: I only do telemetry on the `{:commit, _}` tuples because that indicates when they are first entered into the cache. This way, we don't log 100 `404`'s if there's 1 resource that 100 people try to reference that only _really_ failed once.

lib/ret/http_utils.ex
```diff
@@ -24,6 +24,8 @@ defmodule Ret.HttpUtils do
   end
 
   def retry_until_success(verb, url, body \\ "", headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000) do
+    headers = headers ++ [{"User-Agent", "Mozilla/5.0 (X11; Linux x86_64; rv:84.0) Gecko/20100101 Firefox/84.0"}]
```
johnshaughnessy, line 27: I don't think this is correct, but it seems like we need to spoof a User-Agent for certain sites to respond to the requests we make for screenshot / opengraph resolution (the ones with the `Range` headers below https://github.com/mozilla/reticulum/pull/452/files#diff-466e850938cfeaae016678e1ab2b328f907c2b5c7620a19ded2470dfad504b54R365 )

netpro2k: Seems sane to spoof a normal browser, what specific string we use doesn't really matter much I don't think. Might want to use a Windows one in case the linux one confuses some random website.

lib/ret_web/controllers/api/v1/media_controller.ex
```diff
@@ -74,6 +74,14 @@ defmodule RetWeb.Api.V1.MediaController do
   end
 
   defp resolve_and_render(conn, url, version, quality \\ nil) do
+    query = query_for(conn, url, version, quality)
+    value = Cachex.fetch(:media_urls, query)
```
netpro2k, line 78: You could pipeline value through here (would need to change `render_resolved_media_or_error`)... Not sure if it would actually be clearer since `value` is not being transformed anyway...

This definitely reads more clearly than what we had before.

Also, we didn't have it before but might be worth adding a note that `Cachex.fetch` is what actually kicks off the resolution if its not in the cache.

lib/ret/media_resolver.ex
netpro2k: This sounds correct to me. Mixing the fallback stuff in with resolving is confusing. We should have resolving return an ok/error result like you suggest, and then use that in cachex fallback. Might be nice for the cachex part not even to live in this file.

netpro2k (approved):
lgtm

lib/ret_web/controllers/api/v1/media_controller.ex
johnshaughnessy: RE: adding a note about `Cachex.fetch`. 

I could be persuaded into adding a note (and I did not know this behavior before now either), but since it's used in several places in the client I am also OK with having people find out about this by consulting the [docs](https://hexdocs.pm/cachex/Cachex.html#fetch/4). I don't feel strongly about this though so I'll add it if you think it's helpful

```elixir
./lib/ret/slack_client.ex:78:      case Cachex.fetch(
./lib/ret/slack_client.ex:93:      case Cachex.fetch(
./lib/ret/slack_client.ex:105:      case Cachex.fetch(
./lib/ret/app_config.ex:49:        Cachex.fetch(:app_config, "")
./lib/ret/app_config.ex:117:    case Cachex.fetch(:app_config_value, key) do
./lib/ret/app_config.ex:126:    case Cachex.fetch(:app_config_owned_file_uri, key) do
./lib/ret/discord_client.ex:66:      case Cachex.fetch(:discord_api, "/guilds/#{community_id}/members/#{provider_account_id}") do
./lib/ret/discord_client.ex:71:      case Cachex.fetch(:discord_api, "/users/#{provider_account_id}") do
./lib/ret/discord_client.ex:80:    case Cachex.fetch(:discord_api, "/users/#{provider_account_id}") do
./lib/ret/discord_client.ex:138:      case Cachex.fetch(:discord_api, "/guilds/#{community_id}") do
./lib/ret/discord_client.ex:146:        case Cachex.fetch(:discord_api, "/guilds/#{community_id}/roles") do
./lib/ret/discord_client.ex:172:        case Cachex.fetch(:discord_api, "/channels/#{channel_id}") do
./lib/ret/discord_client.ex:212:      case Cachex.fetch(:discord_api, "/guilds/#{community_id}/members/#{discord_user_id}") do
./lib/ret/coturn.ex:21:    {_, coturn_secret} = Cachex.fetch(:coturn_secret, :coturn_secret)
./lib/ret_web/controllers/api/v1/media_search_controller.ex:86:    case Cachex.fetch(cache_for_query(query), query) do
./lib/ret_web/controllers/api/v1/media_controller.ex:78:    value = Cachex.fetch(:media_urls, query)
```

lib/ret/http_utils.ex
johnshaughnessy: Replaced the linux one with a Windows + Firefox one:
`Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:84.0) Gecko/20100101 Firefox/84.0`

lib/ret/media_resolver.ex
johnshaughnessy: Captured this in https://github.com/mozilla/reticulum/issues/454 for addressing at a later date

PR #458, opened 2021-01-22
mix deps.update phoenix

johnshaughnessy:

Issue #457, opened 2021-01-21
Upgrade phoenix


johnshaughnessy:
Currently we use
- [`mozillareality/erlang/22.0` (habitat.sh)](https://bldr.habitat.sh/#/pkgs/mozillareality/erlang/22.0)
- [`mozillareality/erlang/22.0` (reticulum.io)](https://bldr.reticulum.io/#/pkgs/mozillareality/erlang/22.0/20190828202629)
- [`core/elixir/1.8.0` (habitat.sh)](https://bldr.habitat.sh/#/pkgs/core/elixir/1.8.0)
- [`core/elixir/1.8.0` (reticulum.io)](https://bldr.reticulum.io/#/pkgs/core/elixir/1.8.0/20190327153604)
- [phoenix 1.4.0 (2018-11-07)](https://github.com/phoenixframework/phoenix/blob/v1.4/CHANGELOG.md#140-2018-11-07)

I want to upgrade to `phoenix`.
- Versions 1.4.1 - 1.4.17 have [a boat load of bug fixes](https://github.com/phoenixframework/phoenix/blob/v1.4/CHANGELOG.md).
- [Version 1.5.x](https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md) allows us to add [phoenix_live_dashboard](https://github.com/phoenixframework/phoenix_live_dashboard/) (for ourselves AND for hubs cloud operators.)
- Version 1.6 is currently in development and [requires upgrading elixir to 1.9+](https://github.com/phoenixframework/phoenix/blob/master/CHANGELOG.md#changelog-for-v16). There does not seem to be a compelling reason to upgrade so I recommend we stay on 1.5.x. But if we wanted to, [ elixir habitat packages ]( https://bldr.habitat.sh/#/pkgs/core/elixir/1.10.0 ) are available.

I want to upgrade `absinthe`.
- [ 1.5 ](https://github.com/absinthe-graphql/absinthe/blob/master/CHANGELOG.md#155) includes built in `telemetry` instrumentation and a bunch of bug fixes.
- [ 1.6 ](https://github.com/absinthe-graphql/absinthe/blob/master/CHANGELOG.md#160) reduces memory consumption of subscriptions and has a bunch of bug fixes.

  As far as I can tell, these upgrades will not require any changes to the CloudFormation Stack template.

johnshaughnessy:
This work is partially done in 
https://github.com/mozilla/reticulum/pull/456
and in
[...]

PR #456, opened 2021-01-20
Update absinthe to 1.5

johnshaughnessy:
Update `absinthe` to 1.5 with `mix update absinthe` and some changes to `mix.exs`. I want to do this update first before continuing development of the API (e.g. Set up telemetry) (See https://github.com/mozilla/hubs-cloud/issues/154). 

```
Upgraded:
  absinthe 1.4.16 => 1.5.5
  absinthe_phoenix 1.4.4 => 1.5.0
  absinthe_plug 1.4.7 => 1.5.3
  mime 1.4.0 => 1.5.0
  phoenix_html 2.14.1 => 2.14.3
  plug 1.10.4 => 1.11.0
  plug_crypto 1.1.2 => 1.2.0
New:
  nimble_parsec 1.1.0
```

Updates made to the `mix.exs` file are explained below:

Update `elixir` version in `mix.exs` because:
- Hubs Cloud environments are configured to use elixir 1.8.0 (https://github.com/mozilla/reticulum/blob/4630a3dce2a12a48ddd382e3a1385cdcc83d3456/habitat/plan.sh#L19)
- `absinthe` requires elixir 1.8 (https://github.com/absinthe-graphql/absinthe/#installation)

Update `absinthe` to 1.5 so that we get
- built-in `telemetry` instrumentation
- lower subscription memory usage
- bug fixes
Changelog: https://github.com/absinthe-graphql/absinthe/blob/master/CHANGELOG.md#changelog
If we want to update `absinthe` to 1.6, I think we have to update to `absinthe_phoenix` to 2.0, which requires `phoenix` 1.5+ .

Update `absinthe_plug` to 1.5 because it is required for `absinthe` 1.5+
Changelog: https://github.com/absinthe-graphql/absinthe_plug/blob/master/CHANGELOG.md

Update absinthe_phoenix to 1.5 because it is required for `absinthe` 1.5+
Changelog: https://github.com/absinthe-graphql/absinthe_phoenix/blob/master/CHANGELOG.md#v150
Phoenix 1.4 version info: https://github.com/absinthe-graphql/absinthe_phoenix/tree/v1.5
Phoenix 1.5 info: https://github.com/absinthe-graphql/absinthe_phoenix#phoenix-15

PR #460, opened 2021-01-22
Update phoenix to 1.5.x

johnshaughnessy:
<a id="org5221c56"></a>

# Upgrade Notes

<a id="orgde3a068"></a>

## Preparing to update

I wrote some of the reasons I want to update in <https://github.com/mozilla/reticulum/issues/457>

I made sure phoenix dependencies were up to date before updating anything else by running `mix deps.update phoenix`

The results are in <https://github.com/mozilla/reticulum/pull/458>

<a id="org3221c34"></a>

## Updating to phoenix 1.5

I followed the [phx-1.5-upgrade guide](<https://gist.github.com/chrismccord/e53e79ef8b34adf5d8122a47db44d22f>).

I updated the phx.new project generator. (I believe everyone will need to do this.)

I tried to bump `:phoenix`, `phoenix_pubsub`, and `plug_cowboy`.

I cannot bump `phoenix_pubsub` to 2.0.x without also updating `absinthe_phoenix`. Since I already started this in [another branch](<https://github.com/mozilla/reticulum/pull/456>), I merged that change into my primary branch and then proceeded to update `absinthe_phoenix` to 2.0.0 .

The [ `absinthe_phoenix` Changelog](<https://github.com/absinthe-graphql/absinthe_phoenix/blob/master/CHANGELOG.md>) does not seem to have any changes that we need to worry about.

The [ `phoenix_pubsub` Changelog](<https://github.com/phoenixframework/phoenix_pubsub/blob/master/CHANGELOG.md>) does not have any changes we need to handle directly. Instead, we need to make some changes to our use of the `Phoenix.PubSub` API as described in the `phoenix` 1.5 upgrade guide.

I do not know how to verify that I have migrated the PubSub configuration correctly (in prod.exs, config.exs, and application.ex) [as described in the guide](<https://gist.github.com/chrismccord/e53e79ef8b34adf5d8122a47db44d22f#pubsub-20-changes>). Help with this would be appreciated.

The [ `plug_cowboy` Changelog](<https://github.com/elixir-plug/plug_cowboy/blob/master/CHANGELOG.md>) does not have any changes we need to worry about. A benefit we will get is that telemetry events have been added via [ `cowboy_telemetry` ](<https://github.com/beam-telemetry/cowboy_telemetry>) and are enabled by default.

Updating `plug_cowboy` requires updating `cowboy` to at least 2.7. The [migration guide](<https://ninenines.eu/docs/en/cowboy/2.7/guide/migrating_from_2.6/>) does not appear to have any changes we need to worry about. Cowboy requires Erlang/OTP 20.0+, which is a requirement we satisfy (we use Erlang 22.) Cowboy&rsquo;s dependencies ( `ranch` and `cowlib` ) are used exclusively by `cowboy`, so their updates are not concerning to us. Previously, [we downgraded away from `cowboy 2.7.0` ](<https://github.com/mozilla/reticulum/pull/331>) because of [a bug with multipart uploads](<https://github.com/ninenines/cowboy/issues/1442>). However, this appears to be fixed so I will go ahead and allow it to update. Cowboy is only in our \`mix.exs\` file in the first place because we [wanted to bump it to version 2.7](<https://github.com/mozilla/reticulum/commit/0754ad9424104052ada5970d267858f4ffbe0a64>). However, it is not a direct dependency so I removed it from mix.exs.

<a id="org4ea7d48"></a>

## Things to verify

-   [x] PubSub 2.0 is working correctly
-   [x] The app.html.eex template renders correctly (with `@inner_content` )
-   [x] Removing the `view_module: 2` import from `ret_web.ex` did not break anything.
-   [x] Tests are functioning correctly after updating `RetWeb.ConnCase` <https://gist.github.com/chrismccord/e53e79ef8b34adf5d8122a47db44d22f#update-your-tests>
-   [x] There was [a concern about a bug in the toml lib that causes the server not to start if you update elixir](<https://github.com/mozilla/reticulum/pull/404#issuecomment-749071907>), which referred to [this comment in discord](<https://discord.com/channels/498741086295031808/498929977946013697/777195882923425813>). However, I do not think I am actually updating elixir here (despite the change in the mix.exs file) so I do not think this is relevant here.
-   [x] I have made the changes required by the deprecations in `phoenix 1.5`
-   [x] ~~The admin interface is returning 502 gateway errors. I don&rsquo;t know why. I thought maybe [updating `reverse_proxy_plug` ](<https://github.com/mozilla/reticulum/pull/459>) would fix the issue, but it seemed to have no effect.~~ This was only a problem with my local dev environment, and I fixed it. 

<a id="org0f7fe3e"></a>

## Related PR&rsquo;s and issues

-   We can close <https://github.com/mozilla/reticulum/issues/457> when we merge this PR
-   We can merge <https://github.com/mozilla/reticulum/pull/456> first by itself or just push this PR instead
-   We can merge <https://github.com/mozilla/reticulum/pull/458> first by itself or just push this PR instead
-   We can close <https://github.com/mozilla/reticulum/pull/404> and instead add `phoenix_live_dashboard` in a separate PR built off this branch
-   We can review <https://github.com/mozilla/reticulum/pull/459> independently and decide whether we want to merge it.


lib/ret_web.ex
```diff
@@ -34,7 +34,7 @@ defmodule RetWeb do
         namespace: RetWeb
 
       # Import convenience functions from controllers
-      import Phoenix.Controller, only: [get_flash: 2, view_module: 1]
+      import Phoenix.Controller, only: [get_flash: 2]
```
netpro2k, line 37: I see this is called out specifically to check in the TODOs but I don't see any mention of this in the upgrade notes. Whats this change do?

test/support/channel_case.ex
```diff
@@ -18,7 +18,7 @@ defmodule RetWeb.ChannelCase do
   using do
     quote do
       # Import conveniences for testing with channels
-      use Phoenix.ChannelTest
+      import Phoenix.ChannelTest
```
netpro2k, line 21: The upgrade guide says to replace `use Phoenix.ConnTest` with 
```
import Plug.Conn
import Phoenix.ConnTest
```

Do we need to import Conn as well?

netpro2k (commented):
I can help dig into specifics you run into in "things to verify" but this looks good to me. A few notes, mostly for my own understanding.

Re "I do not know how to verify that I have migrated the PubSub configuration correctly".... Not 100% sure but I think if presence is working correctly we are all set. I would imagine things would break very noticeably if there were issues.

test/support/channel_case.ex
johnshaughnessy: I don't think so -- the note for `ChannelTest` only suggests replacing `use` with `import`:

> [ChannelTest] use Phoenix.ChannelTest is deprecated in favor of import Phoenix.ChannelTest
> [ConnTest] use Phoenix.ConnTest is deprecated in favor of import Plug.Conn; import Phoenix.ConnTest
    

  [source](https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md#deprecations)

lib/ret_web.ex
johnshaughnessy: This change makes it so that `Phoenix.Controller.view_module/1` is not automatically accessible from all the views (modules with `use RetWeb, :view`). 

First I made the changes as per this line [source](https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md#deprecations):
> [Layout] Use <%= @inner_content %> instead of <%= render @view_module, @view_template, assigns %> for rendering the child layout

Then I noticed that `view_module` was never used anywhere else in the project. I figured it was safe (and maybe desirable) for us to remove it, but I don't feel strongly about it.

bryanenders:
Superseded by https://github.com/mozilla/reticulum/pull/668

PR #461, opened 2021-01-26
Feature/phoenix live dashboard

johnshaughnessy:
It appears to be safe to update ecto to 3.4.0 because the error that we were seeing has been downgraded to a warning. We should test this in production, though.
https://github.com/elixir-ecto/ecto/issues/3246#issuecomment-618548463


johnshaughnessy:
Verify: 
- [ ] Make sure that the way we are using encrypted cookies is sound from a security perspective. See https://github.com/mozilla/hubs/pull/3769
- [ ] I need to alter the CSP so that `/telemetry/metrics`, `/telemetry/ret_repo_info`, and `/telemetry/request_logger` work correctly. I temporarily removed the restrictions for `script-src` just to get things working on this branch.
- [ ] Does `RequestLogger` add any security concerns? (does it capture private request info that it shouldn't? e.g. query string parameters)
- [ ] Verify that the `ecto` update is safe. It appears to be safe to update ecto to 3.4.0 because the error that we were seeing has been downgraded to a warning. We should test this in production, though.
https://github.com/elixir-ecto/ecto/issues/3246#issuecomment-618548463
https://github.com/elixir-ecto/ecto/blob/master/CHANGELOG.md#v356-2021-01-20
https://github.com/elixir-ecto/ecto_sql/blob/master/CHANGELOG.md#v354-2020-01-20

PR #462, opened 2021-01-26
Add support for creating scenes directly from GLB

netpro2k:
This adds support for creating Scenes directly from GLB files. This requires allowing partial scene updates and exposing an endpoint to get all of your own scenes (that don't have projects associated with them). This endpoint is a one-off and expect that we would eventually want to use the GraphQL API for this sort of thing, so not a whole lot of though was given to the naming or format of the endpoint.

Related Spoke PR https://github.com/mozilla/Spoke/pull/1086

### DIFF
--- a/lib/ret/project.ex
+++ b/lib/ret/project.ex
@@ -110,21 +110,21 @@ defmodule Ret.Project do
   def changeset(
         %Project{} = project,
         account,
-        %OwnedFile{} = project_owned_file,
-        %OwnedFile{} = thumbnail_owned_file,
+        project_owned_file,
+        thumbnail_owned_file,
         params
       ) do
     project
     |> changeset(account, params)
-    |> put_change(:project_owned_file_id, project_owned_file.owned_file_id)
-    |> put_change(:thumbnail_owned_file_id, thumbnail_owned_file.owned_file_id)
+    |> maybe_put_assoc(:project_owned_file_id, project_owned_file)
+    |> maybe_put_assoc(:thumbnail_owned_file_id, thumbnail_owned_file)
   end
 
   def changeset(
         %Project{} = project,
         account,
-        %OwnedFile{} = project_owned_file,
-        %OwnedFile{} = thumbnail_owned_file,
+        project_owned_file,
+        thumbnail_owned_file,
         nil = _parent_scene,
         params
       ) do
@@ -135,8 +135,8 @@ defmodule Ret.Project do
   def changeset(
         %Project{} = project,
         account,
-        %OwnedFile{} = project_owned_file,
-        %OwnedFile{} = thumbnail_owned_file,
+        project_owned_file,
+        thumbnail_owned_file,
         %Scene{} = parent_scene,
         params
       ) do
@@ -148,8 +148,8 @@ defmodule Ret.Project do
   def changeset(
         %Project{} = project,
         account,
-        %OwnedFile{} = project_owned_file,
-        %OwnedFile{} = thumbnail_owned_file,
+        project_owned_file,
+        thumbnail_owned_file,
         %SceneListing{} = parent_scene_listing,
         params
       ) do
@@ -161,5 +161,13 @@ defmodule Ret.Project do
   defp maybe_add_project_sid_to_changeset(changeset) do
     project_sid = changeset |> get_field(:project_sid) || Ret.Sids.generate_sid()
     put_change(changeset, :project_sid, project_sid)
+  end
+
+  defp maybe_put_assoc(changeset, key, nil) do
+    changeset
+  end
+
+  defp maybe_put_assoc(changeset, key, value) do
+    changeset |> put_assoc(key, value)
   end
 end
--- a/lib/ret/project.ex
+++ b/lib/ret/project.ex
@@ -116,8 +116,8 @@ defmodule Ret.Project do
       ) do
     project
     |> changeset(account, params)
-    |> put_assoc(:project_owned_file_id, project_owned_file)
-    |> put_assoc(:thumbnail_owned_file_id, thumbnail_owned_file)
+    |> put_change(:project_owned_file_id, project_owned_file.owned_file_id)
+    |> put_change(:thumbnail_owned_file_id, thumbnail_owned_file.owned_file_id)
   end
 
   def changeset(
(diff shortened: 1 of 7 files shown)


lib/ret/project.ex
```diff
@@ -110,21 +110,21 @@ defmodule Ret.Project do
   def changeset(
         %Project{} = project,
         account,
-        %OwnedFile{} = project_owned_file,
-        %OwnedFile{} = thumbnail_owned_file,
+        project_owned_file,
```
netpro2k, line 113: Since this is now optional we don't want the typecheck here. It is still checked since we are now using `put_assoc` instead of `put_change`, which will check the type matches the association.

lib/ret/storage.ex
```diff
@@ -111,6 +111,18 @@ defmodule Ret.Storage do
     |> Enum.into(%{})
   end
 
+  # Similar to promote above, but allows for passing nil. Useful for optional upload fields
+  def promote_optional(map, %Account{} = account) when is_map(map) do
```
netpro2k, line 115: I don't love this, but it matches a common usage pattern for us, and is pretty explicit... So I think I am ok with it.

lib/ret/project.ex
```diff
@@ -116,8 +116,8 @@ defmodule Ret.Project do
       ) do
     project
     |> changeset(account, params)
-    |> put_change(:project_owned_file_id, project_owned_file.owned_file_id)
-    |> put_change(:thumbnail_owned_file_id, thumbnail_owned_file.owned_file_id)
+    |> put_assoc(:project_owned_file_id, project_owned_file)
```
netpro2k, line 119: This is unrelated but an iteration of this involved creating dummy projects and this was changed. Feels silly to revert it as its an improvement.

robertlong (commented):
This was simpler than I thought it'd be. Nice work!

lib/ret/storage.ex
```diff
@@ -111,6 +111,18 @@ defmodule Ret.Storage do
...
+    map
+    |> Enum.map(fn
+      {k, {nil, nil}} -> {k, {:ok, nil}}
+      {k, {nil, nil, nil}} -> {k, {:ok, nil}}
+      {k, {id, key}} -> {k, promote(id, key, nil, account)}
+      {k, {id, key, promotion_token}} -> {k, promote(id, key, promotion_token, account)}
```
johnshaughnessy, line 121: Unrelated to promotion semantics (which I haven't thought enough about to comment on), this and the other `promote` functions might be easier to follow if they changed some of the tuples below to maps and accepted named arguments (maps again) instead of positional arguments.

johnshaughnessy (approved):
BTW the PR summary can be updated now that you aren't creating dummy projects:
> it should be fairly easy to filter out these "dummy" projects later if we need to.

lib/ret/project.ex
```diff
@@ -116,8 +116,8 @@ defmodule Ret.Project do
...
+    |> put_assoc(:thumbnail_owned_file_id, thumbnail_owned_file)
```
brianpeiris, line 120: There are a couple of problems with this change. First, put_assoc will currently fail the second time you call it on a project, since these associations use the default `:raise` setting on the `:on_replace` option. This is currently why the CI build is failing on this PR.
Aside, from that, if you do want to go ahead with this change and set a different `:on_replace` option, you'll need to use `:project_owned_file` here instead of `:project_owned_file_id`.
See https://hexdocs.pm/ecto/Ecto.Changeset.html#module-the-on_replace-option and https://hexdocs.pm/ecto/Ecto.Schema.html#belongs_to/3 (although it's weird that we've stated a project `belongs_to` project_owned_file in the project schema)
I'd be inclined to keep using `put_change`, since it already does what we want.

lib/ret/storage.ex
brianpeiris: The code itself looks fine. The only issue I'd raise is that if there's a client bug in the future, or we open up these APIs for general use, it might be surprising if the server essentially fails silently if the client happens to pass nils when it didn't mean to. I think previously we would actually return a 404 via the `promotion_error` step in `scene_controller.ex`, so you've essentially changed the behavior of an existing API.
Not a strong opinion on this, just wanted to point it out.

brianpeiris (changes requested):

lib/ret/storage.ex
netpro2k: Yep thats right, though the DB constraints will still take care of ensuring that the required fields are always set at some point. As in, I wont be able to create a scene without a glb file for example. What does become allowed is updating a scene's thumbnail without also updating the glb for example... But that seems like a good thing. I do agree in general that this feels a bit weird, but I do think its expressing a usage pattern we seem to have in a few spots, so I think I am ok with it.

lib/ret/storage.ex
netpro2k: I do think the tuples make sense for the return value since its quite a standard pattern, but agree that the multiple levels of tuples going on here is confusing. Changing is definitely out of scope for this PR, but worth us thinking about.

lib/ret/project.ex
netpro2k: Ah good catch. Hadn't looked into the failing test yet (its usually a linting thing I just handle at the end). Since I ended up not actually needing to create projects this change was unneeded anyway, so I will just revert.

lib/ret_web/router.ex
```diff
@@ -121,10 +121,17 @@ defmodule RetWeb.Router do
       resources("/hubs", Api.V1.HubController, only: [:create, :delete])
     end
 
+    # Must be defined before :show for scenes
+    scope "/v1", as: :api_v1 do
+      pipe_through([:auth_required])
+      get("/scenes/projectless", Api.V1.SceneController, :index_projectless)
```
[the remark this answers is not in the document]
netpro2k: Though I agree its quite weird to have a specific endpoint just for projectless scenes, its sort of equally weird to me to have a scene index endpoint that only supports 1 very specific set of filters (owned by me and don't have a project). Exposing a more generic scene endpoint that supports multiple kinds of filtering feels out of scope for this, when we do it, we likely want to do it as part of the GraphQL API anyway.

brianpeiris (approved):
I still worry a little bit that we've changed the behavior of the existing scene create_or_update API, and in a way that fails silently, but as long as we're aware of that, LGTM.

PR #477, opened 2021-03-02
Extend asset ttl

johnshaughnessy:
We suspect we are repeatedly downloading the same assets over and over again. This change caches the files as long as someone is still using them. We will measure the impact in this change in grafana:

image


johnshaughnessy:
Closing this in favor of https://github.com/mozilla/reticulum/pull/480

Adding a `ttl` field to the `Cachex` cache for sketchfab assets is no good. It's at the wrong level of caching. We don't need the `Cachex` cache to store the uri of file because at best this just saves a trip to the database to get the uri from a `CachedFile` record, and at worst the stored URI will be invalid because the underlying data will have been vacuumed (see `CachedFile.vacuum` and `Storage.vacuum`).

PR #478, opened 2021-03-02
Exclude scenes with project files from projectless

netpro2k:
Scenes published before Projects were associated with Scenes, and Scenes who's Projects have been deleted should not be considered "projectless" scenes.


robertlong (approved):
Looks good 👍

PR #480, opened 2021-03-08
Increase CachedFile timeout

johnshaughnessy:
We suspect we may be repeatedly downloading the same assets over and over again as their `CachedFile` records expire. This change extends the lifetime of these records to two days instead of one. We will measure the impact in this change in grafana. This will allow us to (roughly) estimate how many new source URLs are introduced each day and how many are re-requested:
image

More work is being done to allow the lifetime of `CachedFiles` to be extended until they are no longer used. This change is meant as a quick diagnostic that will be useful in the meantime.


netpro2k (approved):
lgtm

PR #481, opened 2021-03-09
Change `CachedFile` storage and vacuum strategy

johnshaughnessy:
Before this change: 
- `CachedFile` records are stored in the database. 
- Actual file bits are stored in the filesystem as `blob` and `meta` files by the `Storage` module. 
- `CachedFile`s are `vacuum`ed by removing old records from the database. 
- The actual files are `vacuum`ed by removing old files from the filesystem.

This change:
- saves the actual file bits for new `CachedFile` records into a new directory (the `cached_file_path()` instead of the `expiring_file_path()`),
- migrates the files that are currently in the `expiring_file_path()` for a `CachedFile` to the `cached_file_path()` when those files are requested,
- adds an `accessed_at` field to `CachedFile`s which is updated up to once each day,
- changes the `vacuum` task of `CachedFile` so that the associated files are also removed from the filesystem (and only when the `CachedFile` has not been recently accessed),
- adds tests for downloading sketchfab assets and for `vacuum`ing `CachedFile`s,
- adds configuration to exclude sketchfab tests from normal test routine (because they require an API key)

CAUTION: This change contains a **migration**:
https://github.com/mozilla/reticulum/blob/4eb69ca43d0b78d5ea38db686710f2476ff656f9/priv/repo/migrations/20210304010625_add_accessed_at_to_cached_files.exs#L1-L9

### DIFF
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -22,6 +22,7 @@ defmodule Ret.MediaResolver do
   @sketchfab_rate_limit %{scale: 60_000, limit: 15}
   @poly_rate_limit %{scale: 60_000, limit: 1000}
   @max_await_for_rate_limit_s 120
+  @sketchfab_ttl_ms 1000 * 60 * 60 * 24 * 31
 
   @non_video_root_hosts [
     "sketchfab.com",
@@ -454,7 +455,7 @@ defmodule Ret.MediaResolver do
         _err -> [uri, nil]
       end
 
-    {:commit, uri |> resolved(meta)}
+    {:commit, resolved(uri, meta) |> Map.put(:ttl, @sketchfab_ttl_ms)}
   end
 
   defp resolve_sketchfab_model(model_id, api_key, version \\ 1) do
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -492,7 +492,7 @@ defmodule Ret.MediaResolver do
     loader = fn path ->
       Statix.increment("ret.media_resolver.sketchfab.requests")
 
-      case download_sketchfab_model_to_path(%{model_id: model_id, api_key: api_key, version: version, path: path}) do
+      case download_sketchfab_model_to_path(%{model_id: model_id, api_key: api_key, path: path}) do
         {:ok, metadata} ->
           Statix.increment("ret.media_resolver.sketchfab.ok")
           {:ok, metadata}
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -467,7 +467,7 @@ defmodule Ret.MediaResolver do
 
           res =
             "https://api.sketchfab.com/v3/models/#{model_id}/download"
-            |> retry_get_until_success([{"Authorization", "Token #{api_key}"}])
+            |> retry_get_until_success([{"Authorization", "Token #{api_key}"}], 15_000, 15_000)
 
           case res do
             :error ->
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -464,13 +464,13 @@ defmodule Ret.MediaResolver do
         {:error, "Failed to get sketchfab metadata"}
 
       response ->
-        zip_url =
-          response
-          |> Map.get(:body)
-          |> Poison.decode!()
-          |> Kernel.get_in(["gltf", "url"])
+        case response |> Map.get(:body) |> Poison.decode() do
+          {:ok, json} ->
+            {:ok, Kernel.get_in(json, ["gltf", "url"])}
 
-        {:ok, zip_url}
+          _ ->
+            {:error, "Failed to get sketchfab metadata"}
+        end
     end
   end
 
@@ -483,8 +483,8 @@ defmodule Ret.MediaResolver do
       {:error, error} ->
         {:error, error}
 
-      other ->
-        IO.inspect({"other", other})
+      _ ->
+        {:error, "Failed to get sketchfab url"}
     end
   end
 
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -503,11 +503,7 @@ defmodule Ret.MediaResolver do
       end
     end
 
-    cached_file_result =
-      CachedFile.fetch(%{
-        cache_key: "sketchfab-#{model_id}-#{version}",
-        loader: loader
-      })
+    cached_file_result = CachedFile.fetch("sketchfab-#{model_id}-#{version}", loader)
 
     case cached_file_result do
       {:ok, uri} -> [uri, %{expected_content_type: "model/gltf+zip"}]
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -22,7 +22,6 @@ defmodule Ret.MediaResolver do
   @sketchfab_rate_limit %{scale: 60_000, limit: 15}
   @poly_rate_limit %{scale: 60_000, limit: 1000}
   @max_await_for_rate_limit_s 120
-  @sketchfab_ttl_ms 1000 * 60 * 60 * 24 * 31
 
   @non_video_root_hosts [
     "sketchfab.com",
@@ -455,41 +454,60 @@ defmodule Ret.MediaResolver do
         _err -> [uri, nil]
       end
 
-    {:commit, resolved(uri, meta) |> Map.put(:ttl, @sketchfab_ttl_ms)}
+    {:commit, resolved(uri, meta)}
+  end
+
+  defp get_sketchfab_model_zip_url(%{model_id: model_id, api_key: api_key}) do
+    case "https://api.sketchfab.com/v3/models/#{model_id}/download"
+         |> retry_get_until_success([{"Authorization", "Token #{api_key}"}], 15_000, 15_000) do
+      :error ->
+        {:error, "Failed to get sketchfab metadata"}
+
+      response ->
+        zip_url =
+          response
+          |> Map.get(:body)
+          |> Poison.decode!()
+          |> Kernel.get_in(["gltf", "url"])
+
+        {:ok, zip_url}
+    end
+  end
+
+  def download_sketchfab_model_to_path(%{model_id: model_id, api_key: api_key, path: path}) do
+    case get_sketchfab_model_zip_url(%{model_id: model_id, api_key: api_key}) do
+      {:ok, zip_url} ->
+        Download.from(zip_url, path: path)
+        {:ok, %{content_type: "model/gltf+zip"}}
+
+      {:error, error} ->
+        {:error, error}
+
+      other ->
+        IO.inspect({"other", other})
+    end
   end
 
   defp resolve_sketchfab_model(model_id, api_key, version \\ 1) do
+    loader = fn path ->
+      Statix.increment("ret.media_resolver.sketchfab.requests")
+
+      case download_sketchfab_model_to_path(%{model_id: model_id, api_key: api_key, version: version, path: path}) do
+        {:ok, metadata} ->
+          Statix.increment("ret.media_resolver.sketchfab.ok")
+          {:ok, metadata}
+
+        {:error, _} ->
+          Statix.increment("ret.media_resolver.sketchfab.errors")
+          :error
+      end
+    end
+
     cached_file_result =
-      CachedFile.fetch(
-        "sketchfab-#{model_id}-#{version}",
-        fn path ->
-          Statix.increment("ret.media_resolver.sketchfab.requests")
-
-          res =
-            "https://api.sketchfab.com/v3/models/#{model_id}/download"
-            |> retry_get_until_success([{"Authorization", "Token #{api_key}"}], 15_000, 15_000)
-
-          case res do
-            :error ->
-              Statix.increment("ret.media_resolver.sketchfab.errors")
-
-              :error
-
-            res ->
-              Statix.increment("ret.media_resolver.sketchfab.ok")
-
-              zip_url =
-                res
-                |> Map.get(:body)
-                |> Poison.decode!()
-                |> Kernel.get_in(["gltf", "url"])
-
-              Download.from(zip_url, path: path)
-
-              {:ok, %{content_type: "model/gltf+zip"}}
-          end
-        end
-      )
+      CachedFile.fetch(%{
+        cache_key: "sketchfab-#{model_id}-#{version}",
+        loader: loader
+      })
 
     case cached_file_result do
       {:ok, uri} -> [uri, %{expected_content_type: "model/gltf+zip"}]
--- a/lib/ret/media_search.ex
+++ b/lib/ret/media_search.ex
@@ -347,7 +347,7 @@ defmodule Ret.MediaSearch do
     with api_key when is_binary(api_key) <- resolver_config(:sketchfab_api_key) do
       res =
         "https://api.sketchfab.com/v3/search?#{query}"
-        |> retry_get_until_success([{"Authorization", "Token #{api_key}"}])
+        |> retry_get_until_success([{"Authorization", "Token #{api_key}"}], 15_000, 15_000)
 
       case res do
         :error ->
--- a/lib/ret/media_search.ex
+++ b/lib/ret/media_search.ex
@@ -343,7 +343,7 @@ defmodule Ret.MediaSearch do
   def available?(:tenor), do: has_resolver_config?(:tenor_api_key)
   def available?(:twitch), do: has_resolver_config?(:twitch_client_id)
 
-  defp sketchfab_search(query) do
+  def sketchfab_search(query) do
     with api_key when is_binary(api_key) <- resolver_config(:sketchfab_api_key) do
       res =
         "https://api.sketchfab.com/v3/search?#{query}"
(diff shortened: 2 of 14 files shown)


lib/ret/storage.ex
```diff
@@ -61,19 +72,62 @@ defmodule Ret.Storage do
...
-    fetch_blob(id, key, @owned_file_path)
+    fetch_blob(id, key, owned_file_path())
+  end
+
+  def fetch(%CachedFile{} = cached_file) do
+    fetch_at(cached_file, Timex.now() |> Timex.to_naive_datetime() |> NaiveDateTime.truncate(:second))
+  end
+
+  defp maybe_bump_accessed_at(%CachedFile{accessed_at: accessed_at} = cached_file, time) do
+    one_day_ago = Timex.shift(time, days: -1)
```
brianpeiris, line 87: Presumably this time constant is related specifically to the two-day expiration we've defined in cached_file.ex? In any case, it might be good to add a comment here explaining why this happens to use one day.

lib/ret/storage.ex
```diff
@@ -61,19 +72,62 @@ defmodule Ret.Storage do
...
+
+  def fetch_at(%CachedFile{file_uuid: id, file_key: key} = cached_file, time) do
+    maybe_bump_accessed_at(cached_file, time)
+
+    case fetch_blob(id, key, cached_file_path()) do
+      {:ok, meta, stream} ->
+        {:ok, meta, stream}
+
+      {:error, _} ->
+        migrate(%{cached_file: cached_file})
```
brianpeiris, line 102: As we discussed, we should probably put a lock around this migration, or at least test to ensure that multiple simultaneous requests to the same file don't cause issues.

lib/ret/storage.ex
```diff
@@ -61,19 +72,62 @@ defmodule Ret.Storage do
...
+    with {:ok, _meta, _stream} <- fetch_blob(id, file_key, expiring_file_path()),
+         {:ok, uuid} <- Ecto.UUID.cast(id),
+         [_src_file_directory, src_meta_file_path, src_blob_file_path] <- paths_for_uuid(uuid, expiring_file_path()),
+         [dest_file_directory, dest_meta_file_path, dest_blob_file_path] <- paths_for_uuid(uuid, cached_file_path()),
+         {:ok, _} <- File.mkdir_p(dest_file_directory),
+         {:ok, _} <- File.cp(src_meta_file_path, dest_meta_file_path),
+         {:ok, _} <- File.cp(src_blob_file_path, dest_blob_file_path) do
+      {:ok, cached_file}
+    else
+      _ -> {:error, "Migration failed"}
```
brianpeiris, line 120: Looks like this migration error is never handled anywhere. Should we at least log something here?

lib/ret/storage.ex
```diff
@@ -61,19 +72,62 @@ defmodule Ret.Storage do
...
+  defp migrate(%{
+         cached_file: %CachedFile{file_uuid: id, file_key: file_key} = cached_file
+       }) do
+    with {:ok, _meta, _stream} <- fetch_blob(id, file_key, expiring_file_path()),
+         {:ok, uuid} <- Ecto.UUID.cast(id),
+         [_src_file_directory, src_meta_file_path, src_blob_file_path] <- paths_for_uuid(uuid, expiring_file_path()),
+         [dest_file_directory, dest_meta_file_path, dest_blob_file_path] <- paths_for_uuid(uuid, cached_file_path()),
+         {:ok, _} <- File.mkdir_p(dest_file_directory),
+         {:ok, _} <- File.cp(src_meta_file_path, dest_meta_file_path),
+         {:ok, _} <- File.cp(src_blob_file_path, dest_blob_file_path) do
```
brianpeiris, line 117: If I'm not mistaken, all of these file functions return `:ok`, not {:ok, _}`, so I'm pretty sure this code is just outright wrong.

lib/ret/storage.ex
```diff
@@ -61,19 +72,62 @@ defmodule Ret.Storage do
...
   end
 
   defp fetch_blob(id, key, subpath) do
     with storage_path when is_binary(storage_path) <- module_config(:storage_path),
          {:ok, uuid} <- Ecto.UUID.cast(id),
          [_file_path, meta_file_path, blob_file_path] <- paths_for_uuid(uuid, subpath),
          [{:ok, _}, {:ok, _}] <- [File.stat(meta_file_path), File.stat(blob_file_path)],
-         meta <- File.read!(meta_file_path) |> Poison.decode!(),
+         {:ok, meta_file_data} <- File.read(meta_file_path),
+         {:ok, meta} <- Poison.decode(meta_file_data),
```
brianpeiris, line 130: Previously this would raise an exception related to file read failure or json decode failure, but now it will be caught by fall-through `_ ->` case below, which means the details of the error will be hidden and error handling based on `:not_found` will be misleading.

lib/ret/storage.ex
```diff
@@ -198,52 +252,105 @@ defmodule Ret.Storage do
...
           process_blob,
           nil
         )
 
-        # Clean empty dirs
         # TODO figure out what to do about owned files -- that structure increase over time
-        for type <- [@expiring_file_path] do
+        for type <- [expiring_file_path()] do
           root_path = "#{storage_path}/#{type}"
+          clean_empty_dirs(%{root_path: root_path})
```
brianpeiris, line 265: I think this may be one example where the named params pattern just feels redundant.

lib/ret/storage.ex
```diff
@@ -198,52 +252,105 @@ defmodule Ret.Storage do
...
+            end)
+            |> Enum.reduce(%{vacuumed: [], errors: []}, fn
+              {:ok, cached_file}, %{vacuumed: vacuumed, errors: errors} ->
+                %{vacuumed: vacuumed ++ [cached_file], errors: errors}
+
+              {:error, _, cached_file}, %{vacuumed: vacuumed, errors: errors} ->
+                %{vacuumed: vacuumed, errors: errors ++ [cached_file]}
+            end)
+
+          for type <- [cached_file_path()] do
```
brianpeiris, line 310: This loop is redundant, right?

lib/ret_web.ex
```diff
@@ -55,7 +55,7 @@ defmodule RetWeb do
 
   def channel do
     quote do
-      use Phoenix.Channel
+      use Phoenix.Channel, log_join: false, log_handle_in: false
```
brianpeiris, line 58: Were these changes for debugging only?

test/ret/sketchfab_test.exs
```diff
@@ -0,0 +1,94 @@
...
+        {:ok, url}
+
+      _ ->
+        {:error, "Search failed"}
+    end
+  end
+
+  defp model_id_for(%URI{path: "/models/" <> model_id}), do: model_id
+  defp model_id_for(%URI{path: "/3d-models/" <> s}), do: s |> String.split("-") |> Enum.at(-1)
+end
```
brianpeiris, line 94: I'm glad we are testing this code, but how do you see us using these tests? Should we encourage the team to run tests locally just so that these `dev_only` ones are exercised?

brianpeiris (commented):
Mostly looks good, but I think I did find an issue with the file migration code.

johnshaughnessy (commented):
Work in progress -- not ready for re-review yet.

lib/ret/media_resolver.ex
```diff
@@ -454,41 +454,60 @@ defmodule Ret.MediaResolver do
...
 
-              {:ok, %{content_type: "model/gltf+zip"}}
-          end
-        end
-      )
+    cached_file_result =
+      CachedFile.fetch(%{
+        cache_key: "sketchfab-#{model_id}-#{version}",
+        loader: loader
+      })
```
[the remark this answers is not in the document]
brianpeiris: Personally, I don't find positional arguments confusing enough to have a burning desire to replace them everywhere in our reticulum and javascript code. It's not a strong opinion, so I'd rather not spend a disproportionate amount of time discussing it one way or the other. If we do decide to adopt the pattern for whatever reason, I think I'd like to understand the rules around when we do and don't use it (e.g. when exactly is it a performance concern?), do it all in one go, and implement some linting to prevent my muscle memory from regressing the pattern.

lib/ret/storage.ex
brianpeiris: I'd say we should either change `:not_found` to a more generic error, or revert these lines and put a `try` around whatever calls `fetch_blob`.

lib/ret/storage.ex
brianpeiris: Certainly if the language itself accommodated the pattern, I would have no reason to argue with it.

lib/ret/media_resolver.ex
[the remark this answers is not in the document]
johnshaughnessy: I don't feel strongly about this either and don't want to suggest adopting something that would call for a bunch of refactoring for little, no, or possibly negative value. I've been curious if anyone would feel similarly about this but I'll save it for another time / project.

lib/ret/storage.ex
johnshaughnessy: I updated the code after your comment. Now this should return up through `Storage.fetch`, and then be handled by the caller. 
Do you think returning 500 errors [here](https://github.com/mozilla/reticulum/pull/481/files#diff-3a3972fecb3bde5b14ba6ca1228bb1f8500e1559159db614ece8faf314830a5dR204-R205) is a sufficiently good remedy, or should I also call [`Logger.error`](https://hexdocs.pm/logger/master/Logger.html#error/2)  ?

lib/ret/media_search.ex
```diff
@@ -343,11 +343,15 @@ defmodule Ret.MediaSearch do
   def available?(:tenor), do: has_resolver_config?(:tenor_api_key)
   def available?(:twitch), do: has_resolver_config?(:twitch_client_id)
 
-  defp sketchfab_search(query) do
+  def sketchfab_search(query) do
```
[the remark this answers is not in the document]
johnshaughnessy: I'm a noobie when it comes to testing, so I appreciate the feedback. I might not have thought to test sketchfab downloads specifically, except that they happen to have regressed while I was working on this feature. (See https://github.com/mozilla/reticulum/pull/483 ). 

I'm open to suggestions on how to write better tests. If testing the sketchfab api specifically were something we wanted to do, would you recommend that I move this function into a separate module so that it would have to be public in order to be used in this file (`media_search.ex`)? (That way it wouldn't be public _only_ for the tests, but this feels like cheating... I feel like I'm missing the point.)

test/ret/cached_file_test.exs
```diff
@@ -31,4 +31,141 @@ defmodule Ret.CachedFileTest do
...
+    {:ok, _meta, _stream} = Storage.fetch(file_uuid, file_key, Storage.expiring_file_path())
+    # This line ensures the file is NOT stored in the cached_file_path
+    {:error, _} = Storage.fetch(file_uuid, file_key, Storage.cached_file_path())
+    # This line causes the file to be copied to the cached_file_path
+    {:ok, _meta, _stream} = Storage.fetch(cached_file)
+    # This line ensures the file has been copied to the cached_file_path
+    {:ok, _meta, _stream} = Storage.fetch(file_uuid, file_key, Storage.cached_file_path())
+  end
+
+  test "CachedFiles are migrated from expiring storage to cached storage only once" do
```
johnshaughnessy, line 81: Technically, this test does not ensure that the migration happens only once. However, it _does_ ensure that if 500 agents try to access the file simultaneously, they will all succeed. I used println debugging to make sure that the file was only migrated once (and that the lock was working as intended), but I don't know how to test this programmatically.

lib/ret/storage.ex
brianpeiris: I think we should almost never intentionally respond with a 500 status code. One of the other 400 statuses should be good, like [412](https://httpstatuses.com/412), or [424](https://httpstatuses.com/424)

brianpeiris (approved):
This looks good to me now, except for the question about the 500 status code. I'll respond to that comment separately, since GitHub's code review UX seems to be breaking down at this point.

test/ret/cached_file_test.exs
brianpeiris: I think that's fine. We could do more sophisticated things to ensure that the migration only happens once, but it's enough that the code is exercised and that it doesn't fail.

Is it expensive to do this 500 times? I would think having 10 or 50 agents should be sufficient. Would be nice if the test itself didn't block for more than a few seconds.

lib/ret/media_search.ex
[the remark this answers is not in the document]
brianpeiris: Yeah, that would feel like cheating. I'd say we should think about ways of testing this code without having to expose implementation-specific functions. For example, there could be an integration test that tests the public interface that we already have for sketchfab search, and ideally we would mock out the actual call to the sketchfab API, because our goal here is to test our code surrounding the sketchfab API, and not the sketchfab API itself.

test/ret/sketchfab_test.exs
brianpeiris: As mentioned in another reply, ideally we should not actually have to call out to the sketchfab API to test code in our code base that processes sketchfab API requests and responses. I think it's a net-positive that you've added tests around this code, even though they are :dev_only. You don't need to improve things in this PR, but it's something we should improve at a later date.

test/ret/cached_file_test.exs
johnshaughnessy: Happy to change it to 10 or 50 but it doesn't block for long at all -- less than a second on my machine. Here's a snippet of output that I had while testing:

```
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.301347Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.302087Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.302887Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.303970Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.304901Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.305694Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.306698Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.307423Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.308170Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.308892Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.309627Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.310488Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.311215Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.311948Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.312665Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.313565Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.314357Z]}
{"got lock: migrate_b6e139fc-f671-4785-9a89-25d93c7a73a5",
 ~U[2021-03-18 04:43:16.315083Z]}
 ```

File: test/ret/sketchfab_test.exs (hunk 9 above)
Comment (johnshaughnessy) [reply test/ret/sketchfab_test.exs:94]: Ah, ok that makes sense. I could save a response from the sketchfab API and test everything downstream from that using that mock response data.

File: lib/ret/storage.ex [hunk 13]
```diff
@@ -61,23 +72,104 @@ defmodule Ret.Storage do
...
+    if Timex.before?(accessed_at, one_day_ago) do
+      cached_file |> Ecto.Changeset.change(accessed_at: time) |> Ret.Repo.update()
+    end
+  end
+
+  defp lock_and_migrate(%CachedFile{file_uuid: id, file_key: key} = cached_file) do
+    Ret.Locking.exec_after_lock(
+      "migrate_" <> id,
+      fn ->
+        case fetch_blob(id, key, cached_file_path()) do
```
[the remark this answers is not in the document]
johnshaughnessy: Yes, that's correct. I'll add a comment to clarify this point

test/ret/cached_file_test.exs
brianpeiris: Ah, okay, as long as the test runs quickly, that's fine.

PR #485, opened 2021-03-16
Reduce log spam

johnshaughnessy:
It is difficult to do println debugging in development when so much logging is enabled. This PR disables a bunch of stuff that writes to the console during an iex session. There may be a smarter way to do this. I didn't think these logs were particularly important for debugging problems in production but if they are I'll look for alternative solutions.


johnshaughnessy:
I'm not so sure that disabling logging is a good idea because I don't know when or if we need logs for debugging purposes in production. Closing this PR.

PR #488, opened 2021-04-01
Add support for setting themes as a json string in app_configs.

johnshaughnessy:
The simplest way to support custom themes for the new UI was to allow users to paste a JSON blob describing an array of themes into the admin panel. In order for this to be parsed correctly by the client, we need to parse the JSON blob into an elixir map in the backend, then re-encode it into the json blob sent to the client.

Client changes are in this PR https://github.com/mozilla/hubs/pull/4109 

The shortcoming of this approach is that it does not do this for all future json configs. I was not sure how to achieve that. Instead, it looks specifically for themes. In order to add more json parameters to the admin panel, we will need to update this code.

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -452,26 +452,38 @@ defmodule RetWeb.PageController do
     Ret.AppConfig.get_config(!!module_config(:skip_cache)) |> generate_config("APP_CONFIG")
   end
 
-  # The expected format for theme-related app configs
-  # contains an array of themes for the UI (as of April 2021)
-  # and deprecated color properties that were set for the old UI.
+  # We receive the themes array as stringified JSON. We must decode it here and
+  # then re-encode it so that it will be successfully parsed by the client.
   #
-  # {
-  #   theme: {
-  #     themes: [],
-  #     failed_to_load: false,
-  #     deprecated_color_property_1: "#ffffff",
-  #     deprecated_color_property_n: "#000000"
-  #     some_additional_flag: false,
+  # The data going into this function looks something like this:
+  # %{
+  #   "theme" => %{
+  #     "themes" => "[
+  #       { \"id\": \"theme-1\", ...},
+  #       { \"id\": \"theme-2\", ...},
+  #       ...
+  #     ]",
+  #     "deprecated_color_property_1" => "#ffffff",
+  #     "deprecated_color_property_n" => "#000000"
   #   }
   # }
-  #
-  # The themes array must be decoded from a string and re-encoded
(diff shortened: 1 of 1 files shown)


lib/ret_web/controllers/page_controller.ex
```diff
@@ -452,8 +452,23 @@ defmodule RetWeb.PageController do
...
   end
 
+  defp escape_themes(%{"themes" => themes_string} = config) do
+    case Poison.decode(themes_string) do
+      {:ok, themes_map} ->
+        Map.put(config, "themes", themes_map)
+
+      _ ->
+        config = Map.put(config, "themes", [])
+        Map.put(config, "themes_failed_to_load", true)
```
robinkwilson, line 462: Do we have something on the front end that reads this value and throws an error?

johnshaughnessy: No, but I'll add that.

johnshaughnessy: Added https://github.com/mozilla/hubs/pull/4109/commits/515208958049fd77c616ae81e1e273473f6a4984

lib/ret_web/controllers/page_controller.ex
```diff
@@ -452,8 +452,39 @@ defmodule RetWeb.PageController do
...
+  #     themes: [],
+  #     failed_to_load: false,
+  #     deprecated_color_property_1: "#ffffff",
+  #     deprecated_color_property_n: "#000000"
+  #     some_additional_flag: false,
+  #   }
+  # }
+  #
+  # The themes array must be decoded from a string and re-encoded
+  # so that it can be successfully parsed by the client.
```
robinkwilson, line 470: Yay comment!

lib/ret_web/controllers/page_controller.ex
```diff
@@ -452,8 +452,51 @@ defmodule RetWeb.PageController do
...
+  #   }
+  # }
+  defp escape_themes(%{"theme" => %{"themes" => string} = category} = config) do
+    case Poison.decode(string || "") do
+      {:ok, array} ->
+        Map.put(config, "theme", Map.put(category, "themes", array))
+
+      _ ->
+        category = Map.put(category, "themes", [])
+        category = Map.put(category, "failed_to_load", true)
```
brianpeiris, line 489: nit: I'd suggest `Map.put(category, "error", "failed_to_load")` or `Map.put(category, "errors", ["failed_to_load"])`

brianpeiris (approved):
LGTM. In the future, we might want to figure out how to read the schema.toml file from the client repo and use that to determine which fields contain JSON more in a more generic way, but since this is internal implementation, it's fine if we hard code it for now.

PR #494, opened 2021-04-21
Properly handle join attempt to non-existent hubs

brianpeiris:
This was particularly a problem with the Discord Bot, where it was attempting to connect to hub channels with hub ids that were either entered incorrectly by the user, or for hubs that had subsequently been deleted from the database. Previously we would error out trying to access `hubs_bindings` of `nil`. This PR changes accounts for a `nil` hub and allows the code to fall through to the corresponding `join_with_hub` overload.
The Discord Bot will look for the `"not_found"` reason and stop attempting to reconnect to this missing hub id with the changes in https://github.com/MozillaReality/hubs-discord-bot/pull/117

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -66,7 +66,12 @@ defmodule RetWeb.HubChannel do
         _ -> nil
       end
 
-    hub_requires_oauth = hub.hub_bindings |> Enum.empty?() |> Kernel.not()
+    hub_requires_oauth =
+      case hub do
+        nil -> false
+        _ -> hub.hub_bindings |> Enum.empty?() |> Kernel.not()
+      end
+
     bot_access_key = Application.get_env(:ret, :bot_access_key)
     has_valid_bot_access_key = !!(bot_access_key && params["bot_access_key"] == bot_access_key)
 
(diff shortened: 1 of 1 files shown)


lib/ret_web/channels/hub_channel.ex
```diff
@@ -66,7 +66,12 @@ defmodule RetWeb.HubChannel do
         _ -> nil
       end
 
-    hub_requires_oauth = hub.hub_bindings |> Enum.empty?() |> Kernel.not()
+    hub_requires_oauth =
```
netpro2k, line 69: Feels slightly weird to be going through the rest of this function with hub being nil... It might burn us in the future. Might be better to instead match on `  defp perform_join(socket, nil, context, params) do` and just return an error right away (or immediately call `join_with_hub(nil,nil,nil,nil,nil)` if we want to keep a single codepath for error handling.

netpro2k (commented):
I might instead capture the nil at an earlier level just to prevent accidentally introducing issues later, but what you have right now looks like it should fix the issue at hand.

lib/ret_web/channels/hub_channel.ex
brianpeiris: Hmm, yeah, good point. I'll try a path that exits faster.

netpro2k (approved):
lgtm

PR #499, opened 2021-05-21
Add amplify_audio permission

netpro2k:
This is for usage with "zone audio source" components and any other "presenter" audio features we might add. Still not sure about the name. Originally had `broadcast_audio` but that seems like it might be confused with sending any audio at all. Also considered `present_audio` but that also felt a bit weird.. so I landed on `amplify_audio`. Open to suggestions.

For starters just mirroring the same permission model as `mute_user`. Creators and owners are allowed to do it in unbound rooms, and in bound rooms its tied to `HubBinding.can_moderate_users` which is slightly weird, but we don't really have finer granularity right now.

Don't have an easy way to test room binding locally so I will just test that part in staging. Since I am using the same exact semantics as `mute_user` I think it should work fine.

This is all quite easy to tweak in the future.

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -878,6 +878,7 @@ defmodule Ret.Hub do
       embed_hub: account |> can?(embed_hub(hub)),
       kick_users: account |> can?(kick_users(hub)),
       mute_users: account |> can?(mute_users(hub)),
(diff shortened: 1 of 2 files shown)


johnshaughnessy (approved):

brianpeiris (approved):
LGTM

PR #507, opened 2021-06-22
Let clients tell the ret channel they changed hub

johnshaughnessy:
This PR adds a `"change_hub"` event handler to the `"ret"` channel, which we use in https://github.com/mozilla/hubs/pull/4347/commits/6f0506fdcd29a655bbbc98cd76e66ae276cb2dcb (https://github.com/mozilla/hubs/pull/4347)


netpro2k (approved):
lgtm

PR #525, opened 2021-09-27
Fix dev storage host configuration

### LINKED ISSUE #522: Reticulum fails to load assets from local scenes
### STRs:
- Start Reticulum
- Start Spoke in local mode so it runs against the local reticulum instance: `npm run run-local-reticulum`
- Start Hubs in local mode: `npm run local`
- Open Spoke: `https://hubs.local:4000/spoke`
- Create a scene and publish
- Open the scene

### Expected result
The scene opens

### Actual result
The scene fails to load

Context
Reticulum:
Screen Shot 2021-09-08 at 17 55 28

Hubs:
Screen Shot 2021-09-08 at 17 53 59

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/HUBS-899)

brianpeiris:
We introduced a storage host configuration in #508 but neglected to add it to the dev config.

Fixes #522

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -142,6 +142,7 @@ config :ret, Ret.MediaResolver,
 config :ret, Ret.Speelycaptor, speelycaptor_endpoint: "https://1dhaogh2hd.execute-api.us-west-1.amazonaws.com/public"
 
 config :ret, Ret.Storage,
+  host: "https://#{host}:4000",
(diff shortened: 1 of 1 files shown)


keianhzo:
Thanks @brianpeiris this works good.

Fixes #522

PR #531, opened 2021-11-16
Fix Discord integration

brianpeiris:
In #452 we added a browser user-agent header to all requests made through all the functions in our HttpUtils module. That change is now breaking Discord integration, since it seems Discord now rejects API requests that include a user-agent header.

This PR removes the user-agent header from all requests by default, unless it is enabled with the new `append_browser_ua` parameter. Incidentally, the functions in HttpUtils have also been changed to use a keyword-list for their optional parameters, to avoid having to worry about positional arguments.

As far as I can tell, only the screenshot and opengraph code paths require a user-agent header enabled, so I've had all the other calls omit the header by default.

### DIFF
--- a/lib/ret/http_utils.ex
+++ b/lib/ret/http_utils.ex
@@ -2,16 +2,16 @@ defmodule Ret.HttpUtils do
   use Retry
 
   def retry_head_until_success(url, options \\ []),
-    do: retry_until_success(:head, url, options)
+    do: retry_until_success(:head, url, "", options)
 
   def retry_get_until_success(url, options \\ []),
-    do: retry_until_success(:get, url, options)
+    do: retry_until_success(:get, url, "", options)
 
   def retry_post_until_success(url, body, options \\ []),
-    do: retry_until_success(:post, url, Keyword.merge(options, body: body))
+    do: retry_until_success(:post, url, body, options)
 
   def retry_put_until_success(url, body, options \\ []),
-    do: retry_until_success(:put, url, Keyword.merge(options, body: body))
+    do: retry_until_success(:put, url, body, options)
 
   def retry_head_then_get_until_success(url, options \\ []) do
     case url |> retry_head_until_success(options) do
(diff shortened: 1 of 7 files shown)


lib/ret/http_utils.ex
```diff
@@ -1,31 +1,46 @@
...
 
-  def retry_post_until_success(url, body, headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000),
-    do: retry_until_success(:post, url, body, headers, cap_ms, expiry_ms)
+  def retry_post_until_success(url, body, options \\ []),
+    do: retry_until_success(:post, url, Keyword.merge(options, body: body))
 
-  def retry_put_until_success(url, body, headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000),
-    do: retry_until_success(:put, url, body, headers, cap_ms, expiry_ms)
+  def retry_put_until_success(url, body, options \\ []),
+    do: retry_until_success(:put, url, Keyword.merge(options, body: body))
```
johnshaughnessy, line 14: These wrappers are convenient in cases where we like to write `url |> retry_foo_until_success(options)` Is it also convenient for `body` to be a top level argument here instead of requiring the call site to merge it into `options`?

lib/ret/http_utils.ex
```diff
@@ -1,31 +1,46 @@
...
   end
 
-  def retry_until_success(verb, url, body \\ "", headers \\ [], cap_ms \\ 5_000, expiry_ms \\ 10_000) do
+  defp retry_until_success(verb, url, options) do
+    default_options = [
+      body: "",
+      headers: [],
+      cap_ms: 5_000,
+      expiry_ms: 10_000,
+      append_browser_ua: false
```
johnshaughnessy, line 32: nit:  `ua` may be a difficult acronym to guess without looking at where it's used. Might be good to type it out. But not sure.

johnshaughnessy (approved):
Looks good to me. I did not run this code to verify that there aren't new regressions where some of the other requests needed the user agent header.

lib/ret/http_utils.ex
brianpeiris: Ah, yeah, that would be a bit cleaner. Good call

PR #532, opened 2021-11-16
Add failing test for discord integration bug

brianpeiris:
This is a failing test for #531. I've put this in its own PR since I'm a bit unsure of the testing technique I've used here, since it introduces and uses a mocking library in order to test our usage of the external HTTPoison library, in the context of Discord API calls. The mocking mechanism also requires some fairly complex setup, since it needs to operate across Erlang process boundaries. I'm also unsure if we do want to write a test around this particular bug, since this code is unlikely to break in this particular way.

### DIFF
--- a/test/ret/discord_client_test.exs
+++ b/test/ret/discord_client_test.exs
@@ -25,6 +25,8 @@ defmodule Ret.DiscordClientTest do
     channel_id: @restricted_channel_id
   }
 
+  @discord_api_base "https://discordapp.com/api/v6"
+
   setup_all do
     Cachex.put(:discord_api, "/guilds/#{@community_id}", %{"owner_id" => @owner_user_id})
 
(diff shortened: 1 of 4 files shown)


test/ret/discord_client_test.exs
```diff
@@ -65,4 +67,44 @@ defmodule Ret.DiscordClientTest do
...
+        {mock_asserts,
+         [
+           http_verb: verb,
+           was_discord_api_request: was_discord_api_request,
+           discord_request_has_ua_header: discord_request_has_ua_header
+         ]}
+      )
+    end)
+
+    DiscordClient.has_permission?("test", %Ret.HubBinding{community_id: "test", channel_id: "test"}, :view_channel)
```
johnshaughnessy, line 98: Might not be an issue if we flush `Cachex` caches between tests, but just want to flag that the downstream request is behind a `Cachex.fetch`.

test/ret/discord_client_test.exs
```diff
@@ -65,4 +67,44 @@ defmodule Ret.DiscordClientTest do
   test "moderator should be able to view restricted channel" do
     assert @moderator_user_id |> DiscordClient.has_permission?(@restricted_channel_binding, :view_channel)
   end
+
+  @tag only: true
+  test "should make a properly structured http request" do
```
johnshaughnessy, line 72: May want an additional test for a request that _should_ add the user agent, if we have a specific example.

johnshaughnessy (commented):
`mox` seems to be a useful testing library.

I agree though that it's not clear whether this is an important test case.

brianpeiris:
I'll close this PR since #538 will introduce Mox with a test that is more useful.

PR #533, opened 2021-11-19
Scrub room name when it is closed

johnshaughnessy:
This PR makes it so that when a room is closed, we scrub the room name so that user-submitted data does not appear in the page title.


johnshaughnessy:
Another way we could fix this problem is instead of changing the room name in the database, don't expose the room name in the page if the room's entry state is `:deny`. That would also fix the problem for rooms that have already been closed without requiring a database migration.

johnshaughnessy:
Closing this in favor of https://github.com/mozilla/reticulum/pull/535

PR #535, opened 2021-11-24
Hide room name if the room is closed

johnshaughnessy:


brianpeiris (approved):
LGTM except for the format lint error

lib/ret_web/controllers/page_controller.ex
```diff
@@ -792,4 +792,19 @@ defmodule RetWeb.PageController do
   defp with_script_tags(script), do: "<script>#{script}</script>"
 
   defp module_config(key), do: Application.get_env(:ret, __MODULE__)[key]
+
+  defp maybe_scrub_room_data(hub) do
+    if hub.entry_mode == :deny do
+      Map.merge(hub, %{
```
netpro2k, line 798: I might just return a new struct here instead, we could easily add new fields and then forget to "scrub" them later.

lib/ret_web/controllers/page_controller.ex
```diff
@@ -792,4 +792,19 @@ defmodule RetWeb.PageController do
   defp with_script_tags(script), do: "<script>#{script}</script>"
 
   defp module_config(key), do: Application.get_env(:ret, __MODULE__)[key]
+
+  defp maybe_scrub_room_data(hub) do
+    if hub.entry_mode == :deny do
```
netpro2k, line 797: Probably more idiomatic, not sure if better.. But you could pattern match on the entry_mode instead of using a conditional. As in:
```elixir
defp maybe_scrub_room_data(%Hub{entry_mode: :deny}) do
```

lib/ret_web/controllers/page_controller.ex
```diff
@@ -435,7 +435,7 @@ defmodule RetWeb.PageController do
 
     hub_meta_tags =
       Phoenix.View.render_to_string(RetWeb.PageView, "hub-meta.html",
-        hub: hub,
+        hub: hub |> maybe_scrub_room_data(),
         scene: hub.scene,
```
netpro2k, line 439: If we were trying to "scrub" the scene we are still returning it here.

netpro2k (commented):
Made some suggestions. I wonder if we should instead just be returning an error outright.. I guess the main reason we don't is we want the client to render the "room closed" page.

PR #538, opened 2021-12-06
Fix asset import helper

brianpeiris:
A regression introduced in #531 broke the asset import feature in the admin panel. The function signature for `retry_head_then_get_until_success` was changed, but the call site in `fetch_content_type` was not changed accordingly.

This PR fixes `fetch_content_type` and also adds a test that uses the Mox library to exercise the function with a basic test.

### DIFF
--- a/lib/ret/http_utils.ex
+++ b/lib/ret/http_utils.ex
@@ -49,7 +49,9 @@ defmodule Ret.HttpUtils do
       end
 
     retry with: exponential_backoff() |> randomize |> cap(options[:cap_ms]) |> expiry(options[:expiry_ms]) do
-      case HTTPoison.request(verb, url, body, headers,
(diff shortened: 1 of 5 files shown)


tanfarming:
yep i tested with ret:38 too the import issue's fixed

lib/ret/http_utils.ex
```diff
@@ -92,7 +94,7 @@ defmodule Ret.HttpUtils do
   end
 
   def fetch_content_type(url) do
-    case url |> retry_head_then_get_until_success([{"Range", "bytes=0-32768"}]) do
+    case url |> retry_head_then_get_until_success(headers: [{"Range", "bytes=0-32768"}]) do
```
johnshaughnessy, line 97: :+1:

johnshaughnessy (approved):
lgtm

PR #542, opened 2022-01-06
Respond with an error if twitter might be misconfigured

brianpeiris:


lib/ret/twitter_client.ex
```diff
@@ -129,21 +142,27 @@ defmodule Ret.TwitterClient do
...
-      |> Map.get(:body)
-      |> to_string
 
-    case response_type do
-      :urlencoded -> body |> URI.decode_query()
-      :json -> body |> Poison.decode!()
-      _ -> body
+    case result do
+      :error ->
+        {:error, :twitter_probably_misconfigured}
```
brianpeiris, line 156: Should we use better token here, instead of `:twitter_probably_misconfigured`?

brianpeiris:
Superseded by #542

PR #543, opened 2022-01-06
Respond with an error if twitter might be misconfigured

brianpeiris:
If twitter is misconfigured, either with a typo in the tokens configured, or a configuration error in the twitter developer console, it may result in API calls failing in reticulum. Previously, this would cause the hub_channel websocket to crash, and break the client in different ways (perhaps causing the game loop to stop). 
This PR adds error handling so that we forward the error to the client, instead of crashing entirely.

Goes with https://github.com/mozilla/hubs/pull/4997

### DIFF
--- a/lib/ret/twitter_client.ex
+++ b/lib/ret/twitter_client.ex
@@ -18,8 +18,21 @@ defmodule Ret.TwitterClient do
 
     token = Ret.OAuthToken.token_for_hub_and_account(hub_sid, account_id)
     callback_url = "#{get_redirect_uri()}?state=#{token}"
-    res = post("#{@twitter_api_base}/oauth/request_token", [{"oauth_callback", callback_url}], creds)
-    "#{@twitter_api_base}/oauth/authorize?" <> URI.encode_query(%{oauth_token: res["oauth_token"]})
+
+    token_response =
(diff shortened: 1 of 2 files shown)


lib/ret/twitter_client.ex
```diff
@@ -129,21 +142,27 @@ defmodule Ret.TwitterClient do
...
-      |> Map.get(:body)
-      |> to_string
 
-    case response_type do
-      :urlencoded -> body |> URI.decode_query()
-      :json -> body |> Poison.decode!()
-      _ -> body
+    case result do
+      :error ->
+        {:error, :twitter_probably_misconfigured}
```
brianpeiris, line 156: Should we use better token here, instead of `:twitter_probably_misconfigured`?

netpro2k (commented):
This change seems sane. We should figure out why the client ends up breaking though, it should be acceptable for the socket to crash. Is it specific to the twitter flow in particular?

brianpeiris:
Thanks for taking a look. There are some client changes that go along with this PR, though I don't think we took any time to figure out why the client breaks in that particular way. https://github.com/mozilla/hubs/pull/4997

PR #547, opened 2022-01-19
QA stage into master for HC fixes

robinkwilson:
Merges twitter fixes from : https://github.com/mozilla/reticulum/pull/546


netpro2k (approved):
lgtm. The old commits from polycosm branch are a noop, if we don't care about maintaining that branch anymore we can ditch them, if we do intend to use it again merging them into master makes sense so we don't keep picking them up.

PR #553, opened 2022-02-09
Hubs Cloud Cloudflare uploads fix

brianpeiris:
Account for Hubs Cloud stacks with Cloudflare CDN enabled. Instead of checking the host header against the configured storage host, we need to use the cf-worker header and check that it matches the suffix of the configured host.


robinkwilson (dismissed):

robinkwilson:
Can you change the PR to merge into qa-stage first please, we'll quickly merge into qa-test afterward. Other than that code looks good

PR #560, opened 2022-02-14
Add fonts.gstattic.com to connect-src for Troika

netpro2k:
Troika text loads fonts using fetch not directly as a CSS font so this needs to be in our connect-src as well (it was previously already in our font-src for use in CSS)


brianpeiris (approved):
LGTM

PR #562, opened 2022-02-16
Fix CSP for troika-text

netpro2k:
Troika Text library needs some additional CSP rules for its workers to load correctly.


brianpeiris (approved):
LGTM

PR #563, opened 2022-02-24
Add cross-origin headers required for SABs

netpro2k:
Adds the required headers to load Hubs in a secure context to access **SharedArrayBuffers** and high resolution timers. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements. This also increases security in general, but will make it more difficult to communicate with iframed hubs clients if that becomes necessary in the future.

Reticulum compliment to https://github.com/mozilla/hubs/pull/4252

### DIFF
--- a/lib/ret_web/router.ex
+++ b/lib/ret_web/router.ex
@@ -4,12 +4,20 @@ defmodule RetWeb.Router do
   use Sentry.Plug
 
   pipeline :secure_headers do
-    plug(:put_secure_browser_headers)
+    plug(:put_secure_browser_headers, %{
+      "cross-origin-opener-policy" => "same-origin",
+      "cross-origin-resource-policy" => "require-corp"
+    })
+
     plug(RetWeb.Plugs.AddCSP)
   end
 
   pipeline :strict_secure_headers do
-    plug(:put_secure_browser_headers)
+    plug(:put_secure_browser_headers, %{
+      "cross-origin-opener-policy" => "same-origin",
+      "cross-origin-resource-policy" => "require-corp"
+    })
+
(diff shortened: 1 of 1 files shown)


rawnsley:
Related PR on the client side https://github.com/mozilla/hubs/pull/4555

lib/ret_web/router.ex
```diff
@@ -4,12 +4,20 @@ defmodule RetWeb.Router do
   use Sentry.Plug
 
   pipeline :secure_headers do
-    plug(:put_secure_browser_headers)
+    plug(:put_secure_browser_headers, %{
+      "cross-origin-opener-policy" => "same-origin",
+      "cross-origin-resource-policy" => "require-corp"
```
brianpeiris, line 9: Pretty sure this is wrong, and you actually meant to use the embedder-policy header, not the resource-policy header.

```suggestion
      "cross-origin-embedder-policy" => "require-corp"
```

lib/ret_web/router.ex
```diff
@@ -4,12 +4,20 @@ defmodule RetWeb.Router do
...
+    })
+
     plug(RetWeb.Plugs.AddCSP)
   end
 
   pipeline :strict_secure_headers do
-    plug(:put_secure_browser_headers)
+    plug(:put_secure_browser_headers, %{
+      "cross-origin-opener-policy" => "same-origin",
+      "cross-origin-resource-policy" => "require-corp"
```
brianpeiris, line 18:
```suggestion
      "cross-origin-embedder-policy" => "require-corp"
```

brianpeiris (changes requested):
I think there was a critical typo here.

PR #566, opened 2022-02-28
New name tags typing and raised hand support

keianhzo:
This PR adds support for typing and hand raised states updates in Reticulum.

Hubs client related PR: https://github.com/mozilla/hubs/pull/5195

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -234,8 +234,12 @@ defmodule RetWeb.HubChannel do
 
   def handle_in("events:begin_recording", _payload, socket), do: socket |> set_presence_flag(:recording, true)
   def handle_in("events:end_recording", _payload, socket), do: socket |> set_presence_flag(:recording, false)
+  def handle_in("events:raise_hand", _payload, socket), do: socket |> set_presence_flag(:handRaised, true)
+  def handle_in("events:lower_hand", _payload, socket), do: socket |> set_presence_flag(:handRaised, false)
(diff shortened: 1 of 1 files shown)


lib/ret_web/channels/hub_channel.ex
```diff
@@ -234,8 +234,12 @@ defmodule RetWeb.HubChannel do
 
   def handle_in("events:begin_recording", _payload, socket), do: socket |> set_presence_flag(:recording, true)
   def handle_in("events:end_recording", _payload, socket), do: socket |> set_presence_flag(:recording, false)
+  def handle_in("events:raise_hand", _payload, socket), do: socket |> set_presence_flag(:handRaised, true)
+  def handle_in("events:lower_hand", _payload, socket), do: socket |> set_presence_flag(:handRaised, false)
```
brianpeiris, line 238: I would suggest using snake_case for tokens, `:hand_raised` instead of `:handRaised`, since that's the style we use consistently in elixir code.

brianpeiris (commented):
Left a comment, LGTM otherwise

lib/ret_web/channels/hub_channel.ex
keianhzo: Updated.

brianpeiris (approved):

lib/ret_web/channels/hub_channel.ex
brianpeiris: :+1: Just a reminder that you'll have to update the corresponding code in the client.

PR #573, opened 2022-03-18
Added presence endpoint for CCU across rooms

robinkwilson:
CCU presence endpoint added. Secured by header.

Ghosts are included in the presence count.

### DIFF
--- a/lib/ret_web/controllers/api/v1/presence_controller.ex
+++ b/lib/ret_web/controllers/api/v1/presence_controller.ex
@@ -1,0 +1,10 @@
+defmodule RetWeb.Api.V1.PresenceController do
+  use RetWeb, :controller
+
+  # Get presence count
+  def show(conn, _) do
+    count = RetWeb.Presence.present_session_count()
+
+    conn |> send_resp(200, %{count: count} |> Poison.encode!())
+  end
+end
--- a/lib/ret_web/controllers/api/v1/presence_controller.ex
+++ b/lib/ret_web/controllers/api/v1/presence_controller.ex
@@ -3,7 +3,7 @@ defmodule RetWeb.Api.V1.PresenceController do
 
   # Get presence count
   def show(conn, _) do
-    count = RetWeb.Presence.present_session_count()
+    count = RetWeb.Presence.present_ccu_in_room_count()
 
     conn |> send_resp(200, %{count: count} |> Poison.encode!())
   end
--- a/lib/ret_web/plugs/portal_header_authorization.ex
+++ b/lib/ret_web/plugs/portal_header_authorization.ex
@@ -1,0 +1,16 @@
+defmodule RetWeb.Plugs.PortalHeaderAuthorization do
+  import Plug.Conn
+  @header_name "x-ret-portal-access-key"
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    expected_value = Application.get_env(:ret, RetWeb.Plugs.PortalHeaderAuthorization)[:portal_access_key]
+
+    case conn |> get_req_header(@header_name) do
+      [""] -> conn |> send_resp(401, "") |> halt()
+      [^expected_value] -> conn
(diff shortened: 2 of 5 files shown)


lib/ret_web/router.ex
```diff
@@ -123,6 +127,11 @@ defmodule RetWeb.Router do
       resources("/slack", Api.V1.SlackController, only: [:create])
     end
 
+    scope "/v1", as: :api_v1 do
+      pipe_through([:portal_header_auth])
+      get("/presence", Api.V1.PresenceController, :show)
```
netpro2k, line 132: We should probably put all these privileged portal apis under some sub-path `/api/v1/internal` or something, just so we can easily identify them and make sure to treat them specially. Maybe also reflect that in the name of the controller like `InternalPresenceController`.

lib/ret_web/plugs/portal_header_authorization.ex
```diff
@@ -0,0 +1,16 @@
...
+  import Plug.Conn
+  @header_name "x-ret-portal-access-key"
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    expected_value = Application.get_env(:ret, RetWeb.Plugs.PortalHeaderAuthorization)[:portal_access_key]
+
+    case conn |> get_req_header(@header_name) do
+      [""] -> conn |> send_resp(401, "") |> halt()
```
netpro2k, line 11: Is this to catch the case where portal_access_key is miss-configured and left blank?

lib/ret_web/controllers/api/v1/presence_controller.ex
```diff
@@ -0,0 +1,10 @@
+defmodule RetWeb.Api.V1.PresenceController do
+  use RetWeb, :controller
+
+  # Get presence count
+  def show(conn, _) do
+    count = RetWeb.Presence.present_session_count()
```
netpro2k, line 6: I am a bit hazy on the difference between this and `present_member_count()`. From a quick glance it seems this will over count a bit since it will count users in the admin panel https://github.com/mozilla/hubs/blob/c5513b1f2d5f4a465189de569acbd4ea2a8e6edf/admin/src/admin.js#L279 and scene pages https://github.com/mozilla/hubs/blob/c5513b1f2d5f4a465189de569acbd4ea2a8e6edf/src/scene.js#L95 and briefly on other pages as part of checking if a user is an admin https://github.com/mozilla/hubs/blob/c5513b1f2d5f4a465189de569acbd4ea2a8e6edf/src/react-components/auth/AuthContext.js#L14 ... Member count conceptually seems more correct though it seems that only includes users not in the lobby. We might need a new stat for this.

That said this number is probably close enough to be useful to start and is easy to change later, so up to you if you feel like changing it here. I'm sure requirements will evolve to needing more from this response at some point anyway, might make more sense to tackle there with more clarity.

netpro2k (approved):
This mostly LGTM. 

Made a comment about the count probably not being exactly what we want, but its probably close enough to get us started, and we can fix it later once we have a better idea what we actually want to be counting. 

I do think namespacing the API would be a good idea though to prevent us from accidentally using it in a spot we shouldn't or accidentally loosening the rules around it in the future.

lib/ret_web/plugs/portal_header_authorization.ex
robinkwilson: Yes, we didn't want the header with an empty string value to pass.

lib/ret_web/controllers/api/v1/presence_controller.ex
robinkwilson: Verified, that it does count users in the admin panel.

robinkwilson: Tested brian's code. A scene page does not factor into the in room ccu count.

piyush2engine:
@robinkwilson Can u share me the endpoint for the presence count

PR #574, opened 2022-03-25
cowboy version bump

tanfarming:
tldr:
cowboy fixed a bug in v2.9 -- basically when 1. http2's dumping concurrent body streaming requests and 2. big blobs are getting streamed, cowboy's stream buffer mgmt code would chock the connection
gke's lb doesn't allow custom alpn settings so we'll get http2, and our blob requests to reticulum's /file would hang, most of the time during hubs room creation process

tl:
https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=162274105


brianpeiris:
The cowboy migration notes from 2.6 to 2.9 say that this upgrade should be safe. There were new features, improvements and bug fixes, but no breaking changes as far as I can tell. 2.9 also seems support Erlang/OTP 22 still, which is what we're using.

https://ninenines.eu/docs/en/cowboy/2.7/guide/migrating_from_2.6/
https://ninenines.eu/docs/en/cowboy/2.8/guide/migrating_from_2.7/
https://ninenines.eu/docs/en/cowboy/2.9/guide/migrating_from_2.8/

brianpeiris:
This also updates sub-dependencies `ranch` from 1.7.1 to 1.8.0, and `cowlib` from 2.7.3 to 2.11.0. 
The ranch update is just a few bug fixes https://ninenines.eu/docs/en/ranch/1.8/guide/migrating_from_1.7/
cowlib is basically a helper library for cowboy, although it doesn't have changelogs, its safe to assume that the update is as safe as a cowboy update. https://github.com/ninenines/cowlib/compare/2.7.3...2.11.0

brianpeiris:
Ran basic integration tests with this running locally. Basic Hubs, admin panel, and Spoke functionality looks to be working correctly.
Will wait till Monday before merging this, since it is a core change and may behave differently in production.

PR #575, opened 2022-04-04
ret-turkey-config-update

tanfarming:


brianpeiris (approved):
LGTM. (`rel/config.toml` is only used for turkey ops, so this will not affect HMC)

PR #580, opened 2022-04-22
Portal/internal endpoints /v1/internal -> /api-internal/v1

robinkwilson:
To ensure the internal endpoints are private and secure, organize the endpoint under /api-internal/v1 and NOT /api/v1.

### DIFF
--- a/lib/ret_web/router.ex
+++ b/lib/ret_web/router.ex
@@ -197,7 +197,7 @@ defmodule RetWeb.Router do
       [:portal_header_auth, :secure_headers, :parsed_body, :api] ++ if(Mix.env() == :prod, do: [:ssl_only], else: [])
     )
 
-    scope "/v1", as: :api_v1 do
+    scope "/v1", as: :api_internal_v1 do
       get("/presence", ApiInternal.V1.PresenceController, :show)
     end
   end
--- a/lib/ret_web/router.ex
+++ b/lib/ret_web/router.ex
@@ -202,7 +202,7 @@ defmodule RetWeb.Router do
     pipe_through([:portal_header_auth, :secure_headers, :parsed_body, :api] ++ if(Mix.env() == :prod, do: [:ssl_only], else: []))
 
     scope "/v1", as: :api_v1 do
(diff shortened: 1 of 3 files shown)


lib/ret_web/router.ex
```diff
@@ -127,9 +128,9 @@ defmodule RetWeb.Router do
       resources("/slack", Api.V1.SlackController, only: [:create])
     end
 
-    scope "/v1/internal", as: :api_v1 do
+    scope "/api-internal/v1", as: :api_v1 do
```
netpro2k: Is there a reason we want these routes on both `/api/api-internal/v1` and `/api-internal/v1`?

lib/ret_web/router.ex
```diff
@@ -127,9 +128,9 @@ defmodule RetWeb.Router do
...
       pipe_through([:portal_header_auth])
-      get("/presence", Api.V1.PresenceController, :show)
+      get("/presence", ApiInternal.PresenceController, :show)
```
netpro2k: Looks like you are missing V1 here?
```suggestion
      get("/presence", ApiInternal.V1.PresenceController, :show)
```

lib/ret_web/router.ex
```diff
@@ -197,6 +197,16 @@ defmodule RetWeb.Router do
     forward "/", Absinthe.Plug, json_codec: Jason, schema: RetWeb.Schema
   end
 
+  scope "/api-internal", RetWeb do
+    pipe_through(
+      [:portal_header_auth, :secure_headers, :parsed_body, :api] ++ if(Mix.env() == :prod, do: [:ssl_only], else: [])
+    )
+
+    scope "/v1", as: :api_v1 do
```
netpro2k, line 200: I forget what `as: :api_v1` does here but we depending on what it does we maybe want something like `as: api_internal_v1`. My guess is its for generating utility functions.

netpro2k (changes requested):
We probably don't want the second copy of the route under `/api` and we should see what `as: :api_v1` does and consider changing it. Otherwise lgtm.

brianpeiris (approved):
LGTM

PR #581, opened 2022-04-22
mix format

robinkwilson:


brianpeiris:
I think this is outdated, and I've now added the files in `config/` to our CI build, so we should catch this automatically in the future.

PR #590, opened 2022-05-12
Add internal endpoint for retrieving storage usage

brianpeiris:


robinkwilson (approved):
Okay much better. Glad we discussed the error case.

PR #606, opened 2022-06-27
Use `du` storage command for turkey instead of `df`

robinkwilson:
`Df` doesn't work because of the combination of how `df` works and the container mounting in gcs. Use `du` instead for turkey.


brianpeiris (approved):
LGTM

PR #609, opened 2022-07-14
Create protected endpoint for getting max ccu in a timerange

robinkwilson:
- Creates protected api-internal endpoint /api-internal/v1/presence/daily_max
- Adds Tests for Ret.NodeStat max_ccu_for_time_range()

Example Req:
https://hubs.local:4000/api-internal/v1/presence/daily_max?start_time=2022-07-13 23:44:30.360455&end_time=2022-07-13 23:44:30.360455
Example Res:
{"max_ccu":0}


lib/ret_web/controllers/api-internal/v1/presence_controller.ex
```diff
@@ -1,10 +1,23 @@
...
   # Get presence count
   def show(conn, _) do
     count = RetWeb.Presence.present_ccu_in_room_count()
 
     conn |> send_resp(200, %{count: count} |> Poison.encode!())
   end
+
+  # Params start_time and end_time should be in iso format such as "2000-02-28 23:00:13"
+  # or what is returned from NaiveDateTime.to_string()
+  def daily_max(conn, %{"start_time" => start_time_str, "end_time" => end_time_str}) do
```
robinkwilson, line 14: Should this be in presence controller or in a new node stat controller?

lib/ret_web/router.ex
```diff
@@ -199,6 +199,7 @@ defmodule RetWeb.Router do
 
     scope "/v1", as: :api_internal_v1 do
       get("/presence", ApiInternal.V1.PresenceController, :show)
+      get("/presence/daily_max", ApiInternal.V1.PresenceController, :daily_max)
```
brianpeiris, line 202: nit: Not a big deal, since this is an internal API, but `daily_max` implies a fixed period, when the API actually accepts any range. So maybe just call it `/presence/max`, or `/presence/range_max`, or something.

Same for the function name in PresenceController.

test/ret/node_stat_test.exs
```diff
@@ -0,0 +1,100 @@
+defmodule Ret.NodeStatTest do
+  use Ret.DataCase
+  alias Ret.{NodeStat, Repo}
+
+  describe "NodeStat.max_ccu_for_time_range() tests" do
+    @tag marked: true
```
brianpeiris, line 6: Remember to remove the tags. Same below.
(We should probably add a lint rule for this if we can some time, I forget to do it too)

test/ret/node_stat_test.exs
```diff
@@ -0,0 +1,100 @@
...
+    test "Return 0 for no NodeStats collected for today" do
+      # Do not seed database
+      %{today: today} = get_times()
+      assert 0 == NodeStat.max_ccu_for_time_range(today.start_time, today.end_time)
+    end
+
+    @tag marked: true
+    test "Do not return outside of time boundary" do
+      seed_today_node_stats([0, 10, 5])
+      seed_tomorrow_and_yesterday_node_stats(20)
```
brianpeiris, line 25: nit: I do think it's nice to keep test code readable like this, but there is also a philosophy that you should minimize the number of helper functions you use, so that you can read the actual behavior of a test without having to jump around to read the behavior of a bunch of helper functions as well. As a middle-ground, I might have used a single generalized helper, like:
```elixir
%{today: today, tomorrow: tomorrow, yesterday: yesterday} = get_times()

seed_node_stats([
  [yesterday, [20]],
  [today, [0, 10, 5],
  [tomorrow, [20]]
])
```

test/ret/node_stat_test.exs
```diff
@@ -0,0 +1,100 @@
+defmodule Ret.NodeStatTest do
```
brianpeiris, line 1: Nice tests!

brianpeiris (approved):
Just nits. LGTM!

lib/ret_web/controllers/api-internal/v1/presence_controller.ex
brianpeiris: Nah, I think it's fine to keep it here.

brianpeiris (approved):

PR #616, opened 2022-08-15
get_dialog-pod fixes for turkeymode

tanfarming:
dialog's port 7000 in turkey arch is internal and only used to report local levels

### DIFF
--- a/lib/ret/load_balancing/janus_load_status.ex
+++ b/lib/ret/load_balancing/janus_load_status.ex
@@ -44,7 +44,7 @@ defmodule Ret.JanusLoadStatus do
         |> Enum.map(fn {a, b, c, d} -> "#{a}.#{b}.#{c}.#{d}" end)
 
       for host <- hosts do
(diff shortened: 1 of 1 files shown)


lib/ret/load_balancing/janus_load_status.ex
```diff
@@ -44,7 +44,7 @@ defmodule Ret.JanusLoadStatus do
         |> Enum.map(fn {a, b, c, d} -> "#{a}.#{b}.#{c}.#{d}" end)
 
       for host <- hosts do
-        %{body: body} = HTTPoison.get!("https://#{host}:7000/meta", [], hackney: [:insecure])
+        %{body: body} = HTTPoison.get!("http://#{host}:7000/meta", [], hackney: [:insecure])
```
brianpeiris, line 47: We can remove the hackney config if we're using http.
```suggestion
        %{body: body} = HTTPoison.get!("http://#{host}:7000/meta")
```

brianpeiris (approved):
Left a comment to simplify it a bit, but LGTM

PR #617, opened 2022-08-25
Account Deletion: Initial pass at account deletion

brianpeiris:
Opening a draft PR to validate some initial decisions about this approach.

- Start using the Ret module for business logic. Since we'll be interacting with a bunch of different resources for account deletion, it seemed right to adopt the context pattern, instead of just adding more helper functions to the existing Account module
- Alters the oauth_providers and api_credentials tables to add proper foreign key relationships to the accounts table. This allows us to take advantage of cascading deletes in the database
- Adds a new Canada authorization rule to ensure that only admin accounts can delete other accounts.
- Adds a single test to exercise deletion of the account resources -- Login, Identity, OAuthProvider, and Credentials
- Adds a delete action to AccountController, though this may not be the final API form


lib/ret/account_context.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Ret.AccountContext do
+  import Canada, only: [can?: 2]
+  alias Ret.{Account}
+
+  def delete_account(%Account{} = acting_account, %Account{} = account_to_delete) do
+    if can?(acting_account, delete_account(account_to_delete)) do
+      case Ret.Repo.delete(account_to_delete) do
+        {:ok, _} -> {:ok}
```
bryanenders, line 8: The idiom would be to return `:ok` bare. You’d only wrap it in a tuple if you were tagging a value (e.g., `{:ok, 42}`). Monuples (1-tuples) aren’t ever needed in Erlang or Elixir.

lib/ret/account.ex
```diff
@@ -38,6 +38,10 @@ defmodule Ret.Account do
   def has_admin_accounts?(), do: from(a in Account, limit: 1) |> where(is_admin: true) |> Repo.exists?()
   def exists_for_email?(email), do: account_for_email(email) != nil
 
+  def get_account_by_id(account_id) do
+    Repo.get(Account, account_id)
```
bryanenders, line 42: This isn’t the first use of `Repo` inside this module, but this is a good opportunity to do the right thing. A good pattern for managing state is: functional core, imperative shell. Struct manipulation is low-level, which make stateful calls (e.g., `Repo` calls) within a struct module a design smell.

How to fix? Reserve this module for pure functions that operate on `Ret.Account.t()` and put any new repo calls in the context boundary module.

lib/ret_web/controllers/api/v1/account_controller.ex
```diff
@@ -30,6 +30,17 @@ defmodule RetWeb.Api.V1.AccountController do
     exec_api_create(conn, params, @record_schema, &process_account_update_record/2)
   end
 
+  def delete(conn, params) do
+    current_account = Guardian.Plug.current_resource(conn)
+    account_to_delete = Ret.Account.get_account_by_id(params[:account_id])
+
+    case Ret.AccountContext.delete_account(current_account, account_to_delete) do
```
bryanenders, line 37: What if `delete_account/2` took the id of the account to delete, rather than the data structure? Then there would only be one domain call instead of two.

priv/repo/migrations/20220819164826_add_account_references_and_delete_behaviors_to_tables.exs
```diff
@@ -0,0 +1,15 @@
...
+    alter table(:oauth_providers) do
+      modify(:account_id, references(:accounts, column: :account_id, on_delete: :delete_all))
+    end
+
+    execute("alter table api_credentials drop constraint api_credentials_account_id_fkey")
+
+    alter table(:api_credentials) do
+      modify(:account_id, references(:accounts, column: :account_id, on_delete: :delete_all))
+    end
+  end
```
bryanenders, line 14: The `change/0` signature is reserved for reversible migrations. Since the reverse cannot be inferred, you’ll want an `up/0` and `down/0` definition.

lib/ret/account_context.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Ret.AccountContext do
```
bryanenders, line 1: Since there are rules governing the use of things within a bounded context, I often start with a single, general-purpose context, then split if needed once I can see the patterns in module collaboration.

bryanenders (commented):
Glad to see a nice boundary forming between the web layer and the business layer! Here are a few suggestions for improvement.

lib/ret_web/controllers/api/v1/account_controller.ex
bryanenders: This would obviate `Account.get_account_by_id/1` entirely.

lib/ret/account_context.ex
brianpeiris: Are you suggesting that this should just be `Ret.Context`?

lib/ret_web/controllers/api/v1/account_controller.ex
brianpeiris: I think I'll stick to passing structs around, since that's the pattern we generally use across the codebase. Guardian's current_resource is already an Account struct. The Canada authorization rule also requires a struct.

lib/ret/account_context.ex
bryanenders: I’d probably just call it `Ret`

bryanenders: If you’re partial to having `Context` in the name, I might go with `Ret.GeneralContext`.

brianpeiris: Went ahead and put the functions directly under the main `Ret` module. Though I wasn't sure what the test should be called. I went with `context_test.exs` and `Ret.ContextTest`.

bryanenders: Elixir doesn’t care, but if you want it to match you can move it to `test/ret_test.exs` and name it `RetTest`.

brianpeiris: :+1: makes sense

test/ret/context_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  describe "account deletion" do
+    test "deletes account, login, identity, oauthproviders, and api_credentials" do
+      {:ok, admin_account: admin_account} = create_admin_account("admin")
+      test_account = create_account("test")
+
+      Account.set_identity!(test_account, "test identity")
+
+      %Ret.OAuthProvider{}
+      |> Ecto.Changeset.change(source: :discord, account: test_account, provider_account_id: "discord-test-user")
+      |> Ret.Repo.insert_or_update()
```
bryanenders: This could probably simplified. Like so:
`Repo.insert(%OAuthProvider{source: :discord, account: test_account, provider_account_id: "discord-test-user"})`

test/ret_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  describe "account deletion" do
+    test "deletes account, login, identity, oauthproviders, and api_credentials" do
+      {:ok, admin_account: admin_account} = create_admin_account("admin")
+      test_account = create_account("test")
+
+      Account.set_identity!(test_account, "test identity")
+
+      %Ret.OAuthProvider{}
+      |> Ecto.Changeset.change(source: :discord, account: test_account, provider_account_id: "discord-test-user")
+      |> Ret.Repo.insert_or_update()
```
bryanenders, line 16: This could probably simplified. Like so:
`Repo.insert(%OAuthProvider{source: :discord, account: test_account, provider_account_id: "discord-test-user"})`

bryanenders: (A changeset isn’t required for an insert)

PR #618, opened 2022-09-01
Account Deletion: Cascade account deletion to hubs and associated entities

brianpeiris:
This PR builds on top of #617 and cascades an account deletion to the hubs that the account own, and all the descendant entities related to a Hub, including HubBindings, HubInvites, WebPushSubscriptions, AccountFavorites, HubRoleMemberships, and RoomObjects.

The migration from #617 was generalized and DRY'd up to handle several tables where the reference to the Hub table was either missing or did not have an on_delete action specified.

A test was added to verify the cascaded deletion of these entities.


priv/repo/migrations/20220819164826_add_foreign_references_and_delete_behaviors_to_tables.exs
```diff
@@ -0,0 +1,54 @@
+defmodule Ret.Repo.Migrations.AddForeignReferencesAndDeleteBehaviorToTables do
+  use Ecto.Migration
+
+  @tables_missing_references [
```
bryanenders, line 4: I would drop the `tables_` prefix. It’s confusing from the module attribute.

priv/repo/migrations/20220819164826_add_foreign_references_and_delete_behaviors_to_tables.exs
```diff
@@ -0,0 +1,54 @@
...
+    # table, foreign_table, column
+    {:hub_bindings, :hubs, :hub_id},
+    {:hub_invites, :hubs, :hub_id},
+    {:oauth_providers, :accounts, :account_id},
+    {:web_push_subscriptions, :hubs, :hub_id}
+  ]
+
+  @tables_missing_on_delete_action [
```
bryanenders, line 12: Ditto

bryanenders (approved):
Looks good. Just one suggestion.

priv/repo/migrations/20220819164826_add_foreign_references_and_delete_behaviors_to_tables.exs
bryanenders: Since it’s three pieces of data, I’d just call them tuples or triples or drop the noun. E.g.`@missing_reference_triples`, `@missing_reference_tuples`, `@missing_reference`

bryanenders: It’s just a nit though 😄

brianpeiris:
During manual testing, I found that we failed to account for deletion of a hub owned by UserA, where UserB also had objects associated with that hub. Previously, we had a test to cover deletion of objects associated with UserB, when UserB was deleted. Commit https://github.com/mozilla/reticulum/pull/618/commits/29ff18354761290a2fad8252ebbc73579e2871b8 adds a test and on_delete actions to cover the opposite side of that relationship.

bryanenders:

PR #620, opened 2022-09-11
Account Deletion: Avatar deletion

brianpeiris:
This PR builds on top of #618 and adds code to delete an account's avatars.

Avatar deletion is a bit trickier, since we have to handle deletion of OwnedFiles, and physical files on disk. We also have to deal with AvatarListings, which depend on avatars, as well as parent/child relationships between avatars.

To this end, when deleting an account with AvatarListings or parent avatars, we reassign those listings and parent avatars to the admin account that is doing the deletion. 

Simple avatars, with no listings or children, are deleted after their OwnedFiles are collected, which are subsequently deleted as well.


lib/ret.ex
```diff
@@ -16,4 +21,76 @@ defmodule Ret do
...
     end
   end
+
+  defp delete_avatars_for_account(%Account{} = account) do
+    account_avatars_query = from(avatar in Avatar, where: avatar.account_id == ^account.account_id)
+    account_avatars = Repo.all(account_avatars_query) |> Repo.preload(Avatar.file_columns())
+
+    account_avatar_owned_files =
+      account_avatars
+      |> Enum.flat_map(fn avatar ->
```
bryanenders, line 31: Does this actually need to be flattened? The variable name in `Enum.filter/2` would seem to indicate that it does not.

lib/ret.ex
```diff
@@ -16,4 +21,76 @@ defmodule Ret do
...
+        Avatar.file_columns()
+        |> Enum.map(fn column -> Map.fetch!(avatar, column) end)
+        |> Enum.filter(fn owned_file -> owned_file != nil end)
```
bryanenders, line 34: Nested pipelines are difficult to read. The anonymous function given to `Enum.flat_map/2` is a strong contender for a private function.

Additionally the map and filter calls make this a good opportunity for a list comprehension.

```elixir
defp avatar_owned_files(avatar) do
  for column <- Avatar.file_columns(),
      owned_file = Map.fetch!(avatar, column),
      do: owned_file
end
```

The first line of the comprehension is the generator. The second line is the filter. It rejects any falsey values. We bind in the filter so we can access the value in the final line.

lib/ret.ex
```diff
@@ -16,4 +21,76 @@ defmodule Ret do
...
+      end)
+
+    Repo.delete_all(account_avatars_query)
+    delete_owned_files(account_avatar_owned_files)
+  end
+
+  defp delete_owned_files(owned_files) when is_list(owned_files) do
+    for owned_file <- owned_files, do: OwnedFile.set_inactive(owned_file)
+    for owned_file <- owned_files, do: Storage.rm_files_for_owned_file(owned_file)
+    for owned_file <- owned_files, do: Repo.delete(owned_file)
```
bryanenders, line 44: Is there any reason we couldn’t do this in one iteration?

```elixir
for owned_file <- owned_files do
  OwnedFile.set_inactive(owned_file)
  Storage.rm_files_for_owned_file(owned_file)
  Repo.delete(owned_file)
end
```

test/ret_test.exs
```diff
@@ -157,6 +161,147 @@ defmodule RetTest do
...
+
+      1 = count(Avatar, test_account)
+      7 = count(OwnedFile, test_account)
+      true = owned_files_exist?(avatar_owned_files)
+
+      assert :ok = Ret.delete_account(admin_account, test_account)
+
+      assert 0 === count(Avatar, test_account)
+      assert 0 === count(OwnedFile, test_account)
+      assert false === owned_files_exist?(avatar_owned_files)
```
bryanenders, line 179: This might read a little nicer as `refute owned_files_exist?(avatar_owned_files)`

test/ret_test.exs
```diff
@@ -157,6 +161,147 @@ defmodule RetTest do
...
+
+      assert 0 === count(Avatar, target_account)
+      assert 0 === count(OwnedFile, target_account)
+      assert 0 === count(AvatarListing, target_account)
+
+      assert 1 === count(Avatar, admin_account)
+      assert 7 === count(OwnedFile, admin_account)
+      assert 1 === count(AvatarListing, admin_account)
+
+      assert true === owned_files_exist?(avatar_owned_files)
```
bryanenders, line 207: This might read a little nicer as `assert owned_files_exist?(avatar_owned_files)`

test/ret_test.exs
```diff
@@ -157,6 +161,147 @@ defmodule RetTest do
...
+      assert 0 === count(Avatar, target_account)
+      assert 0 === count(OwnedFile, target_account)
+
+      assert 1 === count(Avatar, other_account)
+      assert 0 === count(OwnedFile, other_account)
+
+      assert 1 === count(Avatar, admin_account)
+      assert 7 === count(OwnedFile, admin_account)
+
+      assert true === owned_files_exist?(avatar_owned_files)
```
bryanenders, line 238: This might read a little nicer as `assert owned_files_exist?(avatar_owned_files)`

bryanenders (commented):
Just some readability suggestions

lib/ret.ex
brianpeiris: Hmm. Each avatar has multiple OwnedFiles associated with it. This iterates through each avatar, fetches each OwnedFile, filters the nil ones, and then flattens them such that we have a single array of all the OwnedFiles associated with all the avatars belonging to the user. So yeah, it does need to be flattened.

lib/ret.ex
brianpeiris: Initially I figured it would be better to maintain an order of operations, where we first ensure that all the files are inactive, and cannot be accessed, then we delete the physical files on disk, before we delete all the DB records. But I suppose there's no harm in doing them one by one and it is probably more performant.

lib/ret.ex
```diff
@@ -16,4 +21,76 @@ defmodule Ret do
...
+    account_avatars_query = from(avatar in Avatar, where: avatar.account_id == ^account.account_id)
+    account_avatars = Repo.all(account_avatars_query) |> Repo.preload(Avatar.file_columns())
+    account_avatar_owned_files = Enum.flat_map(account_avatars, &avatar_owned_files/1)
+    Repo.delete_all(account_avatars_query)
+    delete_owned_files(account_avatar_owned_files)
+  end
+
+  defp avatar_owned_files(%Avatar{} = avatar) do
+    for column <- Avatar.file_columns(),
+        owned_file = Map.fetch!(avatar, column),
```
bryanenders, line 35: @brianpeiris Does this fetch return a list of owned files?

brianpeiris: Nope, each column represents a single owned file, but each avatar has multiple columns that reference owned files, so this function will return a flat array of files for a single avatar.

brianpeiris:
Doing some actual testing with a full stack, and I found that we don't actually associate an account_id with avatar_listings, so we need to reassign OwnedFiles using their account_id instead. Modified the tests to reflect that scenario as well in https://github.com/mozilla/reticulum/pull/620/commits/fcc0a7ee5429bb1fa3cc1c905458f6755a1d1aaf

PR #622, opened 2022-09-12
Account Deletion: Project and Asset deletion

brianpeiris:
Builds on top of #620 and handles Project, Asset, and ProjectAsset deletion for an account. Similar to Avatars, Projects and Assets have OwnedFiles associated with them, so this PR turned `delete_avatars_for_account` into a generic function that handles OwnedFile deletion for other entities as well.


lib/ret.ex
```diff
@@ -1,17 +1,27 @@
...
-      delete_avatars_for_account(account_to_delete)
+
+      for {query, file_columns} <- [
+            {from(avatar in Avatar, where: avatar.account_id == ^account_to_delete.account_id), Avatar.file_columns()},
+            {from(asset in Asset, where: asset.account_id == ^account_to_delete.account_id), @asset_file_columns},
+            {from(project in Project, where: project.created_by_account_id == ^account_to_delete.account_id),
+             @project_file_columns}
+          ] do
+        delete_entities_with_owned_files(query, file_columns)
+      end
```
bryanenders, line 24: Since this is just sending the double to the function as arguments, this might be easier to grok without the list comprehension.

```elixir
delete_entities_with_owned_files(
  from(avatar in Avatar, where: avatar.account_id == ^account_to_delete.account_id),
  Avatar.file_columns()
)

delete_entities_with_owned_files(
  from(asset in Asset, where: asset.account_id == ^account_to_delete.account_id),
  @asset_file_columns
)

delete_entities_with_owned_files(
  from(project in Project, where: project.created_by_account_id == ^account_to_delete.account_id),
  @project_file_columns
)
```

It repeats characters (i.e., `delete_entities_with_owned_files`) but not logic.

bryanenders (commented):
One little readability suggestion

PR #623, opened 2022-09-12
Prepare for local orchestration

bryanenders:
Why
---
Content administration screens cannot load and the following error messages appear in the browser when Hubs is orchestrated with Docker Compose:
* Missing file index.html. Please try again.
* Missing file admin.html. Please try again.

Reticulum is run against an external Dialog instance.  I want to be able to tweak, and thus run against, a local instance.

I want to be able to easily run common commands inside the orchestrated container from my own shell.

What
----
* Allow hostnames for the following to be supplied by OS vars:
    + hubs client
    + hubs admin
    + postgREST
    + spoke
    + dialog
* Add `bin/iex`, `bin/mix`, and `bin/psql`

### DIFF
--- a/bin/iex
+++ b/bin/iex
@@ -1,0 +1,3 @@
+#!/bin/bash
+basedir=$(dirname $0)
+docker-compose -f $basedir/../../../docker-compose.yml exec reticulum iex $*
--- a/lib/ret_web/plugs/proxies.ex
+++ b/lib/ret_web/plugs/proxies.ex
@@ -1,6 +1,20 @@
 defmodule RetWeb.Plugs.PostgrestProxy do
   use Plug.Builder
-  plug ReverseProxyPlug, upstream: "http://localhost:3000"
+
+  plug :call
+
+  @spec call(Plug.Conn.t(), []) :: Plug.Conn.t()
+  def call(conn, []) do
+    opts = ReverseProxyPlug.init(upstream: "http://#{hostname()}:3000")
+    ReverseProxyPlug.call(conn, opts)
+  end
+
+  @spec hostname :: String.t()
+  defp hostname,
+    do:
+      :ret
+      |> Application.fetch_env!(__MODULE__)
+      |> Keyword.fetch!(:hostname)
 end
 
 defmodule RetWeb.Plugs.ItaProxy do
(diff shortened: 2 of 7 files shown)


bin/iex
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+docker-compose -f ../docker-compose.yml exec reticulum iex $*
```
brianpeiris, line 2: Might be good to add a README to this `bin/` directory to explain what these scripts are, and how they should be used. Maybe it would just be one sentence, with a link to the compose repo.

lib/ret_web/plugs/proxies.ex
```diff
@@ -1,6 +1,20 @@
...
+    opts = ReverseProxyPlug.init(upstream: "http://#{hostname()}:3000")
+    ReverseProxyPlug.call(conn, opts)
+  end
+
+  @spec hostname :: String.t()
+  defp hostname,
+    do:
+      :ret
+      |> Application.fetch_env!(__MODULE__)
+      |> Keyword.fetch!(:hostname)
```
brianpeiris, line 17: Looking at this again, I wonder if we should use the same pattern we use elsewhere in the code. More of a nit, but it is a bit cleaner:
```suggestion
  defp hostname, do: Application.get_env(:ret, __MODULE__)[:hostname]
```

brianpeiris (approved):
Minor comments. LGTM otherwise.

lib/ret_web/plugs/proxies.ex
bryanenders: `Application.get_env/2` and `Access` are permissive. If the application is misconfigured and either  `RetWeb.Plugs.PostgrestProxy` or `:hostname` don’t exist, a rogue `nil` is propagated through the call chain. `Application.fetch_env!/2` and `Keyword.fetch!/2` force the misconfigured application to fail within `hostname/0`.

brianpeiris: Ah, gotcha. Makes sense. Carry on.

bin/iex
bryanenders: Done. Will add a link to the Hubs Compose repo once that is up.

brianpeiris:
Just wanted to note that the change to `proxies.ex` is a code change that will run in production and in Hubs Cloud. The change itself should be equivalent, but worth calling out when this is merged and deployed, and especially worth testing in the next Hubs Cloud release.

PR #624, opened 2022-09-13
Account Deletion: Scene deletion, and Scene listing reassignment

brianpeiris:
Builds on #622 and handles scene deletion and scene listing reassignment, when an account is deleted. 
The foreign reference migration was amended so that the parent_scene_id column in projects and scenes can be set to `:nilify_all`
The parent scenes relationship only exists for book-keeping purposes, so it is safe to nilify the foreign reference in this case.


test/ret_test.exs
```diff
@@ -262,6 +264,110 @@ defmodule RetTest do
...
+      {:ok, admin_account: admin_account} = create_admin_account("admin")
+      target_account = create_account("target")
+      other_account = create_account("other")
+
+      project = create_project(target_account)
+      scene = create_scene(project, target_account)
+      other_project = create_project(other_account)
+      other_project |> Ecto.Changeset.change(parent_scene_id: scene.scene_id) |> Repo.update!()
+      other_scene = create_scene(other_project, other_account)
+      other_scene |> Ecto.Changeset.change(parent_scene_id: scene.scene_id) |> Repo.update!()
```
bryanenders, line 296: 2 tiny nits:
1. Vertical pipelines are much easier to follow than horizontal ones
2. `other_scene` (and potentially `other_project` can be inlined if you aren’t relying on them being explanatory

You could put it all in a single pipeline like so:
```elixir
other_account
|> create_project()
|> Ecto.Changeset.change(parent_scene_id: scene.scene_id)
|> Repo.update!()
|> create_scene(other_account)
|> Ecto.Changeset.change(parent_scene_id: scene.scene_id)
|> Repo.update!()
```

I’m not sure I like two side effects in one pipeline though, so this seems like the happy compromise:
```elixir
other_project =
  other_account
  |> create_project()
  |> Ecto.Changeset.change(parent_scene_id: scene.scene_id)
  |> Repo.update!()

other_project
|> create_scene(other_account)
|> Ecto.Changeset.change(parent_scene_id: scene.scene_id)
|> Repo.update!()
```

test/ret_test.exs
```diff
@@ -262,6 +264,110 @@ defmodule RetTest do
...
+        scene_owned_file: scene.scene_owned_file,
+        screenshot_owned_file: scene.screenshot_owned_file,
+        model_owned_file: scene.model_owned_file
+      })
+
+    scene_listing
+  end
+
+  defp create_scene(%Project{} = project, %Account{} = account) do
+    scene_owned_files = 1..3 |> Enum.map(fn _ -> generate_temp_owned_file(account) end)
```
bryanenders, line 354: An alternate way to write this:
```elixir
Stream.repeatedly(fn -> generate_temp_owned_file(account) end) |> Enum.take(3)
```

bryanenders (commented):
Very minor suggestions

test/ret_test.exs
brianpeiris: Put this into a separate PR for refactoring: https://github.com/mozilla/reticulum/pull/626

brianpeiris:
Commit https://github.com/mozilla/reticulum/pull/624/commits/0a7e73edc721f4043a291f6791edcdf7112d3b6a adds a test case for a scenario where a project and scene share owned files. The deletion code was changed to account for this scenario, since the projects and scenes both need to be deleted before their owned files are deleted.

lib/ret.ex
```diff
@@ -24,10 +26,17 @@ defmodule Ret do
...
+      scene_query = from(scene in Scene, where: scene.account_id == ^account_to_delete.account_id)
+
+      project_and_scene_owned_files =
+        collect_owned_files(project_query, @project_file_columns) ++
+          collect_owned_files(scene_query, @scene_file_columns)
+
+      Repo.delete_all(project_query)
+      Repo.delete_all(scene_query)
+
+      delete_owned_files(project_and_scene_owned_files)
```
bryanenders, line 39: We probably want to make this atomic so that we can completely comply with GDPR, even if a user is adding files as we are deleting their account. The way to do so would be to put all the database operations in a transaction, and only if that transaction is committed to delete the actual files on disk.

I know it’s a pain to rewrite that way. I can help. I’m sorry I didn’t catch it sooner!

bryanenders (changes requested):

bryanenders:

PR #625, opened 2022-09-13
Account Deletion: Delete owned files

brianpeiris:
Builds on #624 and handles deletion of OwnedFiles, which are typically uploaded independently by a user, for use in a room.

PR #626, opened 2022-09-13
Account Deletion: Refactoring

brianpeiris:
This PR is just a refactoring pass that builds on top of #625

- Refactor the generation of temporary owned files into a helper function that uses Stream.repeatedly.


test/ret_test.exs
```diff
@@ -576,6 +576,10 @@ defmodule RetTest do
     end)
   end
 
+  defp generate_temp_owned_files(num_files, %Account{} = account) when is_integer(num_files) do
```
bryanenders, line 579: Tiny nit, but including the number of files as a parameter adds to the cognitive overhead of reading/using this function.

`Enum.take/2` is part of the language, so anyone diving into this codebase would have an intuitive understanding for what this does:
```elixir
account
|> generate_temp_owned_files()
|> Enum.take(5)
```

Whereas they’d have to look at the function definition to be sure what the `5` means in `generate_temp_owned_files(5, account)`.

bryanenders:

PR #628, opened 2022-09-15
Account Deletion

brianpeiris:
This amends the account deletion migration to ensure that handles cases where existing databases do not match the new constraints that we're adding. It will delete orphaned data when data was manually deleted in the database. For example, if a hub had a hub_invite, and then the hub was deleted while the hub_invite remained, the migration will delete hub_invites that no longer have hubs associated with them. It does this for all tables and columns where we are adding foreign key reference.

### DIFF
--- a/lib/ret.ex
+++ b/lib/ret.ex
@@ -1,9 +1,158 @@
 defmodule Ret do
-  @moduledoc """
-  Ret keeps the contexts that define your domain
-  and business logic.
+  import Canada, only: [can?: 2]
+  import Ecto.Query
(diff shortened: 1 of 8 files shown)


bryanenders (changes requested):
I would separate the data migration (i.e., delete orphaned data) from the schema migration. In this case, the data migration should precede the schema migration.

bryanenders (approved):

bryanenders:
Used as the basis for https://github.com/mozilla/reticulum/pull/634

PR #629, opened 2022-09-26
Mute room

keianhzo:


lib/ret/hub.ex
```diff
@@ -59,7 +63,9 @@ defmodule Ret.Hub do
     spawn_drawing: false,
     pin_objects: false,
     spawn_emoji: false,
-    fly: false
+    fly: false,
+    voice_chat: false,
+    text_chat: false
   }
```
bryanenders, line 69: It seems like there is redundancy between these two module attributes and `@object_actions` in our `Canada.Can` implementations. Not necessary at this moment, but we could set one source of truth for both.

priv/repo/migrations/20220908132321_add_voice_text_chat.exs
```diff
@@ -0,0 +1,45 @@
...
+          }
+          |> Ret.Hub.member_permissions_to_int()
+      )
+    end
+
+    execute("UPDATE ret0.hubs SET member_permissions = member_permissions | 192;")
+  end
+
+  def down do
+    # Probably don't ever want to revert this migration, but in case we do:
```
bryanenders, line 26: We should at least be able to test rolling this data migration back. Assuming no data is lost either direction, why not uncomment it?

bryanenders (approved):
One question

priv/repo/migrations/20220908132321_add_voice_text_chat.exs
keianhzo: I saw that other migrations had it disabled so I wasn't sure if there was a reason for it. I've re-enabled it.

bryanenders (approved):

lib/ret/hub.ex
johnshaughnessy: I think `@object_actions` enumerates all the things a user might be allowed to do while `@default_member_permissions` and `@default_restrictive_member_permissions` are specific configurations to apply to users. There's a flag in the admin panel called "permissive rooms" that changes the default permissions applied to users across the instance.

https://github.com/mozilla/reticulum/blob/9967e50e3d970e01f8e121fe4f801cf7dae6cd1d/lib/ret/hub.ex#L200-L206

image

PR #634, opened 2022-10-12
Hub Account Deletion

bryanenders:
Why
---
When a user asks for their account to be deleted
I need to be able to wipe out all their owned data
So I can honor their right to be forgotten.

What
----
* Add DELETE action to account route
* Add `delete_account/2` to `Ret`
* Create data migration to delete orphaned data
* Create schema migration to update references
* Add `stream_data` dev dependency

### DIFF
--- a/lib/ret.ex
+++ b/lib/ret.ex
@@ -1,12 +1,11 @@
 defmodule Ret do
   @moduledoc """
-  Ret keeps the contexts that define your domain
-  and business logic.
-
-  Contexts are also responsible for managing your data, regardless
-  if it comes from the database, an external API or others.
+  The boundary of the Ret context.
   """
-  alias Ret.{Account, Login, Repo}
+  alias Ret.{Account, Asset, Avatar, AvatarListing, Login, OwnedFile, Project, Repo, Scene, SceneListing, Storage}
+  import Canada, only: [can?: 2]
+  import Ret.Schema, only: [is_serial_id: 1]
+  require Ecto.Query
 
   def change_email_for_login(%{old_email: old_email, new_email: new_email}) do
     if not valid_email_address?(new_email) do
@@ -40,4 +39,185 @@ defmodule Ret do
   defp valid_email_address?(string) do
     string =~ ~r/\S+@\S+/
   end
+
+  @doc """
+  Obliterates the account with the given `account_id` and all its owned assets.
+
+  Returns `:ok` if successful, otherwise it returns `{:error, reason}`.
+
+  Error reasons are:
+
+  * `:forbidden` - the current user does not have sufficient authorization or
+    an account could not be found with the given `account_id`
+  * `:failed` - the account data was unable to be deleted for an unknown reason
+
+  """
+  @spec delete_account(Account.id(), Account.t()) :: :ok | {:error, :forbidden | :failed}
+  def delete_account(account_id, %Account{} = current_user) when is_serial_id(account_id) do
+    account_to_delete = Repo.get(Account, account_id)
+
+    cond do
+      !account_to_delete ->
+        {:error, :forbidden}
+
+      not can?(current_user, delete_account(account_to_delete)) ->
+        {:error, :forbidden}
+
+      true ->
+        reassignment = [set: [account_id: current_user.account_id]]
+
+        Ecto.Multi.new()
+        |> Ecto.Multi.delete_all(:delete_assets, asset_query(account_id))
+        |> Ecto.Multi.delete_all(:delete_projects, project_query(account_id))
+        |> delete_avatars_multi(account_id, reassignment)
+        |> delete_scenes_multi(account_id, reassignment)
+        |> delete_owned_files_multi(account_id)
+        |> Ecto.Multi.delete(:delete_account, account_to_delete)
+        |> Repo.transaction()
+        |> case do
+          {:ok, _} ->
+            :ok
+
+          {:error, _} ->
+            {:error, :failed}
+        end
+    end
+  end
+
+  @spec account_owned_file_query(Account.id()) :: Ecto.Query.t()
+  defp account_owned_file_query(account_id) when is_serial_id(account_id),
+    do: Ecto.Query.where(OwnedFile, account_id: ^account_id)
+
+  @spec asset_query(Account.id()) :: Ecto.Query.t()
+  defp asset_query(account_id) when is_serial_id(account_id),
+    do: Ecto.Query.where(Asset, account_id: ^account_id)
+
+  @spec avatar_owned_file_query(Account.id()) :: Ecto.Query.t()
+  defp avatar_owned_file_query(account_id) when is_serial_id(account_id),
+    do:
+      Ecto.Query.from(owned_file in OwnedFile,
+        join: avatar in Avatar,
+        on:
+          owned_file.owned_file_id == avatar.gltf_owned_file_id or
+            owned_file.owned_file_id == avatar.bin_owned_file_id or
+            owned_file.owned_file_id == avatar.thumbnail_owned_file_id or
+            owned_file.owned_file_id == avatar.base_map_owned_file_id or
+            owned_file.owned_file_id == avatar.emissive_map_owned_file_id or
+            owned_file.owned_file_id == avatar.normal_map_owned_file_id or
+            owned_file.owned_file_id == avatar.orm_map_owned_file_id,
+        where: avatar.account_id == ^account_id
+      )
+
+  @spec avatar_query(Account.id()) :: Ecto.Query.t()
+  defp avatar_query(account_id) when is_serial_id(account_id),
+    do: Ecto.Query.where(Avatar, account_id: ^account_id)
+
+  @spec delete_avatars_multi(Ecto.Multi.t(), Account.id(), Keyword.t()) :: Ecto.Multi.t()
+  defp delete_avatars_multi(%Ecto.Multi{} = multi, account_id, reassignment)
+       when is_serial_id(account_id) and is_list(reassignment),
+       do:
+         multi
+         |> Ecto.Multi.update_all(:reassign_avatar_owned_files, avatar_owned_file_query(account_id), reassignment)
+         |> Ecto.Multi.update_all(:reassign_listed_avatars, listed_avatar_query(account_id), reassignment)
+         |> Ecto.Multi.update_all(:reassign_parent_avatars, parent_avatar_query(account_id), reassignment)
+         |> ecto_multi_all(:avatar_owned_files, avatar_owned_file_query(account_id))
+         |> Ecto.Multi.delete_all(:delete_avatars, avatar_query(account_id))
+         |> Ecto.Multi.run(:delete_avatar_owned_files, &delete_owned_files(&2.avatar_owned_files, &1))
+
+  @spec delete_owned_files([OwnedFile.t()], module) :: {:ok, nil} | {:error, Ecto.Changeset.t(OwnedFile.t())}
+  defp delete_owned_files(owned_files, repo) when is_list(owned_files) and is_atom(repo) do
+    Enum.reduce_while(owned_files, {:ok, nil}, fn owned_file, acc ->
+      with {:ok, _} <- OwnedFile.set_inactive(owned_file),
+           Storage.rm_files_for_owned_file(owned_file),
+           {:ok, _} <- repo.delete(owned_file) do
+        {:cont, acc}
+      else
+        error ->
+          {:halt, error}
+      end
+    end)
+  end
+
+  @spec delete_owned_files_multi(Ecto.Multi.t(), Account.id()) :: Ecto.Multi.t()
+  defp delete_owned_files_multi(%Ecto.Multi{} = multi, account_id) when is_serial_id(account_id),
+    do:
+      multi
+      |> ecto_multi_all(:account_owned_files, account_owned_file_query(account_id))
+      |> Ecto.Multi.run(:delete_account_owned_files, &delete_owned_files(&2.account_owned_files, &1))
+
+  @spec delete_scenes_multi(Ecto.Multi.t(), Account.id(), Keyword.t()) :: Ecto.Multi.t()
+  defp delete_scenes_multi(%Ecto.Multi{} = multi, account_id, reassignment)
+       when is_serial_id(account_id) and is_list(reassignment),
+       do:
+         multi
+         |> Ecto.Multi.update_all(:reassign_scene_owned_files, scene_owned_file_query(account_id), reassignment)
+         |> Ecto.Multi.update_all(:reassign_listed_scenes, listed_scene_query(account_id), reassignment)
+         |> Ecto.Multi.update_all(:reassign_parent_scenes, parent_scene_query(account_id), reassignment)
+         |> ecto_multi_all(:scene_owned_files, scene_owned_file_query(account_id))
+         |> Ecto.Multi.delete_all(:delete_scenes, scene_query(account_id))
+         |> Ecto.Multi.run(:delete_scene_owned_files, &delete_owned_files(&2.scene_owned_files, &1))
+
+  # TODO: Replace calls with Ecto.Multi.all/3 after Ecto updgrade
+  @spec ecto_multi_all(Ecto.Multi.t(), atom, Ecto.Query.t()) :: Ecto.Multi.t()
+  defp ecto_multi_all(%Ecto.Multi{} = multi, name, %Ecto.Query{} = query) when is_atom(name) do
+    Ecto.Multi.run(multi, name, fn repo, _changes ->
+      {:ok, repo.all(query)}
+    end)
+  end
+
+  @spec listed_avatar_query(Account.id()) :: Ecto.Query.t()
+  defp listed_avatar_query(account_id) when is_serial_id(account_id),
+    do:
+      Ecto.Query.from(avatar in Avatar,
+        join: listing in AvatarListing,
+        on: avatar.avatar_id == listing.avatar_id,
+        where: avatar.account_id == ^account_id
+      )
+
+  @spec listed_scene_query(Account.id()) :: Ecto.Query.t()
+  defp listed_scene_query(account_id) when is_serial_id(account_id),
+    do:
+      Ecto.Query.from(scene in Scene,
+        join: listing in SceneListing,
+        on: scene.scene_id == listing.scene_id,
+        where: scene.account_id == ^account_id
+      )
+
+  @spec parent_avatar_query(Account.id()) :: Ecto.Query.t()
+  defp parent_avatar_query(account_id) when is_serial_id(account_id),
+    do:
+      Ecto.Query.from(parent_avatar in Avatar,
+        join: child_avatar in Avatar,
+        on: parent_avatar.avatar_id == child_avatar.parent_avatar_id,
+        where: parent_avatar.account_id == ^account_id
+      )
+
+  @spec parent_scene_query(Account.id()) :: Ecto.Query.t()
+  defp parent_scene_query(account_id) when is_serial_id(account_id),
+    do:
+      Ecto.Query.from(parent_scene in Scene,
+        join: child_scene in Scene,
+        on: parent_scene.scene_id == child_scene.parent_scene_id,
+        where: parent_scene.account_id == ^account_id
+      )
+
+  @spec project_query(Account.id()) :: Ecto.Query.t()
+  defp project_query(account_id) when is_serial_id(account_id),
+    do: Ecto.Query.where(Project, created_by_account_id: ^account_id)
+
+  @spec scene_owned_file_query(Account.id()) :: Ecto.Query.t()
+  defp scene_owned_file_query(account_id) when is_serial_id(account_id),
+    do:
+      Ecto.Query.from(owned_file in OwnedFile,
+        join: scene in Scene,
+        on:
+          owned_file.owned_file_id == scene.model_owned_file_id or
+            owned_file.owned_file_id == scene.screenshot_owned_file_id or
+            owned_file.owned_file_id == scene.scene_owned_file_id,
+        where: scene.account_id == ^account_id
+      )
+
+  @spec scene_query(Account.id()) :: Ecto.Query.t()
+  defp scene_query(account_id) when is_serial_id(account_id),
+    do: Ecto.Query.where(Scene, account_id: ^account_id)
 end
--- a/lib/ret/account.ex
+++ b/lib/ret/account.ex
@@ -6,12 +6,13 @@ defmodule Ret.Account do
 
   import Canada, only: [can?: 2]
 
+  @type id :: pos_integer
   @type t :: %__MODULE__{}
+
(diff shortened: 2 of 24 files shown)


bryanenders:
Sorry this is a big one. When it’s merged into https://github.com/mozilla/reticulum/pull/628, the diff against `master` will be cleaner.

lib/ret.ex
```diff
@@ -1,158 +1,306 @@
...
+  * `:not_found` - an account could not be found with the given `id`
+  * `:forbidden` - the user does not have sufficient authorization
+  * `:failed` - the account data was unable to be deleted for an unknown reason
 
-      project_and_scene_owned_files =
-        collect_owned_files(project_query, @project_file_columns) ++
-          collect_owned_files(scene_query, @scene_file_columns)
+  """
+  @spec delete_account(Account.id(), Account.t()) :: :ok | {:error, :not_found | :forbidden | :failed}
+  def delete_account(id, %Account{} = user) when is_serial_id(id) do
```
robinkwilson, line 23: I might call "user" a different variable name in this function call. I had to read this a few times to understand "user" is the person that requests the account deletion. 

Which made it difficult to understand "reassign" meant because first read through it seemed like it was reassigning back to the account that was going to be deleted.

lib/ret.ex
```diff
@@ -1,158 +1,306 @@
...
+      !account_to_delete ->
+        {:error, :not_found}
 
-      delete_owned_files(project_and_scene_owned_files)
+      not can?(user, delete_account(account_to_delete)) ->
+        {:error, :forbidden}
 
-      delete_owned_files(Repo.all(from(o in OwnedFile, where: o.account_id == ^account_to_delete.account_id)))
+      true ->
+        reassignment = [set: [account_id: user.account_id]]
```
robinkwilson, line 34: Trying to understand this syntax: 

Is this an array of tuples `[ {set: [ {account_id: serial_number} ] } ]` OR simply `{Set: {account_id: serial_number }}`

lib/ret.ex
```diff
@@ -1,158 +1,306 @@
...
+        |> Ecto.Multi.update_all(
+          :reassign_avatar_owned_files,
+          avatar_owned_file_assignment_query(id),
+          reassignment
+        )
+        |> Ecto.Multi.update_all(
+          :reassign_parent_avatars,
+          parent_avatar_assignment_query(id),
+          reassignment
+        )
```
robinkwilson, line 61: I don't see a reassign_parent_scenes... Where is that handled for "remixed" project scenes?

robinkwilson: I see reassign_scenes but it doesnt seem to search for remixed scenes like remixed avatars. -- Maybe this isn't necessary by how the "remixing" scenes works. -- I saw a test case that checks this, but I'm not clear in the code where it happens.

priv/repo/migrations/20220915094102_add_foreign_references_and_delete_behaviors_to_tables.exs
```diff
@@ -25,33 +23,35 @@ defmodule Ret.Repo.Migrations.AddForeignReferencesAndDeleteBehaviorToTables do
...
-      execute_drop_foreign_constraint(table, column)
+      drop_foreign_constraint(table, column)
     end
 
     for {table, column, foreign_table, foreign_column, _on_delete_action} <- @missing_on_delete_action do
-      execute_drop_foreign_constraint(table, column)
-      add_foreign_reference(table, column, foreign_table, foreign_column, on_delete: :nothing)
-    end
-  end
+      drop_foreign_constraint(table, column)
```
robinkwilson, line 46: on `drop()` function, shouldn't it do the opposite of `up()` This looks to be the same as `up()`

test/ret/schema_test.exs
```diff
@@ -0,0 +1,56 @@
...
+  describe "is_schema/1" do
+    test "with an Ecto schema" do
+      assert true === Schema.is_schema(%DummySchema{})
+    end
+
+    test "with a non-schema struct" do
+      assert false === Schema.is_schema(%DummyStruct{})
+    end
+
+    property "with a non-struct map" do
```
robinkwilson, line 17: Neat I have not seen property tests / used property tests. Looks useful.

test/ret_test.exs
```diff
@@ -342,210 +237,149 @@ defmodule RetTest do
...
-      assert 0 === count(Scene, target_account)
-      assert 1 === count(Project, other_account)
-      assert 1 === count(Scene, other_account)
+      assert :ok === Ret.delete_account(target_account.account_id, admin_account)
+      assert 0 === count_where(Scene, account_id: target_account.account_id)
+      assert 1 === count_where(Scene, account_id: other_account.account_id)
     end
 
-    test "listed scenes are reassigned to admin account" do
+    test "reassigns listed scenes to admin account" do
```
robinkwilson, line 284: Shouldn't it delete the scene listings and scenes that aren't remixed? Not reassign the scenes.

robinkwilson (commented):
Have a few clarifying comments -- the only thing I saw was that scenes didn't get the same treatment as remixed avatars. Maybe it's covered and it didn't look like that in the code. Did my best to be thorough! This is so exciting to see and a doozy of a feature to integrate. So awesome.

lib/ret.ex
bryanenders: Good call! I renamed `id` to `account_id` and `user` to `current_user`. Does that help?

test/ret_test.exs
bryanenders: @johnshaughnessy @netpro2k Do you have any thoughts?

johnshaughnessy: I'm not sure what the correct behavior is. It was clearly an intentional choice (Brian noted it in the original PR https://github.com/mozilla/reticulum/pull/624 ). I'm trying to understand how scene listings are used, and also reached out to Brian to ask why he implemented things this way.

Since it's not an accident and we have enough information to change the implementation later, I think it's safe to ship before we get to the bottom of it, in case it takes us a while.

johnshaughnessy: Ok I'm ramped up now. This is ultimately a decision that product should make, because it's about "who owns what" and what it means to allow promotion / featuring of scenes. It is related to remixing. Similar questions come up for avatars. Let's discuss options with Elgin.

johnshaughnessy: @bryanenders and I spoke with @emclaren about this topic. Summarizing for @robinkwilson and posterity.

(Disclaimer: I'm not a lawyer.)

Reassigning `scene_listings` (not deleting them) seems to be the desired behavior, since `scene_listings` only exist for scenes that have been marked promote-able, which seems to give Mozilla permission: https://github.com/mozilla/Spoke/blob/master/PROMOTION.md 

Similarly, marking an asset as "remixable" comes with certain terms: https://github.com/mozilla/Spoke/blob/master/REMIXING.md

This means that:
- Alice publishes a scene. Marks it promotable.
- Bob (hub owner) approves / features it, which creates a scene listing.
- Charlie creates a room with that scene listing.
- Alice deletes her account. This deletes the scene, but not the listing. Ownership of the underlying asset (Owned file) is (lazily) transferred to Bob.
- Charlie's room doesn't change. 

We may want to revisit this behavior in the future and make sure users are totally aware of who is allowed to do what with their assets. As Elgin recalled, once-upon-a-time we had these "remixable" and "promotable" options on by default, and we changed the defaults to be disabled to make it less likely to happen on accident. We may consider building a formal mechanism for users to ask the hub owner to un-publish an asset.

We should also make sure the copy is correct for individual hubs. The links above say that _Mozilla_ is given rights to do things, but I believe these capabilities are conferred on _the hub admin_, who is usually _not_ Mozilla.

test/ret/schema_test.exs
```diff
@@ -0,0 +1,56 @@
+defmodule Ret.SchemaTest do
```
netpro2k, line 1: Confused what this test is testing. Seems like it tests DummySchema and DummyStruct, but those only appear to be used in this test. Was this just to test that stream_data was working? Wonder if we should drop the test now that it is.

lib/ret/account.ex
```diff
@@ -6,12 +6,13 @@ defmodule Ret.Account do
...
   import Canada, only: [can?: 2]
 
+  @type id :: pos_integer
   @type t :: %__MODULE__{}
 
-  @schema_prefix "ret0"
-  @primary_key {:account_id, :id, autogenerate: true}
   @account_preloads [:login, :identity]
 
+  @schema_prefix "ret0"
```
netpro2k, line 14: Does swapping the order of these lines actually have an impact or was it just shuffled in the process of doing other things?

netpro2k: Or was it just that you actually have your editor correctly configured with elixir format and whoever edited before did not 😄

test/ret/schema_test.exs
netpro2k: Ah nevermind. It's for testing our schema.ex utilities. Any advantage to using DummySchema/DummyStruct instead of a real one?

test/ret_test.exs
netpro2k: > The links above say that Mozilla is given rights to do things, but I believe these capabilities are conferred on the hub admin, who is usually not Mozilla.

We will need to decide. It might be both for our hosted service.

priv/repo/migrations/20220915094102_add_foreign_references_and_delete_behaviors_to_tables.exs
```diff
@@ -2,15 +2,13 @@ defmodule Ret.Repo.Migrations.AddForeignReferencesAndDeleteBehaviorToTables do
   use Ecto.Migration
```
netpro2k, line 2: Hmm.. Is it ok to edit a migration like this? At first I thought it was just formatting changes but looks like there might be others. Aren't migrations immutable?

priv/repo/migrations/20220820164826_delete_orphaned_data_from_tables_missing_references.exs
```diff
@@ -7,29 +7,25 @@ defmodule Ret.Repo.Migrations.DeleteOrphanedDataFromTablesMissingReferences do
   """
```
netpro2k, line 7: This looks like just a formatting change, yes?

netpro2k (commented):
Didn't realize we had already landed account deletion, so this PR wasn't as scary as I was expecting. 

Mostly LGTM. Made some small comments here and there mostly for my own education. The only major thing that stood out to me is that it looks like we modified a db migration. Is that ok to do? I was under the impression they were immutable. Also the potential security issue I pointed out with not_allowed/not_found.

test/ret_test.exs
```diff
@@ -29,307 +30,201 @@ defmodule RetTest do
...
+      assert 0 === count_where(HubBinding, hub_id: hub.hub_id)
+      assert 0 === count_where(HubInvite, hub_id: hub.hub_id)
+      assert 0 === count_where(HubRoleMembership, hub_id: hub.hub_id)
+      assert 0 === count_where(RoomObject, hub_id: hub.hub_id)
+      assert 0 === count_where(WebPushSubscription, hub_id: hub.hub_id)
     end
 
     test "deletes entities associated with an account, even when they belong to a hub owned by another account" do
       {:ok, admin_account: admin_account} = create_admin_account("test_admin")
       hub_owner = create_account("test_owner")
```
johnshaughnessy, line 93: It is unfortunate that what we call _rooms_, Reticulum calls _hubs_, especially now that `Dash.Hub` exists for the managed product and a `Dash.Hub` is what we mean when we say "a hub".

test/ret_test.exs
```diff
@@ -342,210 +237,149 @@ defmodule RetTest do
...
+      1 = count_where(Scene, account_id: target_account.account_id)
+      1 = count_where(Scene, account_id: other_account.account_id)
 
-      assert 0 === count(Project, target_account)
-      assert 0 === count(Scene, target_account)
-      assert 1 === count(Project, other_account)
-      assert 1 === count(Scene, other_account)
+      assert :ok === Ret.delete_account(target_account.account_id, admin_account)
+      assert 0 === count_where(Scene, account_id: target_account.account_id)
+      assert 1 === count_where(Scene, account_id: other_account.account_id)
```
johnshaughnessy, line 281: It seems that we lost the `Project` checks here. Were they needed to test this functionality?

Also, do we expect the `parent_scene_id` of `other_account`'s scene point to a now non-existent scene, or will it be nilified?

johnshaughnessy (commented):
I only read through `ret_test.exs` for this review so that I could understand the intended behavior of account deletion. I don't have a solid grasp yet on how owned_files, projects, scenes, and parent scenes interact, so I'll need to spend more time reading that code.

lib/ret.ex
```diff
@@ -1,158 +1,306 @@
...
+    account_to_delete = Repo.get(Account, account_id)
 
-      Repo.delete_all(project_query)
-      Repo.delete_all(scene_query)
+    cond do
+      !account_to_delete ->
+        {:error, :not_found}
 
-      delete_owned_files(project_and_scene_owned_files)
+      not can?(current_user, delete_account(account_to_delete)) ->
```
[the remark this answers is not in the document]
bryanenders: Good catch!

lib/ret.ex
```diff
@@ -1,158 +1,306 @@
...
+    do: from(f in OwnedFile, where: f.account_id == ^account_id)
 
-  defp delete_entities_with_owned_files(query, file_columns) when is_list(file_columns) do
-    entities = Repo.all(query) |> Repo.preload(file_columns)
-    entity_owned_files = Enum.flat_map(entities, fn entity -> entity_owned_files(entity, file_columns) end)
-    Repo.delete_all(query)
-    delete_owned_files(entity_owned_files)
-  end
+  @spec asset_owned_file_query(Account.id()) :: Ecto.Query.t()
+  defp asset_owned_file_query(account_id) when is_serial_id(account_id),
```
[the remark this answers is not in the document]
bryanenders: According to the tests, yes all but `avatar_owned_files` are covered by `account_owned_files`. Avatars can be owned by one user and the OwnedFile by another; as for the others, those cases aren’t reflected in the tests.

Why do I refer to the tests? I didn’t map out or write this functionality, Brian P did. I merely refactored and made the behavior atomic.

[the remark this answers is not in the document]
bryanenders: Removed those operation that, according to the tests, are redundant

test/ret/schema_test.exs
bryanenders: Good question. `DummySchema` is in fact a real Ecto schema, and `DummyStruct` is a real struct. By using these dedicated data structures instead of other in-use schemas and structs we decouple this module from other unrelated modules. It would be troubling if a rename to `Avatar` for instance were to require a change to this test. As a fringe benefit we can easily move `Schema` into another project or out into its own library.

lib/ret/account.ex
bryanenders: `@schema_prefix` and `@primary_key` are options read by the code injected via `use Ecto.Schema`, so I put them right next to the `Ecto.Schema` DSL. `@account_preloads` is merely used as a constant so I left it on its own.

To answer your question, the order of those two doesn’t matter, I just oriented them such that the higher-level concern (tenancy) preceded the lower-level one (primary key name).

priv/repo/migrations/20220915094102_add_foreign_references_and_delete_behaviors_to_tables.exs
bryanenders: You are correct, migrations should not be functionally changed. In this case the change is a refactor, the behavior remains the same. More importantly though, this migration hasn’t yet been merged into `master`.

priv/repo/migrations/20220820164826_delete_orphaned_data_from_tables_missing_references.exs
bryanenders: Bingo

test/ret_test.exs
bryanenders: There are some good refactoring opportunities. I’d love to do some [event storming](https://www.eventstorming.com/) and [domain modeling](https://en.wikipedia.org/wiki/Domain-driven_design) and work towards a [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html).

lib/ret.ex
bryanenders: From what I recall Brian P telling me, parent scenes are different from parent avatars. Remixed scenes are copied, whereas remixed avatars are merely referenced. This means we can delete parent scenes, whereas we have to reassign parent avatars.

test/ret_test.exs
bryanenders: These test was testing numerous things. Many `assert` calls were redundant. I’ve broken them up so that each test has only one logical assertion.

priv/repo/migrations/20220908132321_add_voice_text_chat.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Ret.Repo.Migrations.AddFlyPermission do
+defmodule Ret.Repo.Migrations.AddVoiceTextChatPermission do
```
netpro2k, line 1: based on the filename, this seems like a correct (unrelated) change. Just making sure the module name doesn't impact existing applied migrations or rollbacks at all.

lib/ret/hub.ex
```diff
@@ -862,6 +862,14 @@ defimpl Canada.Can, for: Ret.Account do
     false
   end
 
+  def can?(
+        %Ret.Account{is_admin: true} = admin_account,
+        :delete_account,
+        %Ret.Account{is_admin: false} = account_to_delete
```
netpro2k, line 868: Intent here is that you should not be able to delete admin accounts? I guess having to de-admin an account before deleting it adds a nice layer of added prevention from accidental deletion.

lib/ret.ex
```diff
@@ -1,9 +1,207 @@
...
+  require Ecto.Query
+
+  @doc """
+  Obliterates the account with the given `account_id` and all its owned assets.
+
+  Returns `:ok` if successful, otherwise it returns `{:error, reason}`.
+
+  Error reasons are:
+
+  * `:forbidden` - the current user does not have sufficient authorization or
```
netpro2k, line 17: While I do think its right that an unauthenticated user should not be able to find out if an account is valid or not, an authenticated user should be able to differentiate between an auth error and a non existing account.

lib/ret.ex
netpro2k: Ah interesting. Did you poke around some production data to confirm that? seems slightly surprising we would copy them.

lib/ret.ex
```diff
@@ -1,9 +1,207 @@
...
+    account_to_delete = Repo.get(Account, account_id)
+
+    cond do
+      !account_to_delete ->
+        {:error, :forbidden}
+
+      not can?(current_user, delete_account(account_to_delete)) ->
+        {:error, :forbidden}
+
+      true ->
```
netpro2k, line 33: Probably ok, but the fallthrough case being the actual deleting of accounts feels slightly scary to me. Would kind of like `can?(current_user, delete_account(account_to_delete)) ->` to be the case here with the fallthrough being forbidden. Unless this is more idiomatic.

priv/repo/migrations/20220820164826_delete_orphaned_data_from_tables_missing_references.exs
```diff
@@ -0,0 +1,29 @@
+defmodule Ret.Repo.Migrations.DeleteOrphanedDataFromTablesMissingReferences do
+  @moduledoc """
+  This migration prepares the database for the migration that immediately follows it.
+  Here we delete orphaned data, where an administrator might have manually, and partially
+  deleted records that were referenced by these tables. The following migration adds
+  those references to ensure constraints going forward.
+  """
```
netpro2k, line 7: Something weird happened with the PR notes and the review comment below seems to be about another file. I usually try not to rebase on PRs once reviews have happened since it can do screwy things.

In any case, are there any other possible orphaned pieces of data? Like OwnedFiles with invalid or empty account ids?

test/ret_test.exs
```diff
@@ -0,0 +1,681 @@
...
+    test "reassigns parent avatars to current user", %{account_to_delete: account_to_delete, current_user: current_user} do
+      avatar = create_avatar(account_to_delete)
+      account_to_delete_id = account_to_delete.account_id
+      create_child_avatar(avatar)
+
+      %Avatar{account_id: ^account_to_delete_id} = repo_reload(avatar)
+
+      assert :ok === Ret.delete_account(account_to_delete.account_id, current_user)
+      assert avatar = repo_reload(avatar)
+      assert current_user.account_id === avatar.account_id
```
netpro2k, line 137: The idea here is to prevent breaking any avatars down the remix chain from the deleted account, by moving ownership of that over to the admin account that is currently doing the deleting? I guess that makes sense but not sure how it works in terms of compliance with deleting all of a users data... I guess remixing was a specific opt in that allowed us a CC_BY license to the content. https://github.com/mozilla/hubs/blob/master/REMIXING.md

netpro2k (approved):
LGTM. Made some very minor suggestions and asked some questions but nothing that I see needing to hold up the PR.

priv/repo/migrations/20220908132321_add_voice_text_chat.exs
bryanenders: I’m glad you brought it up. The module name doesn’t impact the existing applied migrations. The reason is that those are keyed with the unique version (e.g., `20220908132321`), which happens to be a timestamp, in the `schema_migrations` table managed by Ecto.

lib/ret/hub.ex
bryanenders: Bingo

lib/ret.ex
bryanenders: Fall-throughs for unexpected conditions are generally avoided in Erlang/Elixir. “Let it crash” is the mantra. With a `case` statement that means it’s best not to include a clause for a case you haven’t planned for. With `cond` that means the `true` condition isn’t a last-resort but instead the happy path. The conditions that yielded `{:error, reason}` can be thought of as [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html), a fail-fast technique.

test/ret_test.exs
bryanenders: Correct. As soon as content is made promotable, it becomes public data.

By analogy, if someone were to remove their repo from GitHub, you would be shocked if your fork of that repo were to disappear along with the original! 😱

brianpeiris:
Hello. Noticed this corresponding UI PR was still in my list of open PRs, so I just wanted to remind you that it exists, in case you missed it. https://github.com/mozilla/hubs/pull/5716

bryanenders:
Rebased

lib/ret.ex
bryanenders: @robinkwilson @netpro2k You two were absolutely correct. Pushing up a code fix now.

bryanenders:
Rebased

PR #636, opened 2022-10-19
Add internal API for changing email for Login

johnshaughnessy:
Adds an authenticated API to change a `Login`'s email address. To use, send a `PUT` to `/api-internal/v1/change_email_for_login` with the old and new email addresses and an `x-ret-dashboard-access-key` header.


lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,29 @@
...
+      _err -> send_resp(conn, 500, %{error: :change_login_email_failed} |> Poison.encode!())
+    end
+  end
+
+  defp is_empty_or_whitespace(str) do
+    String.trim(str) === ""
+  end
+
+  defp is_valid_email_address(_str) do
+    # TODO
```
johnshaughnessy, line 26: @bryanenders Is there a "standard" way to validate that a user-provider string is a valid email address?

lib/ret/login.ex
```diff
@@ -22,4 +24,22 @@ defmodule Ret.Login do
...
+    old_identifier_hash = Account.identifier_hash_for_email(old_email)
+    new_identifier_hash = Account.identifier_hash_for_email(new_email)
+
+    login_to_update = Repo.get_by(Login, identifier_hash: old_identifier_hash)
+
+    case login_to_update do
+      %Login{} ->
+        login_to_update
+        |> cast(%{identifier_hash: new_identifier_hash}, [:identifier_hash])
+        |> unique_constraint(:identifier_hash)
```
netpro2k, line 38: Is the idea here that if an account already exists for new_email this will just fail? I suspect we will need to be able to handle this case upstream so we might not want to just return a 500.

johnshaughnessy: Yes, and that's a good point. I added some useful error messages to disambiguate failures:
- `:new_email_already_in_use`
- `:no_account_for_old_email`
- a catch-all for unknown error

https://github.com/mozilla/reticulum/pull/636/commits/0977f6635632cb2cdd1831fce6c74327f5adf847

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -11,10 +11,18 @@ defmodule RetWeb.ApiInternal.V1.ChangeLoginEmailController do
...
            true <- is_valid_email_address(old_email),
            true <- is_valid_email_address(new_email),
-         {:ok, _} <- Login.update_identifier_hash(%{old_email: old_email, new_email: new_email}) do
+         :ok <- Login.update_identifier_hash(%{old_email: old_email, new_email: new_email}) do
       send_resp(conn, 200, %{success: true} |> Poison.encode!())
     else
-      _err -> send_resp(conn, 500, %{error: :change_login_email_failed} |> Poison.encode!())
+
+      {:error, :new_email_already_in_use} ->
+        send_resp(conn, 409, %{error: :new_email_already_in_use} |> Poison.encode!())
```
johnshaughnessy, line 19: Is `409` an appropriate error code? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

Unrelated, TIL `418` https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
bryanenders:
```elixir
email =~ ~r/(?:(?:\r\n)?[ \t])*(?:(?:(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*|(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)*\<(?:(?:\r\n)?[ \t])*(?:@(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*(?:,@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*)*:(?:(?:\r\n)?[ \t])*)?(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*\>(?:(?:\r\n)?[ \t])*)|(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)*:(?:(?:\r\n)?[ \t])*(?:(?:(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*|(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)*\<(?:(?:\r\n)?[ \t])*(?:@(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*(?:,@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*)*:(?:(?:\r\n)?[ \t])*)?(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*\>(?:(?:\r\n)?[ \t])*)(?:,\s*(?:(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*|(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)*\<(?:(?:\r\n)?[ \t])*(?:@(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*(?:,@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*)*:(?:(?:\r\n)?[ \t])*)?(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\](?:(?:\r\n)?[ \t])*))*\>(?:(?:\r\n)?[ \t])*))*)?;\s*)/
```
or

```elixir
email =~ ~r/\S+@\S+/
```

RFCs 822, 2822, and 5322 are more permissive than most programmers realize, and it’s better to err on the side of letting an invalid email through than keeping a valid email out.

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
     |> unique_constraint(:identifier_hash)
     |> put_assoc(:account, account)
   end
+
+  def update_identifier_hash(%{old_email: old_email, new_email: new_email} = _) do
```
bryanenders, line 28: This isn’t the right module for this function.

Additionally the `= _` is needless.

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+    old_identifier_hash = Account.identifier_hash_for_email(old_email)
+    new_identifier_hash = Account.identifier_hash_for_email(new_email)
```
bryanenders, line 30: Consider inlining both of these variables. They aren’t reused and the names don’t aid understanding.

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+    login = Repo.get_by(Login, identifier_hash: old_identifier_hash)
+    case update_identifier_hash_for_login(login, new_identifier_hash) do
+      {:error, :no_login} -> {:error, :no_account_for_old_email}
```
bryanenders, line 34: Why not let `update_identifier_hash_for_login/2`, which is only used right here, return the right value?

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+      {:error, :new_identifier_hash_is_not_unique} -> {:error, :new_email_already_in_use}
```
bryanenders, line 35: Ditto

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+      {:error, _} -> {:error, :failed_to_update_login}
```
bryanenders, line 36: Ditto

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+      _ -> :ok
```
bryanenders, line 37: Ditto

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+    end
+  end
+
+  defp update_identifier_hash_for_login(nil, new_identifier_hash) do
```
bryanenders: `_new_identifier_hash`

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+    {:error, :no_login}
+  end
+
+  defp update_identifier_hash_for_login(%Login{} = login, new_identifier_hash) do
+    result = login
+    |> cast(%{identifier_hash: new_identifier_hash}, [:identifier_hash])
+    |> unique_constraint(:identifier_hash)
+    |> Ret.Repo.update()
```
bryanenders: `Ret.Repo` has been aliased, so the `Ret.` bit can be dropped

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+    end
+  end
+
+  defp update_identifier_hash_for_login(nil, new_identifier_hash) do
+    {:error, :no_login}
+  end
+
+  defp update_identifier_hash_for_login(%Login{} = login, new_identifier_hash) do
+    result = login
+    |> cast(%{identifier_hash: new_identifier_hash}, [:identifier_hash])
```
bryanenders: `cast/4` is for handling a full set of params. This can be simplified to:
```elixir
|> change(identifier_hash: new_identifier_has)
```

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+      _ -> :ok
+    end
+  end
+
+  defp update_identifier_hash_for_login(nil, new_identifier_hash) do
+    {:error, :no_login}
+  end
+
+  defp update_identifier_hash_for_login(%Login{} = login, new_identifier_hash) do
+    result = login
```
bryanenders: It looks like this needs to be `mix format`ted

lib/ret/login.ex
```diff
@@ -22,4 +24,33 @@ defmodule Ret.Login do
...
+    |> cast(%{identifier_hash: new_identifier_hash}, [:identifier_hash])
+    |> unique_constraint(:identifier_hash)
+    |> Ret.Repo.update()
+
+    case result do
```
bryanenders, line 52: Here’s a fun Elixir trick: You can pipe into `case`.

```elixir
...
|> Repo.update()
|> case do
  ...
end
```

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+
+      {:error, :no_account_for_old_email } ->
+        send_resp(conn, 409, %{error: :no_account_for_old_email} |> Poison.encode!())
+
+      {:error, _ } ->
+        send_resp(conn, 500, %{error: :failed_to_update_email} |> Poison.encode!())
+    end
+  end
+
+  defp is_empty_or_whitespace(str) do
```
bryanenders, line 30: The Elixir convention is to name boolean functions, macros, and variables with a `?` suffix instead of an `is_` prefix _unless it is a guard clause (and thus can be inlined by the compiler)_.

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+    String.trim(str) === ""
+  end
+
+  defp is_valid_email_address(_str) do
```
bryanenders: Ditto

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+  use RetWeb, :controller
+
+  alias Ret.Login
+
+  def post(conn, %{"old_email" => old_email, "new_email" => new_email})
+  when is_binary(old_email) and is_binary(new_email) do
+    conn = put_resp_header(conn, "content-type", "application/json")
+
+    with false <- is_empty_or_whitespace(old_email),
+         false <- is_empty_or_whitespace(new_email),
```
bryanenders, line 11: If it’s valid, then it won’t be empty or whitespace. These `is_empty_or_whitespace/1` calls can be removed.

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+           true <- is_valid_email_address(old_email),
+           true <- is_valid_email_address(new_email),
```
bryanenders: This seems more like business logic than web logic. I’d consider moving this validation to the business layer.

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
johnshaughnessy: The "user" in this case is just _us_, so I don't expect this to ever fail... but I figured it wouldn't hurt to validate the format.

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
+defmodule RetWeb.ApiInternal.V1.ChangeLoginEmailController do
+  use RetWeb, :controller
+
+  alias Ret.Login
+
+  def post(conn, %{"old_email" => old_email, "new_email" => new_email})
+  when is_binary(old_email) and is_binary(new_email) do
+    conn = put_resp_header(conn, "content-type", "application/json")
```
bryanenders, line 8: You probably want `json/2` rather than `put_resp_header/3`+`send_resp/3`+`Poison.encode/1`.

On a tangential note, prefer Jason to Poison.

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+      _ ->
+        send_resp(conn, 400, %{error: :invalid_parameters} |> Poison.encode!())
+    end
+  end
+
+  defp is_empty_or_whitespace(str) do
+    String.trim(str) === ""
+  end
+
+  defp is_valid_email_address(str) do
```
bryanenders, line 34: nit: Abbreviations such as `str` are unidiomatic in Elixir

lib/ret_web/router.ex
```diff
@@ -202,6 +202,7 @@ defmodule RetWeb.Router do
       get("/presence/range_max", ApiInternal.V1.PresenceController, :range_max)
       get("/storage", ApiInternal.V1.StorageController, :show)
       post("/rewrite_assets", ApiInternal.V1.RewriteAssetsController, :post)
+      post("/change_email_for_login", ApiInternal.V1.ChangeLoginEmailController, :post)
```
bryanenders, line 205: From a REST perspective, wouldn’t this be an update? Consider the following:

```elixir
 put("/change_email_for_login", ApiInternal.V1.LoginEmailController, :update)
```

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+         false <- is_empty_or_whitespace(new_email),
+         true <- is_valid_email_address(new_email),
+         :ok <- Login.update_identifier_hash(%{old_email: old_email, new_email: new_email}) do
+      send_resp(conn, 200, %{success: true} |> Poison.encode!())
+    else
+      {:error, :new_email_already_in_use} ->
+        send_resp(conn, 409, %{error: :new_email_already_in_use} |> Poison.encode!())
+
+      {:error, :no_account_for_old_email} ->
+        send_resp(conn, 409, %{error: :no_account_for_old_email} |> Poison.encode!())
```
bryanenders, line 20: This would be a 404 I believe

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
+defmodule RetWeb.ApiInternal.V1.ChangeLoginEmailControllerTest do
+  use RetWeb.ConnCase
+  import Ret.TestHelpers
```
bryanenders, line 3: nit: Consider separating `use` from `alias`, `import`, and `require`. The former injects code, the latter are just types of imports.

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  alias Ret.Account
+
+  @dashboard_access_header "x-ret-dashboard-access-key"
+  @dashboard_access_key "test-key"
+
+  setup_all do
+    merge_module_config(:ret, RetWeb.Plugs.DashboardHeaderAuthorization, %{dashboard_access_key: @dashboard_access_key})
+
+    on_exit(fn ->
+      Ret.TestHelpers.merge_module_config(:ret, RetWeb.Plugs.DashboardHeaderAuthorization, %{dashboard_access_key: nil})
```
bryanenders, line 13: Since `Ret.TestHelpers` was imported, the `Ret.TestHelpers.` bit can be dropped

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  test "email addresses validation only applies to new emails", %{conn: conn} do
+    Account.find_or_create_account_for_email("not_an_email_address")
+    assert Account.exists_for_email?("not_an_email_address")
+    assert %{status: 200} = post_change_email_for_login(conn, "alice@reticulum.io", "not_an_email_address")
+    assert Account.exists_for_email?("alice@reticulum.io")
+    refute Account.exists_for_email?("not_an_email_address")
+  end
+
+  test "account emails can be changed", %{conn: conn} do
+    refute Account.exists_for_email?("alice@reticulum.io")
```
bryanenders, line 34: Don’t verify the preconditions of your setup, only the preconditions of the function under test. Cut this line.

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+
+  setup_all do
+    merge_module_config(:ret, RetWeb.Plugs.DashboardHeaderAuthorization, %{dashboard_access_key: @dashboard_access_key})
+
+    on_exit(fn ->
+      Ret.TestHelpers.merge_module_config(:ret, RetWeb.Plugs.DashboardHeaderAuthorization, %{dashboard_access_key: nil})
+    end)
+  end
+
+  test "new email addresses must be valid", %{conn: conn} do
```
bryanenders, line 17: Consider describing the function under test. In controller tests I like to wrap tests for a given action in a describe like so:

```elixir
describe "PUT update" do
  ...
end
```

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+    Account.find_or_create_account_for_email("alice@reticulum.io")
+    assert Account.exists_for_email?("alice@reticulum.io")
+    assert %{status: 400} = post_change_email_for_login(conn, "not_an_email_address", "alice@reticulum.io")
+    refute Account.exists_for_email?("not_an_email_address")
+    assert Account.exists_for_email?("alice@reticulum.io")
+  end
+
+  test "email addresses validation only applies to new emails", %{conn: conn} do
+    Account.find_or_create_account_for_email("not_an_email_address")
+    assert Account.exists_for_email?("not_an_email_address")
```
bryanenders, line 27: Is this precondition worth verifying? It’s just testing the line before it.

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+    merge_module_config(:ret, RetWeb.Plugs.DashboardHeaderAuthorization, %{dashboard_access_key: @dashboard_access_key})
+
+    on_exit(fn ->
+      Ret.TestHelpers.merge_module_config(:ret, RetWeb.Plugs.DashboardHeaderAuthorization, %{dashboard_access_key: nil})
+    end)
+  end
+
+  test "new email addresses must be valid", %{conn: conn} do
+    Account.find_or_create_account_for_email("alice@reticulum.io")
+    assert Account.exists_for_email?("alice@reticulum.io")
```
bryanenders, line 19: To make it clear that your are verifying the preconditions and not the post-conditions of the function under test, avoid `refute` and `assert` and just raise.

E.g.
```elixir
true = Account.exists_for_email?("alice@reticulum.io")
```

That being said, is this precondition worth verifying? It’s just testing the line before it.

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+    assert Account.exists_for_email?("not_an_email_address")
+    assert %{status: 200} = post_change_email_for_login(conn, "alice@reticulum.io", "not_an_email_address")
+    assert Account.exists_for_email?("alice@reticulum.io")
+    refute Account.exists_for_email?("not_an_email_address")
+  end
+
+  test "account emails can be changed", %{conn: conn} do
+    refute Account.exists_for_email?("alice@reticulum.io")
+    Account.find_or_create_account_for_email("alice@reticulum.io")
+    assert Account.exists_for_email?("alice@reticulum.io")
```
bryanenders, line 36: Is this precondition worth verifying? It’s just testing the line before it.

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+    refute Account.exists_for_email?("alicia@anotherdomain.com")
```
bryanenders, line 37: Raise instead of refute

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  test "emails cannot be shared between multiple accounts", %{conn: conn} do
+    Account.find_or_create_account_for_email("alice@reticulum.io")
+    Account.find_or_create_account_for_email("bob@reticulum.io")
+    assert %{status: 409} = post_change_email_for_login(conn, "bob@reticulum.io", "alice@reticulum.io")
+  end
+
+  test "email changes are rejected if old_email is not associated with an account", %{conn: conn} do
+    assert %{status: 409} = post_change_email_for_login(conn, "bob@reticulum.io", "alice@reticulum.io")
+    Account.find_or_create_account_for_email("bob@reticulum.io")
+    assert %{status: 409} = post_change_email_for_login(conn, "bob@reticulum.io", "alice@reticulum.io")
```
bryanenders, line 52: `exercise-assert-exercise-assert` is a testing antipattern

bryanenders (changes requested):
A few suggestions

lib/ret/login.ex
johnshaughnessy: Where does it belong? I don't have a good sense of code organization in elixir apps

lib/ret/login.ex
johnshaughnessy: :exploding_head:

lib/ret/login.ex
bryanenders: A useful pattern for managing state is: functional core, imperative shell. Struct manipulation is low-level, which make stateful calls (e.g., `Repo` calls) within a struct module a design smell.

How to fix? Reserve this module for pure functions that operate on `%Ret.Login{}` and put any new repo calls in a module outside this one (e.g., a context boundary module).

johnshaughnessy: Resolved in https://github.com/mozilla/reticulum/pull/636/commits/eee56865456bf34ae00958478c1ed8e15a49007b

lib/ret/login.ex
johnshaughnessy: Resolved in https://github.com/mozilla/reticulum/commit/eee56865456bf34ae00958478c1ed8e15a49007b

lib/ret/login.ex
johnshaughnessy: Resolved in https://github.com/mozilla/reticulum/commit/eee56865456bf34ae00958478c1ed8e15a49007b

lib/ret/login.ex
johnshaughnessy: Resolved in https://github.com/mozilla/reticulum/pull/636/commits/733d0dabda589d475c944c66b7c5a03c34529820

lib/ret/login.ex
johnshaughnessy: Resolved in eee56865456bf34ae00958478c1ed8e15a49007b

lib/ret/login.ex
johnshaughnessy: Resolved in 7b4ec57db4bea5e0a057382c5ed463592cd93da1

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
johnshaughnessy: 6194c80a33edbef2bc63aa5c56f1e35df1e19cac

lib/ret_web/controllers/api-internal/v1/change_login_email_controller.ex
johnshaughnessy: 6194c80a33edbef2bc63aa5c56f1e35df1e19cac

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
johnshaughnessy: Makes sense. I don't have a lot of practice writing tests, so advice like this is really helpful. Thank you!

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
johnshaughnessy: I removed the last two lines, but don't understand the comment / antipattern. Please explain.

bryanenders (changes requested):
This is looking good. Almost there!

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
bryanenders: Happily! Each unit test is comprised of three stages:
1. Given…
2. when…
3. then…

Or put in the imperative mood:
1. Arrange.
2. Act.
3. Assert.

A unit test is a formal experiment and, as with any experiment, there is an inverse relationship between the number of variables and the reliability of the findings. By limiting our test to these three discrete stages we design the most conclusive test possible. Moreover, tests are the most reliable form of documentation and an experiment in three parts is eminently more readable than one with repeat stages.

If I arrange, act, assert, act, and assert again it leaves the question, “What is it that I am actually testing?”

When I write a test in Elixir I like to highlight the flow by diving the _Given…_ from the _when…_ and _then…_ with a blank line. The `when…` is always in the first line after the blank line.

**E.g.**
```elixir
do_some_setup()
foo = get_a_foo_i_need()

assert {:ok, bar} = the_behavior_under_test(foo)
assert something_about_bar(bar)
assert something_about_foo(foo)
```

lib/ret_web/controllers/api-internal/v1/login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
...
+        |> json(%{error: reason})
+    end
+  end
+
+  defp status_code_for_error(:new_email_already_in_use) do
+    Status.code(:conflict)
+  end
+
+  defp status_code_for_error(:no_account_for_old_email) do
+    Status.code(:not_found)
```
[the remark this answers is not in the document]
johnshaughnessy: Unrelated question... Was it silly for me to write these as a bunch of functions? All I wanted was to look up a value in a map.
```elixir
status_code_for_error()[my_error]
```
with
```elixir
  defp status_code_for_error() do
    [
      new_email_already_in_use: :conflict,
      no_account_for_old_email: :not_found,
      invalid_parameters: :bad_request,
      failed_to_update_login: :internal_server_error
    ]
  end
```
or a module attribute
```
@status_code_for_error ...
```

lib/ret_web/controllers/api-internal/v1/login_email_controller.ex
```diff
@@ -0,0 +1,37 @@
+defmodule RetWeb.ApiInternal.V1.LoginEmailController do
+  use RetWeb, :controller
+
+  alias Ret.ChangeEmailForLogin
+  alias Plug.Conn.Status
+
+  def update(conn, %{"old_email" => old_email, "new_email" => new_email})
+      when is_binary(old_email) and is_binary(new_email) do
```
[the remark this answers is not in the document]
johnshaughnessy: Removed!

lib/ret/change_email_for_login.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Ret.ChangeEmailForLogin do
```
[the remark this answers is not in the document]
johnshaughnessy: I did the suggestion but don't understand the thinking behind it. Is the idea that modules should exist "above" what they refers to? 

In this case, `change_email_for_login` uses `Ret.Account`, `Ret.Login`, and `Ret.Repo`... So we put it in `Ret`? (Importantly: not a sibling module `Ret.Foo` or unrelated module `Bar.Baz`)

test/ret_web/controllers/api-internal/change_login_email_controller_test.exs
johnshaughnessy: Gotcha. Thanks!

lib/ret_web/controllers/api-internal/v1/login_email_controller.ex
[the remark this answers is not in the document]
bryanenders: Function clauses are more performant than map or keyword list lookups and are therefore use more often. You followed the idiom. In fact, one bit of metaprogramming that is pretty common in Elixir looks like this:
```elixir
name_by_number = %{
  1 => "one",
  2 => "two",
  3 => "three",
  4 => "four",
  ...
}

@spec name(pos_integer) :: String.t()
for {number, name} <- name_by number do
  def name(unquote(number)),
    do: unquote(name)
end
```

Basically the compiler turns an associative data structure, in this case a map, into function clauses.

lib/ret/change_email_for_login.ex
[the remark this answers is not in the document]
bryanenders: Over time we may discover the line between different domains within the code (e.g., accounting, customer relationships, content management). If we try to split before those lines become clear, we could end up breaking cohesion. With `Ret` as the boundary module, it means outside actors (e.g., `RetWeb` modules) may call `Ret`, but nothing nested under it.

lib/ret_web/controllers/api-internal/v1/login_email_controller.ex
```diff
@@ -0,0 +1,23 @@
...
+    end
+  end
+
+  defp status_code_for_error() do
+    [
+      new_email_already_in_use: :conflict,
+      no_account_for_old_email: :not_found,
+      invalid_parameters: :bad_request
+    ]
+  end
```
bryanenders, line 22: Your original was idiomatic; it’s just that the `Status.code/1` call wasn’t necessary. You can just do this:
```elixir
defp status_code(:new_email_already_in_use),
  do: :conflict

defp status_code(:no_account_for_old_email),
  do: :not_found

defp status_code(:invalid_parameters),
  do: :bad_request

defp status_code(_error_reason),
  do: :internal_server_error

```

lib/ret_web/controllers/api-internal/v1/login_email_controller.ex
```diff
@@ -0,0 +1,23 @@
...
+  use RetWeb, :controller
+
+  def update(conn, %{"old_email" => old_email, "new_email" => new_email}) do
+    case Ret.change_email_for_login(%{old_email: old_email, new_email: new_email}) do
+      :ok ->
+        json(conn, %{success: true})
+
+      {:error, reason} ->
+        conn
+        |> put_status(status_code_for_error()[reason] || :internal_server_error)
```
bryanenders, line 11:
```suggestion
        |> put_status(status_code(reason))
```

bryanenders (approved):
One minor change would make this more idiomatic. Not a show-stopper though!

PR #638, opened 2022-10-28
Fix non-mods unable to chat in bound rooms

netpro2k:
Currently in Discord bound rooms only moderators can text/voice chat, regardless of the room permission setting. This clause will always match when checking for voice/text chat permission and only checks if a user is a moderator. We want this check to fall through to the next clause where moderators are always allowed, and normal users are checked against the room permission.

Testing this locally is pretty tricky so I haven't actually tested this fix...


keianhzo (approved):
LGTM

PR #642, opened 2022-11-17
adding internal jwt generation endpoint for turkeyorch ops

tanfarming:
note: `/api-internal` endpoints are unreachable from public internet on turkey clusters, they also requires an turkey cluster scoped api key

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -60,6 +60,10 @@ config :peerage, log_results: false
 
 config :statix, prefix: "ret"
 
+config :ret,
+  ecto_repos: [Ret.Repo, Ret.SessionLockRepo],
+  turkey_mode?: System.get_env("TURKEY_MODE") === "1"
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{Mix.env()}.exs"
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,10 +7,6 @@ use Mix.Config
 
 # General application configuration
 
-config :ret,
-  ecto_repos: [Ret.Repo, Ret.SessionLockRepo],
-  turkey_mode?: System.get_env("TURKEY_MODE") === "1"
-
 config :ret, RetWeb.Plugs.PostgrestProxy, hostname: System.get_env("POSTGREST_INTERNAL_HOSTNAME") || "localhost"
 
 config :phoenix, :format_encoders, "json-api": Jason
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,6 +7,9 @@ use Mix.Config
 
 # General application configuration
 
+config :ret,
+  ecto_repos: [Ret.Repo, Ret.SessionLockRepo]
+
 config :ret, RetWeb.Plugs.PostgrestProxy, hostname: System.get_env("POSTGREST_INTERNAL_HOSTNAME") || "localhost"
 
 config :phoenix, :format_encoders, "json-api": Jason
--- a/config/config.exs
+++ b/config/config.exs
@@ -6,7 +6,10 @@
 use Mix.Config
 
 # General application configuration
-config :ret, ecto_repos: [Ret.Repo, Ret.SessionLockRepo]
+
+config :ret,
+  ecto_repos: [Ret.Repo, Ret.SessionLockRepo],
+  turkey_mode?: System.get_env("TURKEY_MODE") === "1"
 
 config :ret, RetWeb.Plugs.PostgrestProxy, hostname: System.get_env("POSTGREST_INTERNAL_HOSTNAME") || "localhost"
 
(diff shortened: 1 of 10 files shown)


lib/ret_web/controllers/api-internal/v1/AuthTokenController.ex
```diff
@@ -0,0 +1,18 @@
+defmodule RetWeb.ApiInternal.V1.AuthTokenController do
+  require Logger
+
+  use RetWeb, :controller
+
+  def post(conn, %{"email" => email}) when is_binary(email) do
+    {:ok, token, _params} =
+      if !System.get_env("TURKEY_MODE") do
```
bryanenders, line 8: It’s best-practice to quarantine all `System.get_env/1` calls.
```suggestion
      if not Ret.turkey_mode?() do
```

**lib/ret/ret.ex**
```elixir
@spec turkey_mode? :: boolean
def turkey_mode?,
  do: Application.fetch_env!(:ret, :turkey_mode?)
```

**config/config.exs**
```elixir
config :ret,
  ecto_repos: [Ret.Repo, Ret.SessionLockRepo],
  turkey_mode?: System.get_env("TURKEY_MODE") === "1"
```

keianhzo (approved):

bryanenders (changes requested):

lib/ret_web/controllers/api-internal/v1/AuthTokenController.ex
tanfarming: thx for the code lol
commited

bryanenders: No problem! You’ll want to make sure to update `ret.ex` and `config.exs` alongside `AuthTokenController.ex`.

config/config.exs
```diff
@@ -60,6 +60,10 @@ config :peerage, log_results: false
 
 config :statix, prefix: "ret"
 
+config :ret,
+  ecto_repos: [Ret.Repo, Ret.SessionLockRepo],
+  turkey_mode?: System.get_env("TURKEY_MODE") === "1"
+
```
bryanenders, line 66: Could you move this up to replace line 9? While this technically works, it overwrites the effects of line 9 and could be a pain-in-the-butt to find this far down.

lib/ret.ex
```diff
@@ -40,4 +40,8 @@ defmodule Ret do
   defp valid_email_address?(string) do
     string =~ ~r/\S+@\S+/
   end
+
+  @spec turkey_mode? :: boolean
+  def turkey_mode?,
+    do: Application.fetch_env!(:ret, :turkey_mode?)
```
bryanenders, line 46: This is a nitpick, but would you mind moving this function before `valid_email_address?/1` so these public functions are in alphabetical order?

config/config.exs
tanfarming: done

lib/ret.ex
tanfarming: done

bryanenders (approved):

lib/ret.ex
bryanenders: Thank you!

bryanenders:
The lint-and-test failure should be solved with a quick **$** `mix format`

tanfarming:
> The lint-and-test failure should be solved with a quick **$** `mix format`

yeah i ran that but my elixir version was v1.12 where our linter agent's on v1.8
i downgraded mine to 1.10 (oldest still available) and it worked
so there's some breaking changes on their linter somewhere around v1.11
i've also experienced a bunch of breaking changes when i was trying to move us out of distillery with latest elixir versions (was trying on v1.14) -- do you know if this is a pattern with elixir? if so maybe we want to sit on our current version for as long as possible, given that we are now using the cloudflare worker proxy across HC and turkey and are going to redeisgn ret's file system, we could potentially wait longer / jump to a higher version

tanfarming:
ok this `quarantine all System.get_env/1 call` broke the feature lol
i'm guessing the evaluation's happening on build/compile time rather than runtime?
@bryanenders

bryanenders:
There was some churn when the Elixir formatter was new. I haven’t seen changes for awhile, so I don’t expect that to be much of an issue after we upgrade.

tanfarming:
> @tanfarming That makes sense. With distillery all the app config is being baked at compile time
yeah i forogt that runtime.exs wasn't a thing wtih elixir before v1.10, that's actually the reason why distillery's used

> until we can move to Elixir releases.
lets get togheter after EA and see if we can finally get this done. It's a pretty messy job, inaddition to fighting through the dependency hell, update ret codes against elixir's breaking changes, both HC and turkey's config and infra will need to be updated/redesigned against this change

config/config.exs
```diff
@@ -6,7 +6,10 @@
 use Mix.Config
 
 # General application configuration
-config :ret, ecto_repos: [Ret.Repo, Ret.SessionLockRepo]
+
+config :ret,
+  ecto_repos: [Ret.Repo, Ret.SessionLockRepo],
+  turkey_mode?: System.get_env("TURKEY_MODE") === "1"
```
bryanenders, line 12: You’ll want to revert this as well

lib/ret.ex
```diff
@@ -37,6 +37,10 @@ defmodule Ret do
     end
   end
 
+  @spec turkey_mode? :: boolean
+  def turkey_mode?,
+    do: Application.fetch_env!(:ret, :turkey_mode?)
+
```
bryanenders, line 43: And this

bryanenders: Sorry for the hassle!

PR #645, opened 2022-11-23
Standardize formatting

bryanenders:
Why
---

When I move between Elixir projects, I shouldn’t have to adjust the line length in my editor.  Though the `:line_length` `mix format` option exists, most projects forgo it for the aforementioned reason.

Ecto provides a DSL for building schemas, queries, and migrations. The formatter doesn’t know about this DSL and thus formats the keywords as ordinary function/macro calls.

What
----

* Use default formatter line-length
* Import Ecto formatter config
* Import Ecto SQL formatter config

Note
----

While there are technically a lot of line changes, the fundamental change is a mere two lines in [`.formatter.exs`](https://github.com/mozilla/reticulum/pull/645/files#diff-aad92a030ac915c80a2e0d0204d240fc952893bc87b372dc1b6d485609f27260).  Everything else is just application of the formatting rules.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,4 +1,4 @@
 [
-  import_deps: [:phoenix, :stream_data],
+  import_deps: [:ecto, :phoenix, :stream_data],
   inputs: ["*.{ex,exs}", "{config,lib,priv,test}/**/*.{ex,exs}"]
 ]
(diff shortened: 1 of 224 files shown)


bryanenders:
The fundamental change is in [`.formatter.exs`](https://github.com/mozilla/reticulum/pull/645/files#diff-aad92a030ac915c80a2e0d0204d240fc952893bc87b372dc1b6d485609f27260). Everything else is just application of the formatting rules.

bryanenders:
Rebased

johnshaughnessy (approved):
lgtm. Probably some more deps we can import.

PR #647, opened 2022-11-30
Fix GitHub lint-and-test action

bryanenders:
Why
---

Our `ubuntu-latest` changed from `ubuntu-20.04` to `ubuntu-22.04` which
does not support OTP 22

What
----

Pin to lint-and-test to Ubuntu 20.04


tanfarming:
👍

tanfarming (approved):

PR #657, opened 2023-01-05
Fix missing stats partitions

bryanenders:
Why
---
The `node_stats` table was only partitioned through 2022.  While the `session_stats` table was partitioned through 2030, an earlier released version of the migration only partitioned the table through 2022.

What
----
* Create migration to create `node_stats` partitions through 2030
* Create migration to create `session_stats` partitions through 2030
* Fix `StreamData.TooManyDuplicatesError` intermittent test failure

Note
----
If any partitions have already been created, they will be skipped during partition creation

### DIFF
--- a/priv/repo/migrations/20230105144400_create_node_stats_partitions_through_2030.exs
+++ b/priv/repo/migrations/20230105144400_create_node_stats_partitions_through_2030.exs
@@ -1,0 +1,28 @@
+defmodule Ret.Repo.Migrations.CreateNodeStatsPartitionsThrough2030 do
+  use Ecto.Migration
+
+  import Ret.Repo.MigrationHelpers, only: [next_month: 2]
+
+  @max_year 2030
+
+  def up do
+    for y <- 2023..@max_year,
+        m <- 1..12 do
+      next_month = next_month(y, m)
+
+      execute """
+      CREATE TABLE IF NOT EXISTS ret0.node_stats_y#{y}_m#{m} PARTITION OF ret0.node_stats
+        FOR VALUES FROM ('#{y}-#{m}-01') TO ('#{next_month.y}-#{next_month.m}-01')
+      """
+    end
+  end
+
+  def down do
+    for y <- 2023..@max_year,
+        m <- 1..12 do
+      execute """
+      DROP TABLE ret0.node_stats_y#{y}_m#{m}
+      """
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


priv/repo/migrations/20230105144400_create_node_stats_partitions_through_2030.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Ret.Repo.Migrations.CreateNodeStatsPartitionsThrough2030 do
+  use Ecto.Migration
+
+  import Ret.Repo.MigrationHelpers, only: [next_month: 2]
+
+  @max_year 2030
+
+  def up do
```
johnshaughnessy, line 8: Do we need to copy the rest of the table definition like the original migration? https://github.com/mozilla/reticulum/blob/0f94ced4785f1b24e2236d4c3fbe797d2e8e7f38/priv/repo/migrations/20180412225707_create_node_stats_table.exs#L7-L15

priv/repo/migrations/20230105144433_create_session_stats_partitions_through_2030.exs
```diff
@@ -0,0 +1,28 @@
...
+
+  @max_year 2030
+
+  def up do
+    for y <- 2023..@max_year,
+        m <- 1..12 do
+      next_month = next_month(y, m)
+
+      execute """
+      CREATE TABLE IF NOT EXISTS ret0.session_stats_y#{y}_m#{m} PARTITION OF ret0.session_stats
```
johnshaughnessy, line 14: Does the original migration already cover us up to 2030? https://github.com/mozilla/reticulum/blob/0f94ced4785f1b24e2236d4c3fbe797d2e8e7f38/priv/repo/migrations/20180408231444_create_session_stats_table.exs#L4

priv/repo/migrations/20230105144400_create_node_stats_partitions_through_2030.exs
bryanenders: Nope, the logical table already exists so all we have to do is create the partitions

priv/repo/migrations/20230105144433_create_session_stats_partitions_through_2030.exs
bryanenders: This is precisely why changes should never be made to the logic of an existing migration after that migration has been deployed

johnshaughnessy: Ooh, I see. Thanks!

johnshaughnessy (approved):

PR #662, opened 2023-01-20
Entity state api

johnshaughnessy:
Add `Entities` and `SubEntities`. 

Each `Entity` is composed of:
- A `create_message`, which tells clients how to spawn the entity (hierarchy) in the game world, and
- Many `SubEntities`, which tell the client how to configure the components of spawned sub-entities.

Four `Entity`-related apis are added to the `HubChannel`:

| name                  | permission    | notes                                                     |
| --------------        | ---------     | ----------                                                |
| `list_entities`  | none          |                                                           |
| `save_entity_state` | `pin_objects` | Fails (gracefully) if an `Entity` (for the given `nid`) already exists |
| `update_entity_state` | `pin_objects` | Fails (gracefully) if an `Entity` (for the given `root_nid`) does not already exist          |
| `delete_entity_state` | `pin_objects` | Fails (gracefully) if an `Entity` (for the given `nid`) does not already exist          |

Clients refer to `Entities` (and `SubEntities`) via `nid`s, which are shared, client-generated string IDs. 

Clients only care about the embedded `CreateMessage`s (of `Entities`) and `UpdateMessage`s (of `SubEntities)`, so `EntityView` returns this data exclusively. 

Client changes: https://github.com/mozilla/hubs/pull/5904


lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
+defmodule Ret.EntityState do
+  import Ecto.Query, warn: false
+  alias Ret.{Hub, Repo}
+
+  alias Ret.EntityState.{CreateMessage, UpdateMessage}
+
+  def list_entity_states(hub_id) do
```
bryanenders, line 7: Please write tests for this function

lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
...
+    Repo.all(
+      from create_message in CreateMessage,
+        where: create_message.hub_id == ^hub_id,
+        preload: [:entity_update_messages]
+    )
+  end
+
+  def create_entity_state!(
```
bryanenders, line 15: Please write tests for this function

lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
...
+      Repo.one!(
+        from create_message in CreateMessage,
+          where: create_message.hub_id == ^hub_id,
+          where: create_message.nid == ^nid,
+          preload: [:entity_update_messages]
+      )
+    end)
+  end
+
+  def update_entity_state!(
```
bryanenders, line 45: Please write tests for this function

lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
...
+      (entity_update_message || %UpdateMessage{})
+      |> UpdateMessage.changeset(hub, create_message, %{
+        nid: nid,
+        update_message: Poison.encode!(update_message)
+      })
+      |> Repo.insert_or_update!()
+    end)
+  end
+
+  def delete_entity_state!(hub_id, nid) do
```
bryanenders, line 74: Please write tests for this function

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
     {:noreply, socket}
   end
 
+  def handle_in("list_entity_states", _, socket) do
```
bryanenders, line 459: A test for this message would be nice

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
...
+    hub = socket |> hub_for_socket
+    entity_states = EntityState.list_entity_states(hub.hub_id)
+    {:reply, {:ok, EntityStateView.render("index.json", %{entity_states: entity_states})}, socket}
+  end
+
+  def handle_in(
+        "create_entity_state",
```
bryanenders, line 466: A test for this message would be nice

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
...
+        {:reply, {:error, %{reason: reason}}, socket}
+
+      {:error, other} ->
+        IO.inspect({:error, other})
+        {:reply, {:error, %{reason: "TODO"}}, socket}
+    end
+  end
+
+  def handle_in(
+        "update_entity_state",
```
bryanenders, line 495: A test for this message would be nice

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
...
+        {:reply, {:error, %{reason: reason}}, socket}
+
+      {:error, _other} ->
+        # TODO Why might this fail? What error messages would we receive?
+        {:reply, {:error, %{reason: "entity update failed."}}, socket}
+    end
+  end
+
+  def handle_in(
+        "delete_entity_state",
```
bryanenders, line 519: A test for this message would be nice

lib/ret/entity_state/create_message.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Ret.EntityState.CreateMessage do
```
bryanenders, line 1: This is a potential confusing struct name, being both a noun and a predicate. It’s conventional for Elixir functions that retrieve an indeterminate number of elements to be named with a collective noun (e.g., `potatoes/0`). If we were to do that here (i.e., `create_messages/0`), it wouldn’t be unreasonable for the reader to infer that the function is a command rather than a query.

lib/ret/entity_state/update_message.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Ret.EntityState.UpdateMessage do
```
bryanenders, line 1: Ditto on the ambiguous struct name.

lib/ret/entity_state/create_message.ex
bryanenders: I don’t understand the name, but maybe `MessageCreation` is better?

lib/ret/entity_state/update_message.ex
bryanenders: Maybe `MessageUpdate`? I really don’t know.

lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
...
+    Repo.transaction(fn ->
+      %CreateMessage{}
+      |> CreateMessage.changeset(hub, %{
+        nid: nid,
+        create_message: Poison.encode!(create_message)
+      })
+      |> Repo.insert!()
+
+      for update <- updates do
+        # TODO How / where are we supposed to handle string <-> atom conversion?
```
bryanenders, line 28: Let’s chat about this

lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
...
+  def delete_entity_state!(hub_id, nid) do
+    Repo.transaction(fn ->
+      create_message =
+        Repo.one!(
+          from create_message in CreateMessage,
+            where: create_message.hub_id == ^hub_id,
+            where: create_message.nid == ^nid
+        )
+
+      Repo.delete_all(
```
bryanenders, line 83: This would be better handled by the database with an ON DELETE action. That can be defined in the migration.

priv/repo/migrations/20230129061701_add_entity_create_messages_table.exs
```diff
@@ -0,0 +1,20 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:entity_create_messages, primary_key: false) do
+      add :entity_create_message_id, :bigint,
+        default: fragment("ret0.next_id()"),
+        primary_key: true
+
+      add :nid, :string, null: false
+      add :create_message, :binary
```
bryanenders, line 11: Under what circumstances would this be null?

priv/repo/migrations/20230129061719_add_entity_update_messages_table.exs
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    create unique_index(:entity_update_messages, [:nid, :hub_id])
+    create index(:entity_update_messages, [:hub_id])
+    create index(:entity_update_messages, [:nid])
+
+    # create index(:entity_update_messages_entity_create_message_id_hub_id_index, [
+    #          :entity_create_message_id,
+    #          :hub_id
+    #        ])
```
bryanenders, line 28: Commented code can be removed

bryanenders (commented):
I’m having trouble following this schema design; let’s chat. This could also use some tests.

lib/ret/entity_state/create_message.ex
```diff
@@ -0,0 +1,25 @@
...
+  import Ecto.Changeset
+
+  alias Ret.Hub
+  alias Ret.EntityState.{CreateMessage, UpdateMessage}
+  @schema_prefix "ret0"
+  @primary_key {:entity_create_message_id, :id, autogenerate: true}
+
+  schema "entity_create_messages" do
+    field :nid, :string
+    field :create_message, :binary
```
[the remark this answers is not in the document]
johnshaughnessy: Yes. The binary field (here and in `UpdateMessage`) is the main data we are interested in storing to the database. It is the (client-side) `CreateMessage` that was sent to initialize an entity. In API responses, we send only this binary data to clients.

Documentation: https://github.com/mozilla/hubs/blob/6bbf167011917e4fc7d5821ce4e493940463e386/doc/org/build/networking.md#createmessage
Typescript type: https://github.com/mozilla/hubs/blob/7df9cf2e66fb6e73748e5351de4be1640fe8a796/src/utils/networking-types.ts#L12-L17  

We _could_ in theory store nothing but the `binary` blobs and the associated `hub_id`, but this would make it exceptionally difficult to find what we are looking for in any given query, since we would need to load an parse all the `binary`s to find the `networkId`. Instead, we save the `nid` and `hub_id` _next to_ the `binary`, so that we can index / search on them. 

It may be worth mentioning that custom clients will store custom data in these messages. We will not know the shape of the `initialData` within a (client-side) `CreateMessage`, or of the `data` within a (client-side) `UpdateMessage`.

lib/ret/entity_state/create_message.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Ret.EntityState.CreateMessage do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Ret.Hub
+  alias Ret.EntityState.{CreateMessage, UpdateMessage}
+  @schema_prefix "ret0"
+  @primary_key {:entity_create_message_id, :id, autogenerate: true}
```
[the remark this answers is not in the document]
johnshaughnessy: The custom keys we use allow us to shard across multiple data stores, while keeping some desirable properties.

I believe [the code that generates the sharding ids](https://github.com/mozilla/reticulum/blob/6d2d9df5b85a750b974cecb6c2de7fa4ce301ecb/priv/repo/migrations/20170917234137_create_sharding_ids.exs#L9-L23) comes from [this blog post](https://instagram-engineering.com/sharding-ids-at-instagram-1cf5a71e5a5c).

From the blog post:
> The typical solution that works for a single database — just using a database’s natural auto-incrementing primary key feature — no longer works when data is being inserted into many databases at the same time.

...

> Before starting out, we listed out what features were essential in our system:
> 1. Generated IDs should be sortable by time (so a list of photo IDs, for example, could be sorted without fetching more information about the photos)
> 2. IDs should ideally be 64 bits (for smaller indexes, and better storage in systems like Redis)
> 3. The system should introduce as few new ‘moving parts’ as possible — a large part of how we’ve been able to scale Instagram with very few engineers is by choosing simple, easy-to-understand solutions that we trust.

lib/ret/entity_state.ex
johnshaughnessy: Yep, I will (re-)add tests after we get a chance to discuss whether the design is going in the right direction

lib/ret_web/views/entity_state_view.ex
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def render("entity_state.json", %{entity_state: %CreateMessage{} = message}) do
+    %{
+      create_message: Poison.Parser.parse!(message.create_message, %{}),
+      update_messages:
+        Enum.map(
+          message.entity_update_messages,
+          fn u ->
+            Poison.Parser.parse!(u.update_message, %{})
```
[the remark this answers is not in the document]
johnshaughnessy: Indeed it does!

lib/ret/entity_state.ex
johnshaughnessy: TIL

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
...
+      )
+
+      {:reply, :ok, socket}
+    else
+      {:error, %{reason: reason}} ->
+        {:reply, {:error, %{reason: reason}}, socket}
+
+      {:error, other} ->
+        IO.inspect({:error, other})
+        {:reply, {:error, %{reason: "TODO"}}, socket}
```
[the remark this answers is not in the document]
johnshaughnessy: I left this `TODO` because I knew I did not account for the case where `create_entity_state!` fails. I expect to happen whenever two clients attempt to pin an entity at the same time as each other. It is not a "serious" failure (I want the system to continue working as it was before), but I want to tell clients that their pin attempt failed, if the reason for the failure was that "the entity was already pinned".

I don't want to mix "`create_entity_state!` failed because of some unexpected issue" with "`create_entity_state!` failed because another client already pinned this entity", so I need to return better information from `create_entity_state!` to get rid of this TODO. In the "unexpected issue" case, I agree that we can let it fail.

lib/ret_web/channels/hub_channel.ex
```diff
@@ -871,6 +950,21 @@ defmodule RetWeb.HubChannel do
     end
   end
 
+  defp authorize(socket, :write_entity_state) do
+    with account <- Guardian.Phoenix.Socket.current_resource(socket),
+         :ok <- if(account === nil, do: :not_logged_in, else: :ok),
```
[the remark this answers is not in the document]
johnshaughnessy: I reached for the `with` statement here because I don't know how to write this function with nesting one `case` or `if` statement inside another, and another, and another. How would you write this function?

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
...
+             updates: updates
+           }) do
+      broadcast!(
+        socket,
+        "entity_state_created",
+        EntityStateView.render("show.json", %{entity_state: entity_state})
+      )
+
+      {:reply, :ok, socket}
+    else
```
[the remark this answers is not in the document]
johnshaughnessy: I would benefit from seeing a demonstration so I can understand how to avoid nested `case` or `if` statements. (`authorize/2` seems like a good candidate). I suspect I'm missing some basic elixir knowledge

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
...
+             update_message: update_message
+           }) do
+      broadcast!(socket, "entity_state_updated", update_message)
+      {:reply, :ok, socket}
+    else
+      {:error, %{reason: reason}} ->
+        {:reply, {:error, %{reason: reason}}, socket}
+
+      {:error, _other} ->
+        # TODO Why might this fail? What error messages would we receive?
```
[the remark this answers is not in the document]
johnshaughnessy: Ah, yes

lib/ret_web/channels/hub_channel.ex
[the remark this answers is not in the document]
johnshaughnessy: I tried to find some information online to understand why `with-else` is an anti-pattern. From my understanding, it's not `with-else` that is the anti-pattern, but having a complex `else` (e.g. by pattern matching on error tuples). The reason this is a code smell is because it's impossible to tell which clause of the `with` returned the error tuple.

José Valim gives an example and explanation here: https://github.com/lucasvegi/Elixir-Code-Smells#complex-else-clauses-in-with

The suggested remedy is "instead of concentrating all error handlings within a single complex else block, it is better to normalize the return types in specific private functions". I'll do that here.

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +456,84 @@ defmodule RetWeb.HubChannel do
     {:noreply, socket}
   end
 
+  def handle_in("list_entity_states", _, socket) do
+    hub = socket |> hub_for_socket
+    entity_states = EntityState.list_entity_states(hub.hub_id)
```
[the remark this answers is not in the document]
johnshaughnessy: Hmm... `Ret` seems like a noticeably less helpful name than `EntityState`, given that all of the functions in the module manipulate the entity state stored in the database (i.e. the `EntityCreateMessage` and `EntityUpdateMessage` tables). `EntityStateManagement` smells to me -- alarm bells ring for me when I see `FooManager` and `BarManager` and eventually `ManagersManager`.  

But I agree that with a line like `entity_states = EntityState.list_entity_states(hub.hub_id)`, I would expect the type of `entity_states` to be a collection of `EntityState`s. In reality, they are a collection of `EntityCreateMessage`s, each of which `has_many` `EntityUpdateMessage`s. 

I think the problem is that `EntityCreateMessage` is misnamed. Those are the things that are _actually_ supposed to be `EntityState`s. They should have a single `create_message` and many `update_messages`. I think that change might fix all of the other confusing names in this PR.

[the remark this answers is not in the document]
johnshaughnessy: OK, I've updated the nouns introduced by this PR to `Entity` and `SubEntity`. These match the nouns in the client. 

`SubEntity` here refers to "sub" entities that exist within the scene graph hierarchy under a common ancestor to which they belong (the `Entity`). Importantly, it captures the notion that `SubEntities` _belong to_ or _are part of_ a (root) `Entity`. This way, the semantics of the words also mirror the semantics of the client code, where removing a (root) entity will automatically remove all of its descendant (sub) entities, and removing a (sub) entity alone is not allowed.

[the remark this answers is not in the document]
johnshaughnessy: It seems that my bias against `FooManager` is misplaced. I see examples in ecto docs that have `FooManager` modules for managing `Foo`s.

lib/ret/entity_state.ex
```diff
@@ -0,0 +1,92 @@
...
+      |> Repo.insert!()
+
+      for update <- updates do
+        # TODO How / where are we supposed to handle string <-> atom conversion?
+        update_entity_state!(hub, %{
+          root_nid: nid,
+          nid: update["nid"],
+          update_message: update["update_message"]
+        })
+      end
```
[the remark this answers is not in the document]
johnshaughnessy: Done

lib/ret_web/channels/hub_channel.ex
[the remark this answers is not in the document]
johnshaughnessy: Done

lib/ret_web/channels/hub_channel.ex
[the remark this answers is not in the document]
johnshaughnessy: Ok, now each special function returns normalized return types (`{:error, reason}`). I still have an `else` because I want to send the error to the client in a reply, but there's no pattern matching in the `else`, so I think the code smell has been mitigated.

lib/ret_web/channels/hub_channel.ex
```diff
@@ -1357,4 +1432,20 @@ defmodule RetWeb.HubChannel do
...
+
+  defp parse(%{"nid" => nid, "root_nid" => root_nid, "update_message" => update_message}) do
+    %{nid: nid, root_nid: root_nid, update_message: Jason.encode!(update_message)}
+  end
+
+  defp parse(%{"nid" => nid, "create_message" => create_message, "updates" => updates}) do
+    %{
+      nid: nid,
+      create_message: Jason.encode!(create_message),
+      updates: Enum.map(updates, &parse(&1))
```
bryanenders, line 1444:
```suggestion
      updates: Enum.map(updates, &parse/1)
```

lib/ret/entity_manager.ex
```diff
@@ -0,0 +1,36 @@
...
+  def create_entity(hub, %{updates: updates} = entity_params) do
+    Multi.new()
+    |> Multi.insert(:entity, Entity.changeset(%Entity{}, hub, entity_params))
+    |> Multi.insert_all(
+      :sub_entities,
+      SubEntity,
+      fn %{entity: entity} ->
+        Enum.map(updates, &SubEntity.for_bulk_insert(hub, entity, &1))
+      end
+    )
```
bryanenders, line 15: I’m fairly certain this whole `create_entity/2` function is achievable with a single `Repo.insert/1` sans the `Multi`. `Ecto.Changeset.cast_assoc/3` is the key. We can chat.

lib/ret_web/channels/hub_channel.ex
```diff
@@ -455,6 +455,54 @@ defmodule RetWeb.HubChannel do
...
+      {:error, reason} ->
+        reply_error(socket, reason)
+    end
+  end
+
+  def handle_in("update_entity_state", %{"update_message" => update_message} = params, socket) do
+    params = parse(params)
+
+    with {:ok, hub} <- authorize(socket, :write_entity_state),
+         {:ok, _} <- Ret.insert_or_update_sub_entity(hub, params) do
```
bryanenders, line 482: Confirming. Are sub-entity insertions and update both possible for `"update_entity_state"`?

lib/ret/entity_state/create_message.ex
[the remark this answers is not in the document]
bryanenders: I see two counterpoints:
1. We aren’t sharding, so it’s [YAGNI](https://martinfowler.com/bliki/Yagni.html).
2. We are autogenerating the IDs the same as if we used the out-of-the-box behavior.  All this custom configuration (which spans numerous files) is doing is changing the _name_ of the primary ID.

bryanenders:
This is looking really good! It’s almost there!

lib/ret/entity_state/create_message.ex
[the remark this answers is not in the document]
johnshaughnessy: Ah right, we are also giving custom _names_ to the primary keys. 

I don't fully understand the reason why this was important. Perhaps something to do with wanting to avoid passing anonymous `id`s around, and then having to track back to figure out why the thing you thought was a `foo.id` was actually a `bar.id`. 

I'm not repeating the pattern because I endorse it. I'm repeating it because it's used for every other table. If we feel strongly that it's bad then we should change it. (But I don't think this PR is the appropriate place to break that pattern.)

lib/ret_web/channels/hub_channel.ex
johnshaughnessy: Yep! If one exists, it should be updated. If it's a new one, it should be inserted.

test/ret_web/channels/entity_test.exs
```diff
@@ -0,0 +1,143 @@
...
+  defp local_file(filename) do
+    Path.join(Path.dirname(__ENV__.file), filename)
+  end
+
+  defp read_json(filename) do
+    Jason.decode!(File.read!(local_file(filename)))
+  end
+
+  defp payload(:save_entity_state) do
+    read_json("save_entity_state_payload.json")
```
bryanenders, line 120: **nit:** This can be done at compile time, so we don’t have to perform a costly IO call every time this is invoked

test/ret_web/channels/entity_test.exs
```diff
@@ -0,0 +1,143 @@
...
+    assert entity.create_message["networkId"] === payload(:save_entity_state_2)["nid"]
+    assert entity.create_message === payload(:save_entity_state_2)["create_message"]
+  end
+
+  defp local_file(filename) do
+    Path.join(Path.dirname(__ENV__.file), filename)
+  end
+
+  defp read_json(filename) do
+    Jason.decode!(File.read!(local_file(filename)))
```
bryanenders, line 116: **nit:** A pipeline would be idiomatic

bryanenders (approved):

test/ret_web/channels/entity_test.exs
johnshaughnessy: fixed

PR #664, opened 2023-01-27
Upgrade to Elixir v1.14 and native releases

bryanenders:
This PR upgrades to the latest Elixir version. It will be followed by later PRs to upgrade to the latest Erlang version and upgrade all the dependencies.


mix.lock
```diff
@@ -1,92 +1,89 @@
...
-  "plug": {:hex, :plug, "1.10.4", "41eba7d1a2d671faaf531fa867645bd5a3dce0957d8e2a3f398ccff7d2ef017f", [:mix], [{:mime, "~> 1.0", [hex: :mime, repo: "hexpm", optional: false]}, {:plug_crypto, "~> 1.1.1 or ~> 1.2", [hex: :plug_crypto, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "ad1e233fe73d2eec56616568d260777b67f53148a999dc2d048f4eb9778fe4a0"},
+  "phoenix_pubsub": {:hex, :phoenix_pubsub, "2.1.1", "ba04e489ef03763bf28a17eb2eaddc2c20c6d217e2150a61e3298b0f4c2012b5", [:mix], [], "hexpm", "81367c6d1eea5878ad726be80808eb5a787a23dee699f96e72b1109c57cdd8d9"},
+  "plug": {:hex, :plug, "1.14.0", "ba4f558468f69cbd9f6b356d25443d0b796fbdc887e03fa89001384a9cac638f", [:mix], [{:mime, "~> 1.0 or ~> 2.0", [hex: :mime, repo: "hexpm", optional: false]}, {:plug_crypto, "~> 1.1.1 or ~> 1.2", [hex: :plug_crypto, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4.3 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "bf020432c7d4feb7b3af16a0c2701455cbbbb95e5b6866132cb09eb0c29adc14"},
```
johnshaughnessy, line 64: It seems that as of plug `1.14.0`, we cannot assign a `"host"` header with
```elixir
Conn.put_req_header(conn, "host", authority)
```
which we do in `PageController`: https://github.com/mozilla/reticulum/blob/dd14e634699d72c2640b459b31778551feed6e75/lib/ret_web/controllers/page_controller.ex#L720-L722

https://github.com/elixir-plug/plug/pull/1104

> The "host" header will be overridden by conn.host and should not be set with this method. Instead, do `%Plug.Conn{conn | host: value}`

https://hexdocs.pm/plug/Plug.Conn.html#put_req_header/3

I tried following the instructions above but have not been able to get `PageController.cors_proxy/2` working properly. I might not have my local environment configured correctly.

johnshaughnessy: It seems as though requesting a proxied resource results in rapid redirects that fail to resolve. (?)
snip

johnshaughnessy (commented):
Excellent! :tada: We've neglected these upgrades for so long. I'm very glad to see this.

I was not able to get most things running smoothly in my local environment, but not `PageController.cors_proxy` (see the comments below).

We should make sure to build a habitat package and run it on a hubs cloud instance before the next hubs cloud release. 

I didn't understand the changes in the crypto module, but I assume these were necessary. I'd like to learn where to put different configs (`prod.exs` / `dev.exs` vs the new `runtime.exs`). I thought the hubs cloud instances did _some_ of their configuration by manipulating one of these files, but I can't remember which (if any). We'll find this out by testing.

mix.lock
johnshaughnessy: OK. Despite the warnings in `Plug`, it seems like in this case we actually _want_ `conn.host` to be different from the `{"host", value}` in `conn.req_headers`:
```elixir
%Conn{ host: "hubs-proxy.local", req_headers: [{ "host", "uploads-prod.reticulum.io" }] }
```

Here is a fix:

```diff
modified   lib/ret_web/controllers/page_controller.ex
@@ -719,7 +719,7 @@ defmodule RetWeb.PageController do
         |> Map.put(:path_info, [])
         # Since we replaced the host with the IP address in ip_url above, we need to force the host
         # header back to the original authority so that the proxy destination does not reject our request
-        |> Conn.put_req_header("host", authority)
+        |> unsafe_put_req_header("host", authority)
         # Some domains disallow access from improper Origins
         |> Conn.delete_req_header("origin")
         |> ReverseProxyPlug.request(body, opts)
@@ -730,6 +730,14 @@ defmodule RetWeb.PageController do
     end
   end
 
+  # Bypass Plug.Conn.validate_req_header!/2 in order to set the "host" header
+  # https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn.ex#L1887-L1892
+  # https://github.com/elixir-plug/plug/blob/d7940552e365e0d3245bb78ca99429ea5c324f7d/lib/plug/conn.ex#L785-L789
+  defp unsafe_put_req_header(%Conn{req_headers: headers} = conn, "host" = key, value)
+       when is_binary(value) do
+    %{conn | req_headers: headers |> List.delete(key) |> List.keystore(key, 0, {key, value})}
+  end
```

bryanenders: It’s unfortunate that this deprecated behavior didn’t show up in the [Plug changelog](https://github.com/elixir-plug/plug/blob/main/CHANGELOG.md). I reviewed all the changelogs while upgrading.

habitat/plan.sh
```diff
@@ -8,13 +8,13 @@ pkg_deps=(
     core/coreutils/8.30/20190115012313
     core/bash/4.4.19/20190115012619
     core/which/2.21/20190430084037
-    mozillareality/erlang/22.0
+    mozillareality/erlang/23.3.4.18
```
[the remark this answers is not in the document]
bryanenders: Absolutely. Thanks for highlighting this!

bryanenders:
@johnshaughnessy In a release the application configuration in `config.exs` is baked at compile time. Our application happens to be set up so that `config.exs` imports the environment-specific config in `$MIX_ENV.exs`, which is pretty conventional. Since some application configuration isn’t knowable until runtime, `runtime.exs` was added.

If you are adding configuration that isn’t effectful, put it in the `config.exs` files. If the config is dependent upon an outside factor such as the host (e.g., OS variables) or time, put it in `runtime.exs`.

You will notice some configuration values in `runtime.exs` that are not _themselves_ effectful, such as `"ret_dev"` and `"postgres"` in this example:
```elixir
    db_hostname = System.get_env("DB_HOST", "localhost")

    config :ret, Ret.Locking,
      session_lock_db: [
        database: "ret_dev",
        hostname: db_hostname,
        password: "postgres",
        username: "postgres"
      ]
```
Note that `Config.config` comes in only two flavors: a 2-arity and a 3-arity. As such the example config overrides the entirety of the `:session_lock_db` key in the application configuratione but leaves the rest of `Ret.Locking` and `:ret` alone. The effect-free values are included because they are part of a data structure that _is_ constructed based on side effects owing to the use of `$DB_HOST`.

scripts/hab-wrap-and-push.sh
```diff
@@ -76,7 +76,7 @@ do_build() {
 }
 do_install() {
     rm -rf _build/prod/rel/ret/releases
-    MIX_ENV=prod mix distillery.release
+    MIX_ENV=prod mix release
```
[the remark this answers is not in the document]
bryanenders: Those came from https://github.com/mozilla/reticulum/commit/47b65d26e4a4237fd8f247ab8cad8e90dabc15f4. There are some similar comments in https://github.com/mozilla/reticulum/commit/22991f7f335fb9e7ccc8cf359e3af128cb605818. @tanfarming Could you lend some context?

johnshaughnessy (approved):
Looks good to ship to Turkey. I think building the habitat packages will fail until elixir/erlang is updated on the builder instance. (I'm not sure how to update them. I think we have to build a package and upload it.) But that doesn't need to stop us from shipping to the managed infra.

PR #668, opened 2023-02-01
Upgrade to Elixir v1.14 and native releases, Take 2

bryanenders:
This PR upgrades to the latest Elixir version. It will be followed by later PRs to upgrade to the latest Erlang version and upgrade all the dependencies.

# Why Take 2?
I assumed, incorrectly, that a flag in `prod.exs` was equivalent to `Mix.env() == :prod`.  What I failed to account for was the fact that there is an extra production environment `:turkey` that reads from `prod.exs`.  So anything that should only apply to `:prod` should be overridden in `turkey.exs`.

[This](https://github.com/mozilla/reticulum/compare/3799201e79784ca99f92c7bc206b184bf160be85..upgrade-elixir) is the change from Take 1.


bryanenders:
Rebased onto `master` so the build only compiles once

johnshaughnessy (approved):

PR #676, opened 2023-02-16
Update API.md with correction and new info

mikemorran:
Corrections: 
1. Note about alpha testing removing note that users cannot create their own tokens.

New Info:
1. Line about where to generate tokens.
2. Write rooms with info about customizable room features taken from graphQL schema in room_types.ex

### DIFF
--- a/guides/api.md
+++ b/guides/api.md
@@ -1,7 +1,7 @@
 # Hubs Server API
 Reticulum includes a [GraphQL](https://graphql.org/) API that grants programmatic access to server resources. 
 
-Note: This API is currently in alpha testing and will likely experience frequent updates.
+Note: This API is not yet stable, and may undergo breaking changes in future updates.
 
 ## Accessing the API
 Hubs Cloud administrators can enable or disable the API by toggling `App Settings > Features > Public API Access` in the admin panel. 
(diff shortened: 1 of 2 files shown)


mikemorran:
I see that I have a few markdown errors...

guides/api.md
```diff
@@ -1,7 +1,7 @@
 # Hubs Server API
 Reticulum includes a [GraphQL](https://graphql.org/) API that grants programmatic access to server resources. 
 
-Note: This API is currently in alpha testing and is not yet available for use. (Users cannot generate API Access Tokens.)
+Note: This API is currently in alpha testing and will likely experience frequent updates.
```
johnshaughnessy, line 4: I might just say, "This API is not yet stable, and may undergo breaking changes in future updates."

guides/api.md
```diff
@@ -28,6 +30,26 @@ Each API Access Token specifies its `scopes`. Scopes allow a token to be used to
...
+  - `spawnDrawing` _**bool**_ \| Allow non-admin participants to draw with a pen.
+  - `pinObjects` _**bool**_ \| Allow non-admin participants to pin media to the room.
+  - `spawnEmoji` _**bool**_ \| Allow non-admin participants to spawn emoji.
+  - `fly` _**bool**_ \| Allow non-admin participants to toggle fly mode.
+  - `voiceChat` _**bool**_ \| Allow non-admin participants to use the voice chat.
+  - `textChat` _**bool**_ \| Allow non-admin participants to use the text chat.
+- `user_data` _**map**_ \| Arbitrary json data associated with this room.
+
+See [room_types.ex](../lib/ret_web/schema/room_types.ex) for full GraphQL Schema.
+
```
johnshaughnessy, line 52: Thank you for writing this. 

I am torn on whether to include it, because GraphQL has built-in `__schema` field that allows users to learn about a query: https://graphql.org/learn/introspection/

The built-in method has no risk of being outdated or wrong as we change things, whereas having this data duplicated here means that we risk forgetting to update it, or making a mistake.

I would not want to recreate the documentation that the graphql team already worked hard to write -- Perhaps we should link the graphql documentation and include some introspection queries in our sample workspace file.

johnshaughnessy (approved):
Leaving an "approve" review despite my comments because I do not want you to be blocked from proceeding if this is what you want to deploy. 

The "reviews require explicit approval" rule is only in place to protect ourselves from running database migrations that we would regret. This PR does not have that risk whatsoever.

guides/api.md
mikemorran: I did not know about this and agree we should utilize this tool which will prevent us from having to constantly update. I will review the docs tomorrow and put together a bit more of a step-by-step of how to quickly see what is available.

mikemorran:
@johnshaughnessy @bryanenders I have updated the doc and added a new sample workspace based on John's recommendation to share the introspection tools. 

The sample workspace I added also removes the duplicate tabs for [app] or [user] scopes. I did this because it seemed confusing and maybe a bit redundant, but I feel like I might be missing something important about distinguishing the two.

bryanenders:
IIRC there is a vulnerability in GraphiQL, so it may be going away. I’ll have to dig to confirm. The good news is that the `queryType` and `mutationType` should be queryable. Those are types that make up the schema.

PR #681, opened 2023-02-22
Prevent version mismatch with Hubs Compose

bryanenders:
Why
---
The Elixir and Erlang versions are specified in both repositories, this
one and Hubs Compose.  If a developer pulls a commit that contains a
tool version change in one of the repos, a version mismatch occurs.

What
----
* Move dev container build stage from Hubs Compose into Reticulum
* Move dev container files into this repo
* Gather container files in a single directory

Side effects
------------
Hubs Compose will have to be updated to take advantage of these changes


johnshaughnessy (approved):
Makes sense to me to consolidate the dockerfiles so that the elixir/erlang versions always match.

PR #683, opened 2023-02-28
update is_cors_proxy_url check to enable cors_proxy when hosted behind proxy

tanfarming:
problem:
in turkey arch, reticulum is hosted behind ingress controller.
the ingress controller terminates TLS connection for reticulum, so the connection observed by reticulum is http instead of https, this will fail the is_cors_proxy_url check

solution:
- TURKEY_MODE switch
- uses the "de-facto standard" x-forwarded-proto header set at ingress to determine
- no change on host name check
- skip port# check

### DIFF
--- a/lib/ret_web/controllers/api/v1/media_controller.ex
+++ b/lib/ret_web/controllers/api/v1/media_controller.ex
@@ -2,6 +2,7 @@ defmodule RetWeb.Api.V1.MediaController do
   use RetWeb, :controller
   use Retry
   alias Ret.Statix
+  require Logger
 
   def create(conn, %{"media" => %{"url" => url, "quality" => quality}, "version" => version}),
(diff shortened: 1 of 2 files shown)


bryanenders (changes requested):
Could you please
1. Run `mix format`
2. Squash the 15 commits
3. Give the commit a useful description

tanfarming:
> Could you please
> 
>     1. Run `mix format`
my local elixir envs are a mess it may be a long while before i have the time to sort this out
lets merge this real quick so we can start testing and the feature's expected to be released to users by end of this week, linter's nice but shouldn't block releases imo and it will be fixed on your next merge anyway

>     2. Squash the 15 commits
> 
>     3. Give the commit a useful description
I'll squash and merge that should sort out #2 and #3

PR #685, opened 2023-03-01
Reduce number of ytdl retries and expire faster

johnshaughnessy:
Reduce the number of attempts to fetch a URL from ytdl, and expire the retry loop faster.

If ytdl is returning some unexpected error that causes us to retry, then we might be making more requests than is necessary. This protects against that case.

**edit**
I'm not sure that this PR meaningfully improves the situation. I need to measure whether resolving media causes many retries or not. If what's _actually_ happening is that media resolution hangs until the 7 or 15 second timeout, then this change is not helpful or necessary.

### DIFF
--- a/lib/ret/media_resolver.ex
+++ b/lib/ret/media_resolver.ex
@@ -232,6 +232,9 @@ defmodule Ret.MediaResolver do
       %HTTPoison.Response{body: body} ->
         {:error, body}
 
+      :timeout ->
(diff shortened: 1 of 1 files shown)


tanfarming (approved):

johnshaughnessy:
Closing this PR because I don't think it's solving the correct problem.

johnshaughnessy:
Reopened because I realized in prod, we're still retrying past the 7 second timeout. This change is at least an improvement in that case.

PR #688, opened 2023-03-02
Broadcast host_changed events when hub.host changes

johnshaughnessy:
If a client causes a room's `host` to change, broadcast the change so that other clients in the room are notified of this change and can migrate their connections to the new dialog host.

https://github.com/mozilla/hubs/pull/5975

### DIFF
--- a/lib/ret/hub.ex
+++ b/lib/ret/hub.ex
@@ -668,6 +668,18 @@ defmodule Ret.Hub do
     end
   end
 
+  def test_change_host(hub) do
+    host = RoomAssigner.get_available_host(hub.host)
+
+    hub |> changeset_for_new_host(host) |> Repo.update!()
+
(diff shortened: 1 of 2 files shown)


lib/ret/hub.ex
```diff
@@ -656,6 +656,12 @@ defmodule Ret.Hub do
 
       if host && host != hub.host do
         hub |> changeset_for_new_host(host) |> Repo.update!()
+
+        RetWeb.Endpoint.broadcast("hub:" <> hub.hub_sid, "host_changed", %{
+          host: host,
+          port: Hub.janus_port(),
+          turn: Hub.generate_turn_info()
+        })
```
johnshaughnessy, line 664: I accidentally changed the return type of this function. Fixed in #689.

PR #690, opened 2023-03-07
Fix habitat package resolution

johnshaughnessy:
Attempting to fix a build time dependency resolution issue:
image

I don't know _why_ it can't find `mozillareality/erlang/23.3.4.18` (after all, it's here and `stable`). But Bryan mentioned something in chat that indicated she seemed to be able to fix this issue with `core/erlang/23.3.4.9`, so it seems worth a try.


johnshaughnessy:
This was not a problem with any code in reticulum. We resolved it in bldr.

PR #693, opened 2023-03-08
delete string session_id key before all Map.put(:session_id instances

tanfarming:


johnshaughnessy (approved):

johnshaughnessy (commented):
We may need to remove `"from_session_id"` for the same reason.

If there are some invalid keys that we don't allow in the payload, it might be good to find what all of them are.

PR #702, opened 2023-06-23
Permission updates authorization

keianhzo:
Fixes https://mozilla-hub.atlassian.net/browse/HUBSTRKY-773

### DIFF
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -297,22 +297,24 @@ defmodule RetWeb.HubChannel do
     account = Guardian.Phoenix.Socket.current_resource(socket)
     hub = socket |> hub_for_socket
 
-    if (type == "photo" and type == "video" and
-          account
-          |> can?(spawn_camera(hub))
-          |> Kernel.not()) or
-         (type == "permission" and hub |> Ret.Hub.is_owner?(account.account_id) |> Kernel.not()) do
-      {:noreply, socket}
-    end
+    authorized =
+      cond do
+        type in ["photo", "video"] -> account |> can?(spawn_camera(hub))
+        type === "chat" -> account |> can?(text_chat(hub))
+        type === "permission" -> account |> can?(update_hub(hub))
+        true -> true
+      end
 
-    broadcast!(
-      socket,
-      event,
-      payload
-      |> Map.delete("session_id")
-      |> Map.put(:session_id, socket.assigns.session_id)
-      |> payload_with_from(socket)
-    )
+    if authorized do
+      broadcast!(
+        socket,
+        event,
+        payload
+        |> Map.delete("session_id")
+        |> Map.put(:session_id, socket.assigns.session_id)
+        |> payload_with_from(socket)
+      )
+    end
 
     {:noreply, socket}
   end
--- a/lib/ret_web/channels/hub_channel.ex
+++ b/lib/ret_web/channels/hub_channel.ex
@@ -297,16 +297,22 @@ defmodule RetWeb.HubChannel do
     account = Guardian.Phoenix.Socket.current_resource(socket)
     hub = socket |> hub_for_socket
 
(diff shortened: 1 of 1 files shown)


lib/ret_web/channels/hub_channel.ex
```diff
@@ -297,17 +297,23 @@ defmodule RetWeb.HubChannel do
...
-        |> Map.delete("session_id")
-        |> Map.put(:session_id, socket.assigns.session_id)
-        |> payload_with_from(socket)
-      )
+    if (type == "photo" and type == "video" and
+          account
+          |> can?(spawn_camera(hub))
+          |> Kernel.not()) or
+         (type == "permission" and hub |> Ret.Hub.is_owner?(account.account_id) |> Kernel.not()) do
+      {:noreply, socket}
```
johnshaughnessy, line 305: This tuple does not appear to be returned from the function, which means the checks in this `if` have effectively been removed, since the code execution will always fall through to the `broadcast!` call below.

I'm also surprised there is not a specific permission check here, given the linked report that this is responding to. I expected to see something like  `can?(account, text_chat(hub))` somewhere in the `"message"` handler, since `"text_chat"` is the relevant permission.

johnshaughnessy (changes requested):

lib/ret_web/channels/hub_channel.ex
johnshaughnessy: I misunderstood the intended change by this PR. 

I see now that the desired effect is restricting the ability for an unauthorized user to change permissions. I'll need to familiarize myself with how permissions are handled in general. I didn't realize that we sent updates about the state of member permissions across the same way as a chat message.

johnshaughnessy: This check should not be for whether the account `is_owner`, but whether the account `can?(update_hub(hub))`. 

Ideally, though, a client would not need to broadcast this change to its peers in a separate `"message"` -- reticulum would broadcast any changes after handling them in the `handle_in("update_hub"` function (which also means that the authorization is moved to only one spot.

takahirox: If I'm right `type == "photo" and type == "video"` will be always false

keianhzo: Thanks for the feedback. Actually I realized that I was looking at this the wrong way. 

The actual permission updates are handled by `handle_in("update_hub")` as @johnshaughnessy mentioned and that is working fine as we authorize there. What this handler wasn't doing is authorizing the permission and chat messages so a malicious party could: 
- Send a fake permission message making everyone in the room believe that the voice/chat has been disabled (independently of the actual permission value)
- Send a chat message as we are not authorizing those here at all.

I think my latest commit correctly addressed these issues.

takahirox: Thinking where to check the permissions....

Ideally in the clients? With the future Add-on APIs addon creators may add their own permission type and most of them may not want to edit server side code.

I don't mean to block this PR but want to clarify the future direction.

keianhzo: My understanding is that if an add-on needs to add a new permissions, that needs to happen on reticulum first. Checking permissions can happen in the client's both ends as we are currently doing.

This PR addresses the chat messages authorization not the permission updates. Are you suggesting that we should authorize the chat messages also in the client ends instead of here? That can also be an option, not sure what would be the pros/cons but as we are already doing that for some other permissions might make sense. What do you think @johnshaughnessy?

takahirox: Yes, different pros/cons depending on where to check the permissions. I think what we need to do is considernig balance and deciding a rule or policy. If the places where the permissions are checked are scattered, it causes harder maintainability.

We may merge this PR and discuss it in another issue separately.

Random thoughts before digging into the current implementation and performance profile: Perhaps it may be good to basically check the permissions in the clients end to avoid the performance bottleneck in Reticulum? Checking some permissions in the Reticulum may not be avoidable if it has any side effect in Reticulum database tho.

PR #703, opened 2023-07-14
Detects non-referred files

takahirox:
**Problem**

Due to a bug in Reticulum and/or Hubs Client, `OwnedFiles` that are not referred from anywhere can be remained as active in the database then the associated files are not removed. This problem causes a high disk volume usage pressure.

**Solution**

This commit addresses the problem by adding a function that finds and deletes such non-referred files. The function might be removed when the root issue will be resolved.

Non-referred OwnedFile detection algorithm

1. Mark all `OwnedFiles` as inactive
2. Mark the `OwnedFiles` as active that are referred by `RoomObjects`
3. Mark the `OwnedFiles` as active that are referred by `AppConfig`, `Asset`, `Avatar`, `AvatarListing`, `Project`, `Scene`, or `SceneListing`
4. Delete the inactive `OwnedFiles`

Note that in this commit we don't delete the non-referred `OwnedFiles` and associated files. Instead until we are confident for this operation we move the non-referred `OwnedFile` rows to another new table and keep the associated files. Even if this operation has a bug and wrongly detects non-referred owned files we can restore. And users will notice the wrong detection problem by their contents that won't appear because they are moved from `OwnedFile` table.

After we test this operation much enough we will update the code to actually delete the non-referred files in another commit.

**Changes**

- Add a new table that holds non-referred `OwnedFile` rows
- Add a function that detects non-referred `OwnedFiles` and moves them to the new table
- Add unit tests for them

**TODO in the future**

- Test this operation much enough
- Update the function to actually delete non-referred files

**Tasks**
- [x] Implementation
- [ ] Optimization and error handling if needed
- [x] Unit Test
- [ ] Decide how the function should be invoked (eg: Periodically with crontab or manually)
- [ ] Test on instance

### DIFF
--- a/lib/ret/storage.ex
+++ b/lib/ret/storage.ex
@@ -337,6 +337,214 @@ defmodule Ret.Storage do
   defp check_promotion_token(actual_token, token) when actual_token != token,
     do: {:error, :invalid_key}
 
+  # Due to a bug in Reticulum and/or Hubs Client, OwnedFiles that are not referred
+  # from anywhere can be remained as active in the database then the associated
+  # files are not removed. This problem causes a high disk volume usage pressure.
+  # This function finds and deletes such files. The function might be removed
+  # when the root issue will be resolved.
+  # Question: Where (In what file) should this function be defined?
+  # Question: How should this function be invoked? Cron? Manually?
+  def cleanup_nonreferred_files do
+    Logger.info("Stored Files: Attempting to clean up non-referred files.")
+
+    # Lock for the operation because the operation may take relatively longer time
+    # due to many tables access and has a side effect. If other operations
+    # access the database while in this operation this operation or the other
+    # operations may not work correctly. Locking relatively longer time would be
+    # acceptable because this function should not be invoked so often.
+    # Question: Are both Atomic transaction (Ecto.Multi) and Lock needed?
+    #           Or is only Atomic transaction good enough?
+    Ret.Locking.exec_if_lockable(:storage_cleanup_nonreferred_files, fn ->
+      import Ecto.Query, only: [from: 2]
+
+      # TODO: Optimize, if possible
+      # TODO: Error handling if needed
+
+      multi = Ecto.Multi.new()
+
+      # 1. Mark all OwnedFiles as inactive
+
+      multi = multi
+        |> Ecto.Multi.update_all(:inactivate_all, OwnedFile, set: [state: :inactive])
+
+      # 2. Mark the OwnedFiles as active that are referred by RoomObjects
+
+      # file_uuid is encoded in RoomObject.gltf_node.extensions.HUBS_components.media.src as like
+      # https://hubs.local:4000/files/71301c23-8fb3-4388-8eec-83b3627f9ad0.glb?token=xxx
+      multi = multi
+        |> Ecto.Multi.run(:activate_if_referred_by_room_object, fn repo, _changes -> 
+             file_uuids = from(o in Ret.RoomObject, select: o.gltf_node)
+               |> repo.all()
+               |> Enum.map(&Jason.decode!/1)
+               |> Enum.flat_map(fn (gltf_node) ->
+                    # Question: Is it a safe assumption that this property exists?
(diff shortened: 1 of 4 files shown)


lib/ret/storage.ex
```diff
@@ -337,6 +337,111 @@ defmodule Ret.Storage do
...
+    # 2. Mark the OwnedFile as active that referred by RoomObjects
+
+    # file_id is encoded in RoomObject.gltf_node.extensions.HUBS_components.media.src as like
+    # https://hubs.local:4000/files/71301c23-8fb3-4388-8eec-83b3627f9ad0.glb?token=xxx
+    from(object in Ret.RoomObject, select: object.gltf_node)
+      |> Repo.all()
+      |> Enum.map(&Jason.decode!/1)
+      |> Enum.each(fn (gltf_node) ->
+           # Question: Is it guaranteed that this property exists?
+           src = gltf_node["extensions"]["HUBS_components"]["media"]["src"]
```
johnshaughnessy, line 370: As far as I know, they do. But `pinned-entity-to-gltf.js` seems to account for entities that do not have a `media-loader` component, so maybe not. We should log / inspect any nodes that don't have a `media.src`.

lib/ret/storage.ex
```diff
@@ -337,6 +337,111 @@ defmodule Ret.Storage do
...
+           # Question: Is it possible that src doesn't follow the format written above?
+           file_id = Regex.run(~r/files\/([^.]*)\./, src)
```
johnshaughnessy, line 372: If the `RoomObject` does _not_ point to an uploaded file, then the src will be a url to a public resource (`https://www.example.com/some_image.png`).

lib/ret/storage.ex
```diff
@@ -337,6 +337,111 @@ defmodule Ret.Storage do
...
+             |> Enum.at(1)
+           # Question: Use OwnedFile.set_state() instead?
+           # Question: Is it possible that an OwnedFile is referred by multiple RoomObjects?
```
johnshaughnessy, line 375: I think so. I suspect this will be the case if an uploaded file is "cloned" in the room.

lib/ret/storage.ex
```diff
@@ -337,6 +337,111 @@ defmodule Ret.Storage do
...
+
+    from(s in Ret.SceneListing, select: [s.model_owned_file_id, s.screenshot_owned_file_id, s.scene_owned_file_id])
+      |> Repo.all()
+      |> Enum.each(fn (file_ids) ->
+           from(f in OwnedFile, where: f.owned_file_uuid in ^file_ids)
+             |> Repo.update_all(set: [state: :active])
+         end)
+
+    # 4. Delete the inactive OwnedFiles
+    #    (Another option: Temporarily move them to another table and delete later.)
```
johnshaughnessy, line 433: Until we're confident that this is correct, I am very hesitant to actually delete the files and records.

takahirox: Good point. We may move them to another table in this PR, and when we will feel confident about this operation we may make another PR for actually deleting them.

takahirox: Done

lib/ret/storage.ex
takahirox: Thanks

lib/ret/storage.ex
takahirox: Thanks, updated the code to take into such urls account

lib/ret/storage.ex
takahirox: Thanks, updated the code to take into no media.src property account for now

takahirox:
Ready for review

sea922:
@takahirox I have question, why file upload by reticulum after a few day is not display. I want to upload image by using asset_hosts (https://hubs-01-assets.my-domain-url/files/8147c202-03f6-41a0-bd5f-930eb74.jpg?token=c62ccb39d31d7082280a3fe7e81d7)  Is there any way to prevent the file from being deleted? I use my website to manage content. In content have some image, I using api v1/media (reticulum) for upload. Can you give me solution?

PR #708, opened 2023-09-05
fix: embed typo

0o001:


DougReeder:
This is a flag in the Discord API which our code does not use.  So, when someone is doing other work with the Discord bot, it would be nice to first apply this and verify that nothing breaks.   However, until someone has some other reason to work on the Discord bot, there's no point spending time on this.

Exairnous (approved):
I reviewed this as part of updating the Discord bot to the latest API version and it looks good to me.  Thanks.  Merging.

PR #712, opened 2023-10-27
Remove token promotion from entity state

keianhzo:
This PR changes the pinning design for the new entity state to remove the concept of uploaded file promotion. 

To see the design changes check this related PR comment:
https://github.com/mozilla/hubs/pull/6287


johnshaughnessy (changes requested):
In `save_entity_state`, if the entity state has a `file_id`, the file needs to be promoted to permanent storage and the resulting `OwnedFile` record needs to be active.
In `delete_entity_state`, if the entity state has a file id, the associated `OwnedFile` needs to be marked inactive (for delayed deletion).

As written, uploaded files never get promoted, which means `Ret.Storage.vacuum` will delete them after their TTL expires.

johnshaughnessy (approved):

PR #717, opened 2024-04-07
Fix version of reverse_proxy_plug to 2.3.0

taku1201:
# Description

In `Import Content` section on hubs admin console, we input valid source url and click `PREVIEW IMPORT` button, then we met an issue that reticulum returns HTTP 500 error.
We investigated this issue and we found that version `2.3.0` of `reverse_proxy_plug` would not cause this problem.
As the current `mix.exs` configuration seems to allow installation for any minor versions from `2.1.x` onwards, so we propose to pin it to 2.3.0.

### reverse_proxy_plug == 2.4.0

```bash
$ docker compose exec reticulum cat /ret/releases/1.0.0/ret.rel | grep reverse_proxy_plug
          {reverse_proxy_plug,"2.4.0",permanent},
```

reticulum returns 500 error.

image

### reverse_proxy_plug == 2.3.1

```bash
$ docker compose exec reticulum cat /ret/releases/1.0.0/ret.rel | grep reverse_proxy_plug
          {reverse_proxy_plug,"2.3.1",permanent},
```

reticulum returns 500 error too.

image

Sentry Event: https://taku-ito.sentry.io/share/issue/25e12e1405ac415d9c81ab38fb34d17e/

## reverse_proxy_plug <= 2.3.0

```bash
$ docker compose exec reticulum cat /ret/releases/1.0.0/ret.rel | grep reverse_proxy_plug
          {reverse_proxy_plug,"2.3.0",permanent},
```

reticulum returns proper result and the import was successful.

image


mix.lock
```diff
@@ -71,7 +71,7 @@
   "ranch": {:hex, :ranch, "1.8.0", "8c7a100a139fd57f17327b6413e4167ac559fbc04ca7448e9be9057311597a1d", [:make, :rebar3], [], "hexpm", "49fbcfd3682fab1f5d109351b61257676da1a2fdbe295904176d5e521a2ddfe5"},
   "recon": {:hex, :recon, "2.5.0", "2f7fcbec2c35034bade2f9717f77059dc54eb4e929a3049ca7ba6775c0bd66cd", [:mix, :rebar3], [], "hexpm", "72f3840fedd94f06315c523f6cecf5b4827233bed7ae3fe135b2a0ebeab5e196"},
   "retry": {:hex, :retry, "0.15.0", "ba6aaeba92905a396c18c299a07e638947b2ba781e914f803202bc1b9ae867c3", [:mix], [], "hexpm", "93d3310bce78c0a30cc94610684340a14adfc9136856a3f662e4d9ce6013c784"},
-  "reverse_proxy_plug": {:git, "https://github.com/mozillareality/reverse_proxy_plug.git", "271111467623ca1f45c3837b14c4ca0661db0f3e", [branch: "reticulum/master"]},
```
taku1201, line 74: This line should have been changed in ef3fde099, but it has not. 
Therefore, we change this line to match the current situation.

keianhzo (approved):
LGTM Thanks!

hrithikwins (commented):
Much needed, thanks a lot

PR #724, opened 2024-11-02
Adds JSON-LD metadata so web crawlers know what the content of each page is

DougReeder:
There's no body text for web crawlers to work with, so metadata is important.

Also, changes social media previews to use the site-configured app description.

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -101,6 +101,7 @@ defmodule RetWeb.PageController do
         avatar: avatar,
         ret_meta: Ret.Meta.get_meta(include_repo: false),
         translations: app_config["translations"]["en"],
+        root_url: RetWeb.Endpoint.url(),
         app_config_script: {:safe, app_config_script |> with_script_tags},
         extra_script: {:safe, get_extra_script(:avatar) |> with_script_tags},
         extra_html: {:safe, get_extra_html(:avatar) || ""}
@@ -135,6 +136,7 @@ defmodule RetWeb.PageController do
         "index-meta.html",
         root_url: RetWeb.Endpoint.url(),
         translations: app_config["translations"]["en"],
+        images: app_config["images"],
         app_config_script: {:safe, app_config_script |> with_script_tags},
         extra_script: {:safe, get_extra_script(:index) |> with_script_tags},
         extra_html: {:safe, get_extra_html(:index) || ""}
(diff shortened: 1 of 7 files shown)


DougReeder:
Social media cards usually only display one line of description, so now the description uses the App Tagline, which is shorter than the App Description.
social media previews — Screenshot 2024-11-03 at 9 48 02 PM

lib/ret_web/controllers/page_controller.ex
```diff
@@ -99,8 +99,10 @@ defmodule RetWeb.PageController do
     avatar_meta_tags =
       Phoenix.View.render_to_string(RetWeb.PageView, "avatar-meta.html",
         avatar: avatar,
+        description: "“#{avatar.name}” is an avatar you can use in the #{app_config["translations"]["en"]["app-name"]} immersive space.",
```
Exairnous, line 102: Is this hard coding it to the English translation?

Exairnous (changes requested):
Sorry for the delay.  This mostly looks good (Thanks!), but there are a few things I would like to bring up.

1. I don't think we should be setting `isAccessibleForFree` because we can't know whether instance owners will be charging for things or not.
2. I'm not sure "maintainer" should be added to the screenshots.  It sort of suggests to me that the instance owner owns all pictures taken within their space.  If this is not the case and it's only for showing that it's "linked data", then I guess that's fine, but shouldn't it be on all of them then?
3. Could we keep the room description similar to what it was and pattern the scene description after the avatar description?  I think this would be a lot more useful to people (on the sites that show it) than just having the App Tagline.  (Specific wording suggestions below).

Room:
`Join others in an immersive space, right in your browser.`

Avatar (note: I removed the avatar name from the description because it's in the title just above):
`An avatar you can use in the #{app_config["translations"]["en"]["app-name"]} immersive space.`

Scene:
`A scene you can use in the #{app_config["translations"]["en"]["app-name"]} immersive space.`

DougReeder:
It turns out that `maintainer` is not a valid field for VirtualLocation. :-S 

I've added `creditText` to the screenshots, as Google won't display them as Rich Results otherwise. :-(

Exairnous (approved):
LGTM.  Thank you for the latest updates as discussed at the [dev meeting](https://docs.google.com/document/d/1OMh3-_DpjHABMMrXtECzLPNxipQMixx32l6K4zqSoYY/edit?tab=t.0#heading=h.5idumb13qaac).
Merging.

PR #725, opened 2024-11-26
Use reticulum as the docker repository instead of ret

Exairnous:
It was decided that continuing to use ret is too obscure and that now is the time to change it before too much history builds up.

Corresponding change for the Community Edition: https://github.com/Hubs-Foundation/hubs-cloud/pull/366

### DIFF
--- a/.github/workflows/ret-turkey.yml
+++ b/.github/workflows/ret-turkey.yml
@@ -1,4 +1,4 @@
-name: ret
+name: reticulum
 on:
   push:
     branches:


hrithikwins (approved):

DougReeder:
This is a great idea!  We should probably cross-reference the corresponding change in hubs-cloud.

Exairnous:
@DougReeder Thanks!  Good point.  I have now cross referenced them.

Issue #711, opened 2023-10-19
Elrang + Elixir version for ARM64 System


juber-nunes:
This is more a question than an issue.

I'm trying to have Reticulum install on an ARM64 system (Ubuntu 24.04) but the main problem is Erlang, the version suggested here won' t build due to OpenSSL version compatibility. 

Would it be possible to have a more recent version of Erlang, since Elixir 1.14 is compatible with Erlang 23-25?

Thanks in advance

![image](https://github.com/mozilla/reticulum/assets/48788224/ca9970a8-a8f9-478b-b85a-bcd374b47c34)

ajlennon:
fwiw I have done the work to run Hubs CE on arm64 and document the process for (e.g. a Raspberry Pi) here

https://github.com/dynamicdevices/hubs-cloud/

DougReeder:
Working on https://github.com/Hubs-Foundation/reticulum/issues/728 might clear the way to identifying places where the current code does not work well on both ARM and x86.

PR #727, opened 2025-04-16
Updates Elixir, Erlang OTP, Alpine & GitOps

DougReeder:
## What?
Updates Elixir and Erlang/OTP to latest that pass tests

## Why?
The proximal cause was GitOps lint-and-test being blocked for building using an old version of Ubuntu.
As we need to test for regressions, it made sense to update what else we can.

## How to test
I. GitOps lint-and-test Action should succed for Elixir 1.16 and Erlang/OTP 25.3 (the versions in turkeyDockerfile)
II. GitOps turkeyGitops Action should succeed (when running in the HubsFoundation repo)
III. hubs-compose with this should build and not have regressions
    1. run the `down` script
    2. cd to reticulum in hubs-compose
    3. pull this branch
    4. run `mix deps.get`
    5. run the `up` script
    6. if the reticulum container builds successfully, the Elixir automated tests have passed
    7. when all containers are running, smoke test Hubs

(This is currently working without any apparent errors in https://hubs.hominidsoftware.com/

No changes to functionality are expected.

## Limitations
The tests are run under Ubuntu, which has no equivalent to the Alpine used to build the image

## Alternatives considered
If there are regressions, we can switch to older version of Elixir and Erlang, that is still newer than what we're currently using.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -3,24 +3,73 @@ name: lint-and-test
 on: [push, workflow_dispatch]
 
 jobs:
-  test:
+  test-current:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - 'ubuntu-22.04'
+          - 'ubuntu-24.04'
+
+    runs-on: ${{ matrix.os }}
+
+    services:
+      postgres:
+        image: postgres:10
+        env:
+          POSTGRES_USER: admin
+          POSTGRES_PASSWORD: admin
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: erlef/setup-beam@v1
+        with:
+          elixir-version: '1.16'
+          otp-version: '25.3'
+
+      - uses: actions/cache@v4
+        with:
+          path: |
+            deps
+            _build
+          key: mix-${{ hashFiles('mix.lock') }}
+
+      - run: mix local.hex --force
+      - run: mix local.rebar --force
+      - run: mix deps.get
+      - run: mix compile   # after fixing warnings, add --warnings-as-errors
+      - run: mix test
+
+
+  test-matrix:
+    # if a matrix-job with newer versions passes, advance the current versions!
+    continue-on-error: true
+
     strategy:
       fail-fast: false
       matrix:
         versions:
-          - elixir: '1.14'
-            otp: '24.2'
-          - elixir: '1.14'
-            otp: '24.3'
-          - elixir: '1.14'
-            otp: '25.3'
-
-          - elixir: '1.15'
-            otp: '24.3'
-          - elixir: '1.15'
-            otp: '25.3'
-          - elixir: '1.15'
-            otp: '26.2'
+#          - elixir: '1.14'
+#            otp: '24.2'
+#          - elixir: '1.14'
+#            otp: '24.3'
+#          - elixir: '1.14'
+#            otp: '25.3'
+#
+#          - elixir: '1.15'
+#            otp: '24.3'
+#          - elixir: '1.15'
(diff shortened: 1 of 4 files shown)


DougReeder:
It appears that this will fix #711 and support https://discord.com/channels/498741086295031808/1158476691384062012/1361647882305798335

Exairnous (changes requested):
Nice.  Thank you.  This sounds great.

Is the matrix of different version tests planned to be left in or is it just for confirmation that the proposed version we're moving to works?

The lint part fails for me when I run it through my fork, but I noticed that it doesn't include a version for Elixir 1.16 and Erlang 25.3 (it does 1.16 and 24.3 or 1.17 and 25.3).

The ret-turkey.yml changes have been moved to #731, so they can be removed from this PR.

DougReeder:
My thinking is that the "linting" only needs to be done once per version of Elixir.  I use the earliest matching version of Erlang/OTP on the assumption that if the syntax is good for the first version, it's good for the others.

.github/workflows/lint-and-test.yml
```diff
@@ -3,8 +3,105 @@ name: lint-and-test
...
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: $otp_version
+          elixir-version: $elixir_version
```
Exairnous, line 40: Typos:
```suggestion
          otp-version: ${{ env.otp_version }}
          elixir-version: ${{ env.elixir_version }}
```

Exairnous (changes requested):
> My thinking is that the "linting" only needs to be done once per version of Elixir. I use the earliest matching version of Erlang/OTP on the assumption that if the syntax is good for the first version, it's good for the others.

Ah, okay.  That makes sense.  Thanks.

I found a typo in the `test-current` section, but otherwise the tests looks good.

About the linting, I forgot that we had files that weren't formatted as expected, so they should fail the workflow, my bad.  Sorry! :flushed:

Would it be possible to add in a `current` property to the `lint` matrix and set `continue-on-error` to not `current`?  This would seem to fit with the behaviour of the tests (i.e. fails on non-current versions are fine, but a fail on the current version fails the whole workflow).

.github/workflows/lint-and-test.yml
DougReeder: It appears the syntax I was using is allowed in steps, but not here. :-(  Thanks!

Exairnous:
The workflow fails because it can't access env variables in the matrix specification.  This should work to get around that.
```yaml
env:
  elixir_version_current: '1.16'
  otp_version_current: '25.3'

jobs:
  variables:
    runs-on: ubuntu-latest
    outputs:
      elixir_version_current: ${{ env.elixir_version_current }}
      otp_version_current: ${{ env.otp_version_current }}
    steps:
      - name: Job to satisfy syntax requirements
        run: |
          :
      
  <snip>
  
  test-matrix:
    needs: [variables]
    continue-on-error: true

    strategy:
      fail-fast: false
      matrix:
      <snip>
        exclude:
          - versions:
              elixir: ${{ needs.variables.outputs.elixir_version_current }}
              otp: ${{ needs.variables.outputs.otp_version_current }}
```





References:
https://stackoverflow.com/questions/74072206/github-actions-use-variables-in-matrix-definition
https://stackoverflow.com/questions/73305126/passing-environment-variables-as-inputs-to-a-reusable-workflow/77512339#77512339

DougReeder:
Your approach would work fine, but since the Elixir and Erlang/OTP versions need to be updated  in TurkeyDockerfile as well, I think hard-coding is more consistent. (You **have** to update multiple files, so you're more-or-less doing a search and replace anyway.)  This should be documented, but where?

Exairnous (approved):
LGTM.  Thanks.  Merging.

rawnsley:
@DougReeder I think there might be a problem with version upgrades: https://github.com/Hubs-Foundation/reticulum/issues/735

PR #730, opened 2025-04-19
fixes bug: HTML for room generated even if scene is nil

DougReeder:
## What?
Checks that scene is truthy before extracting scene name for metadata

If the scene is falsy, this error arises:
```
key :name not found in: nil. If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
```

## Why?
Apparantly, the scene can be nil, at least in hubs-compose, some of the time

Without this check, reticulum can't generate the HTML for the room

## How to test

1.     run the down script
2.     cd to reticulum in hubs-compose
3.     pull this branch
4.     run the up script
6.     open a Hubs room
7.     use the browser console to verify the JSON-LD metadata is there. `disambiguatingDescription` may be zero-length.

## Limitations
This PR doesn't update the lint-and-test GitOps

## Open questions
It's unclear if there's a more idiomatic way that works in an Elixir template

### DIFF
--- a/lib/ret_web/templates/page/hub-meta.html.eex
+++ b/lib/ret_web/templates/page/hub-meta.html.eex
@@ -8,7 +8,7 @@
         "url": "<%= @hub |> Ret.Hub.url_for %>",
         "image": "<%= @hub |> Ret.Hub.image_url_for %>",
         "description": "<%= @description %>",
-        "disambiguatingDescription": "<%= @scene.name %>"
+        "disambiguatingDescription": "<%= @scene && @scene.name %>"
     }
 </script>
 


Exairnous (approved):
LGTM.  Thanks.

I did some investigating and it looks like `@scene` is `nil` if the room is set to use the default fallback scene or a scene that has been imported from another Hubs instance via the admin console.  Scenes created locally either through a Spoke project or imported from Blender appear to be fine.  This probably points to an underlying bug somewhere (I noticed similar issues with the backup tool).

As to your open question, I don't think there are any drop in replacements that are more idiomatic.  `@scene[:name]` will work when the scene is `nil`, but when it's valid it causes an error because `Ret.Scene` doesn't implement the Access behavior (https://hexdocs.pm/elixir/Access.html).  `@scene |> Map.get(:name)` has the same problem as the current code (it fails when `@scene` is `nil`).  Everything else I found seems like it's even more complicated.

I'm going to merge this and release new stable Docker images.

PR #731, opened 2025-04-22
turkeyGitops only run on push to master in Hubs-Foundation

DougReeder:
## What?
Only builds images when the master branch on the Hubs-Foundatioin fork changes

## Why?
* avoids doomed attempts to run in forked repos
* avoids publishing from non-master branches

## How to test
1. push to master in forked repo
2. observe that action is kicked off, but job turkeyGitops doesn't run

3. push to non-master branch in original repo
4. observe that the action is not kicked off

Avoids failed runs; no new functionality

### DIFF
--- a/.github/workflows/ret-turkey.yml
+++ b/.github/workflows/ret-turkey.yml
@@ -2,11 +2,14 @@ name: reticulum
 on:
   push:
     branches:
-    paths-ignore: ["README.md"]
+      - 'master'
+      - 'hotfix/**'
+    paths-ignore: [".github/**", "guides/**", "README.md", "LICENSE", "CODE_OF_CONDUCT.md"]
   workflow_dispatch:
 
 jobs:
   turkeyGitops:
+    if: github.repository_owner == 'Hubs-Foundation'
     uses: Hubs-Foundation/hubs-ops/.github/workflows/turkeyGitops.yml@master
     with:
       registry: hubsfoundation


Exairnous (changes requested):
Since we have the custom-docker-build-push action, I think it's generally fine to restrict the branches this runs on, however I recently noticed that turkeyGitops supports building branches that start with `hotfix/` ([reference link](https://github.com/Hubs-Foundation/hubs-ops/blob/5de5b298faf117f9c4c92c06c4a19075be880895/.github/workflows/turkeyGitops.yml#L231)), so these should be allowed as well (I thought we were going to have to add support for hotfixes, so finding this was a nice surprise).

You might also want to consider adding in the `.github` folder to `paths-ignore` because changes to that will never need a new docker image.

_Note for future reference:_
Applying this directly to the turkeyGitops workflow was discussed at the last dev meetup, but it was decided that we'd test it here on Reticulum first to make sure it works, then put up an issue for transferring it to the turkeyGitops workflow once we know it works.
https://docs.google.com/document/d/1OMh3-_DpjHABMMrXtECzLPNxipQMixx32l6K4zqSoYY/edit?tab=t.0#heading=h.pinom5pwsi1h

Exairnous (commented):
The changes for hotfix branches are good, but I believe the folders in `paths-ignore` should share the same syntax, e.g. `".github/**"` and `"guides/**"`.  In my testing, the addition of `/**` was needed to have the workflow ignore changes to files within those directories.

PR #732, opened 2025-04-24
Assembles informative & polished metadata from whatever values are available

DougReeder:
## What?
Provides default values and fallback values when admin hasn't set an App Setting or the code doesn't currently supply it.
Assembles the available values so there are no unprofessional-looking gaps.
* For a room, if scene is nil, uses scenelist
* Provides context text, so social media previews are actionable
* If App Name is nil, uses hostname
* if Company Name is blank, uses App Name
* forms descriptions using all relevant values, eliminating nils and concatenating values using em-dashes

## Why?
Admins don't necessarily set all the App Settings.
There are bugs, such as the scene being unavailable for some rooms.
Metadata such as titles and descriptions is often the first contact a user has with an instance. If they contain obvious gaps, the instance looks amateurish.

## Examples
image

## How to test
1. Pull this branch into Hubs Compose
2. Run `bin up`
3. Navigate to root page, a room, a screenshot, a scene, and an avatar. For each, copy the HEAD tag as a snippet to https://validator.schema.org
4. Configure a public instance to pull reticulum Always, and from dougreeder/reticulumbeta:v1
5. Temporarily delete all values from the App Settings in the Admin Panel
6. Navigate to root page, a room, a screenshot, a scene, and an avatar. For each, paste the URL to https://www.opengraph.xyz and observe the previews
7. Restore values in the App Settings
8. Repeat step 6 and observe the expanded metadata

https://schema.org/VirtualLocation etc.
https://ogp.me/

## Limitations
There are values that Hubs doesn't currently record, so far as I know, such as which user took a screenshot, what room it was, and what scene.  Saving those values in the database and retrieving them would be non-trivial.

### DIFF
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -483,6 +483,12 @@ defmodule RetWeb.PageController do
 
     app_name = app_config["translations"]["en"]["app-name"] || RetWeb.Endpoint.host()
     scene = hub.scene || hub.scene_listing
+    name = cond do
+      hub.name && scene && scene.name && scene.name != hub.name -> join_smart([hub.name, scene.name])
+      hub.name -> hub.name
+      scene && scene.name -> scene.name
+      true -> "a room on " <> app_name
+    end
     hub_meta_tags =
       Phoenix.View.render_to_string(RetWeb.PageView, "hub-meta.html",
         hub: hub,
@@ -491,19 +497,16 @@ defmodule RetWeb.PageController do
         available_integrations_script: {:safe, available_integrations_script |> with_script_tags},
         translations: app_config["translations"]["en"],
         title: join_smart([hub.name, app_name]),
-        name: hub.name || scene.name || "a Hubs room",
+        name: name,
         description: join_smart([
           hub.description,
-          scene && scene.name,
           "an immersive space in #{app_name}, right in your browser",
           app_config["translations"]["en"]["app-description"],
           "powered by Hubs."
         ]) |> String.replace("\\n", " "),
         description_social_media: join_smart([
-          "Join others in an immersive space in #{app_name}",
+          "Join others in an immersive space in #{app_name}, right in your browser",
           hub.description,
-          scene && scene.name,
-          "right in your browser",
           app_config["translations"]["en"]["app-description"],
           "powered by Hubs."
         ]) |> String.replace("\\n", " "),
--- a/lib/ret_web/controllers/page_controller.ex
+++ b/lib/ret_web/controllers/page_controller.ex
@@ -17,6 +17,7 @@ defmodule RetWeb.PageController do
 
   alias Plug.Conn
   import Ret.ConnUtils
+  import Ret.HttpUtils
 
   ##
   # NOTE: In addition to adding a route, you must add static html pages to the page_origin_warmer.ex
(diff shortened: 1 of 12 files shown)


lib/ret_web/controllers/page_controller.ex
```diff
@@ -462,14 +478,35 @@ defmodule RetWeb.PageController do
     {_, available_integrations_script} =
       Ret.Meta.available_integrations_meta() |> generate_config("AVAILABLE_INTEGRATIONS")
 
+    app_name = app_config["translations"]["en"]["app-name"] || RetWeb.Endpoint.host()
+    scene = hub.scene || hub.scene_listing
+    name = cond do
+      hub.name && scene && scene.name && scene.name != hub.name -> join_smart([hub.name, scene.name])
+      hub.name -> hub.name
+      scene && scene.name -> scene.name
+      true -> "a room on " <> app_name
```
Exairnous, line 487: Do you have an example where the room name isn't present?

Exairnous (commented):
Generally LGTM.  I get no errors or crashes and it does look more professional.  Thanks.

> For a room, if scene is nil, uses scenelist

I'm not sure what you mean by `scenelist` here, could you elaborate?  From my testing, anything about the scene is omitted if the scene is nil.

Note: I was unable to test leaving any of the branding images omitted (there seems to be no way to remove them once added and I've already added all the ones relevant here), but given that it doesn't fail any of the if checks I think we can assume everything's fine.

lib/ret_web/templates/page/index-meta.html.eex
```diff
@@ -42,17 +53,20 @@
...
 <!-- Microsoft Edge -->
-<meta name="application-name" content="<%= @translations["app-full-name"] || @translations["app-name"] %>">
+<meta name="application-name" property="og:site_name" content="<%= @name %>">
+<meta name="creator" content="<%= @translations["company-name"] || @name %>" >
 <meta name="msapplication-tooltip" content="<%= @translations["app-tagline"] %>">
 <meta name="msapplication-starturl" content="/">
 <link rel="apple-touch-icon" href="/app-icon.png">
 
-<title><%= @translations["app-name"] %> - <%= @translations["app-tagline"] %></title>
+<title><%= @name %></title>
```
Exairnous, line 69: The tagline is no longer included here now, and I think this was pretty much the only place where it is visible on a Hubs instance.  I think it should be included at least somewhere on the instance, and in the future it would be nice to add it to more places.

DougReeder:
> > For a room, if scene is nil, uses scenelist
> 
> I'm not sure what you mean by `scenelist` here, could you elaborate? From my testing, anything about the scene is omitted if the scene is nil.

Hubs keeps track of `scene_listing` objects which contain 15 of the 22 fields in `scene` objects:

image

I don't know why this redudancy exists, but I have found sometime that `hub.scene_listing` is non-nil when `hub.scene` is nil.  Thus, my code uses `hub.scene_listing` if `hub.scene` is nil, to get the `name` for the hub's scene.

lib/ret_web/controllers/page_controller.ex
DougReeder: The only way I know for `hub.name` to be nil is to edit the database directly, but I'm reluctant to presume that `hub.name` is non-nil.
This whole PR was kicked off by finding that `hub.scene` could be nil, and the page wouldn't load, so I'm coding very conservatively.

We could omit line 486 `scene && scene.name -> scene.name`, and if the room name is nil, the metadata would show "a room on foo" as the room name.

lib/ret_web/templates/page/index-meta.html.eex
DougReeder: Good catch!  I've added the app tagline back into the page title.

Exairnous (approved):
Thank you for the explanations and updates.  Merging.

PR #733, opened 2025-05-04
Update Discord API to v10

Exairnous:
## What?

Updates the Discord client to use version 10 of the Discord API and account for its quirks/changes.

## Why?

The version of the Discord API was very old and will likely stop working in the near future.  Plus it's good to do as general maintenance.

## Examples

N/A

## How to test

1. Redeploy your instance using a Reticulum image generated from this PR (e.g. `ghcr.io/exairnous/reticulum:update-discord-api-to-v10-36`). 
2. Log out of the instance the Hubs bot is connected to.
3. Join a room created by the Hubs bot.
4. See that you can't join without being logged into Discord and having permission to view the channel that the room is bridged to.
5. See that your username is set/restricted to your Discord username or per-server nickname.
6. See that the permissions you've been granted are based on your permissions in Discord.

No functionality should have changed, so there is no need for documentation to be updated.

## Limitations

None.

## Alternatives considered

None.

## Open questions

None.

Discord seems to waffle between sending us integers and strings for permissions bitfield values depending on what permission level you are (e.g. users with room owner permissions seemed to have integers sent, but non-privileged users seemed to have strings sent), so this had to be accounted for by testing whether the value was a string or not and converting any strings to integers.

Discord has changed and usernames for people are now enforced to be unique and the discriminator numbers aren't appended to the end anymore (Discord bots still use discriminators, but this isn't relevant to us).
https://discord.fandom.com/wiki/Discriminator

Companion PR to https://github.com/Hubs-Foundation/hubs-discord-bot/pull/145 (although they should be able to be applied independently of each other).

### DIFF
--- a/lib/ret/discord_client.ex
+++ b/lib/ret/discord_client.ex
@@ -3,7 +3,7 @@ defmodule Ret.DiscordClient do
   import Bitwise
 
   @oauth_scope "identify email"
-  @discord_api_base "https://discordapp.com/api/v6"
+  @discord_api_base "https://discord.com/api/v10"
 
   def get_oauth_url(hub_sid) do
     authorize_params = %{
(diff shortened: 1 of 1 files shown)


Exairnous:
@HrithikWins I'd also appreciate your review on this.

DougReeder (approved):
LGTM

PR #734, opened 2025-05-04
Check for the global name if the per-server nickname isn't present

Exairnous:
## What?

If no per-server nickname has been set, this checks to see if Discord's global name is set for the user before falling back to the username.

## Why?

This makes the name used in Hubs match what Discord displays in the Discord server instead of predominantly using their (usually less aesthetically pleasing) username.

## Examples

N/A

## How to test

1. Redeploy your instance using a Reticulum image generated from this PR (e.g. `ghcr.io/exairnous/reticulum:use-discord-display-names-37`). 
2. Follow the instructions for hacking on the bot locally that are listed in PR https://github.com/Hubs-Foundation/hubs-discord-bot/pull/146 and set up the Hubs bot against your Community Edition instance.
3. Create a room with the Hubs bot with `!hubs create`.
4. Join the room created by the Hubs bot.
5. See that your username is set/restricted to your Discord display name or per-server nickname.

This is an implementation detail and doesn't need to be documented.  It is also immediately obvious.

## Limitations

If two Discord accounts join with matching names there is pretty much no way to tell them apart without the username being included.  However, this isn't any different than two people in a non Discord bot created room setting their names to be the same.

## Alternatives considered

Changing to just using the usernames, but this wouldn't look as nice or be in line with what people expect, plus the per-server nicknames being supported previously suggest that this PR is in line with what was originally intended.

## Open questions

Do we go with using the nicer names or enforce the usernames for moderation purposes?  If we enforce the usernames, the companion PR should be modified to do the same.

It is unlikely that the username will ever be used because the global name should always be set.

Depends on: https://github.com/Hubs-Foundation/reticulum/pull/733
[Diff of only changes from this PR](https://github.com/Hubs-Foundation/reticulum/compare/Exairnous:update-discord-api-to-v10...Exairnous:use-discord-display-names)

Companion PR to https://github.com/Hubs-Foundation/hubs-discord-bot/pull/148

### DIFF
--- a/lib/ret/discord_client.ex
+++ b/lib/ret/discord_client.ex
@@ -3,7 +3,7 @@ defmodule Ret.DiscordClient do
   import Bitwise
 
   @oauth_scope "identify email"
-  @discord_api_base "https://discordapp.com/api/v6"
+  @discord_api_base "https://discord.com/api/v10"
 
   def get_oauth_url(hub_sid) do
     authorize_params = %{
@@ -91,7 +91,7 @@ defmodule Ret.DiscordClient do
       }) do
     case Cachex.fetch(:discord_api, "/users/#{provider_account_id}") do
       {status, result} when status in [:commit, :ok] ->
-        "#{result["username"]}##{result["discriminator"]}"
+        "#{result["username"]}"
     end
   end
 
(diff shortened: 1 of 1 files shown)


Exairnous:
@HrithikWins I'd also appreciate your review on this.

DougReeder (approved):
LGTM

lib/ret/discord_client.ex
```diff
@@ -76,12 +76,20 @@ defmodule Ret.DiscordClient do
         {status, result} when status in [:commit, :ok] -> "#{result["nick"]}"
       end
 
-    if nickname == "" do
+    nickname = if nickname == "" do
```
DougReeder, line 79: The cautious check would be `if !nickname or nickname == "" do` but I'm fine either way.

Exairnous: I like cautious :)  Good catch, thanks.  Updated.

PR #738, opened 2025-06-26
pseudo-lints source files before commit & fixes pseudo-lint errors

DougReeder:
## What?
Adds setup script which enables a git pre-commit hook which pseudo-lints

Formats source files that weren't formatted according to the standard, using `mix commit`

## Why?
The pseudo-lint is run by CI, so it should be run before committing locally.
    There is no git-native way to check hooks into a project.
    This does add a step to setup, but other setup actions can and should be moved to the setup script over time.

CI tests that files are formatted.

## How to test
1. create and switch to a throwaway branch
2. run `scripts/setup.sh`
3. edit an Elixir file and add spaces to the end of a line
4. attempt to commit the changed file; observe that commit is blocked w/ error message
5. revert the changed file
6. commit; observer that commit is not blocked

README was updated with new step

## Limitations
Doesn't run automated tests locally, as those are not currently working for me and we need first to understand whether the automated tests can be run on any developer machine.

## Alternatives considered
Just configuring CI to fail unformatted files is a worse developer experience.

Also fails commit if any file contains non-ASCII characters.  This is a new test, but existing files pass it.

Files from a number of commits were not formatted, as that wasn't enforced  before.  
The second commit here formats them.

This configures CI to fail lint-and-test if any file is not formatted.

If we can figure out what's going on, we should also run the automated tests before a commit.


Exairnous:
LGTM!  Thanks.  Merging.

Exairnous (approved):
Add GitHub approval to make it happy.

